PR #17, opened 2018-11-15
Ataymano/unit tests fix

ataymano:
no http servers in unit tests


rlclientlib/utility/http_authorization.h
```diff
@@ -0,0 +1,55 @@
+#pragma once
```
jackgerrits: I love this change, thank you!

rlclientlib/logger/eventhub_client.h
```diff
@@ -91,18 +78,11 @@ namespace reinforcement_learning {
     eventhub_client& operator=(eventhub_client&&) = delete;
 
   private:
-    web::http::client::http_client _client;
-
+    std::unique_ptr<i_http_client> _client;
```
jackgerrits: Could you possibly add a clarifying comment for where we pass this object in that the class takes ownership of it and will delete at the end of lifetime. We have a lot of instances where we do not manage lifetime so we should be explicit here

PR #21, opened 2018-11-21
Ataymano/cs deferred action

ataymano:
Deferred action support in C# bindings.

### DIFF
--- a/bindings/cs/rl.net.native/rl.net.live_model.cc
+++ b/bindings/cs/rl.net.native/rl.net.live_model.cc
@@ -40,6 +40,16 @@ API int LiveModelChooseRank(livemodel_context_t* context, const char * event_id,
     return context->livemodel->choose_rank(event_id, context_json, *resp, status);
 }
 
+API int LiveModelChooseRankEx(livemodel_context_t* context, const char * event_id, const char * context_json, unsigned int flags, reinforcement_learning::ranking_response* resp, reinforcement_learning::api_status* status)
+{
+  return context->livemodel->choose_rank(event_id, context_json, flags, *resp, status);
+}
+
+API int LiveModelReportActionTaken(livemodel_context_t* context, const char * event_id, reinforcement_learning::api_status* status)
+{
+  return context->livemodel->report_action_taken(event_id, status);
+}
+
 API int LiveModelReportOutcomeF(livemodel_context_t* context, const char * event_id, float outcome, reinforcement_learning::api_status* status)
 {
     return context->livemodel->report_outcome(event_id, outcome, status);
--- a/bindings/cs/rl.net/LiveModel.cs
+++ b/bindings/cs/rl.net/LiveModel.cs
@@ -30,7 +30,6 @@ namespace Rl.Net {
         [DllImport("rl.net.native.dll")]
         private static extern int LiveModelReportActionTaken(IntPtr liveModel, [MarshalAs(NativeMethods.StringMarshalling)] string eventId, IntPtr apiStatus);
 
-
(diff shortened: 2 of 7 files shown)


bindings/cs/rl.net/LiveModel.cs
```diff
@@ -71,6 +77,24 @@ public bool TryChooseRank(string actionId, string contextJson, RankingResponse r
             return result == NativeMethods.SuccessStatus;
         }
 
+        public bool TryChooseRank(string actionId, string contextJson, ActionFlags flags, out RankingResponse response, ApiStatus apiStatus = null)
+        {
+            response = new RankingResponse();
+            return this.TryChooseRank(actionId, contextJson, flags, response, apiStatus);
+        }
+
+        public bool TryChooseRank(string actionId, string contextJson, ActionFlags flags, RankingResponse response, ApiStatus apiStatus = null)
```
homezcx: Do we need this method as public? It seems there is no need for a caller to create RankingResponse object and pass it as parameter.

bindings/cs/rl.net/LiveModel.cs
```diff
@@ -71,6 +77,24 @@ public bool TryChooseRank(string actionId, string contextJson, RankingResponse r
...
+        {
+            int result = LiveModelChooseRankEx(this.NativeHandle, actionId, contextJson, (uint)flags, response.NativeHandle, apiStatus.ToNativeHandleOrNullptr());
```
tyclintw: > apiStatus.ToNativeHandleOrNullptr() [](start = 124, length = 36)

Since apiStatus defaults to null we should not throw NullPtrExceptions. Please change in all methods to have null-conditional operators.
```csharp
apiStatus?.ToNativeHandleOrNullptr()
```

ataymano: It is not throwing right now. Please see ToNativeHandleOrNullptr here: https://github.com/VowpalWabbit/reinforcement_learning/blob/master/bindings/cs/rl.net/Native/Global.cs

bindings/cs/rl.net/LiveModel.cs
ataymano: I am ok to remove it, but for now I was just following the previously introduced style (see methods above)

homezcx: I see

bindings/cs/rl.net.native/rl.net.live_model.cc
```diff
@@ -40,6 +40,16 @@ API int LiveModelChooseRank(livemodel_context_t* context, const char * event_id,
     return context->livemodel->choose_rank(event_id, context_json, *resp, status);
 }
 
+API int LiveModelChooseRankEx(livemodel_context_t* context, const char * event_id, const char * context_json, unsigned int flags, reinforcement_learning::ranking_response* resp, reinforcement_learning::api_status* status)
```
sharathmalladi: >Ex [](start = 27, length = 2)

Could this be LiveModelChooseRankWithFlags instead?

jackgerrits: What does Ex indicate here?

jackgerrits:
Looks good to me, as far as marshalling goes you aren't introducing new behavior but just extending with the same procedure as before so I think it's okay.

bindings/cs/rl.net.native/rl.net.live_model.cc
ataymano: Changed ti ChooseRankWithFlags

PR #24, opened 2018-11-26
Rajan/preamble fb

rajan-chari:
- Flatbuffer 
- Object pooling and memory management for FB
- Preamble changes
- Manually merged master


build-linux.sh
```diff
@@ -15,12 +15,23 @@ sudo mkdir /opt/cmake
...
+make -j4
+sudo make install
+
 cd /reinforcement_learning
 
 # Build reinforcement_learning
 mkdir -p build
 cd build
-cmake .. -DTURN_OFF_DEVIRTUALIZE=On
+cmake .. -DTURN_OFF_DEVIRTUALIZE=On -DCMAKE_PREFIX_PATH=/root/flatbuffers/CMake
```
jackgerrits: Why do we need this if it has been installed?

examples/override_interface/override_interface.cc
```diff
@@ -21,30 +21,35 @@ namespace u = r::utility;
 namespace err = r::error_code;
 
 // Custom implementations must inherit from the respective i_* abstract class.
-class ostream_sender : public r::i_sender {
+class ostream_sender : public r::i_sender
+{
```
jackgerrits: braces on same line

include/data_buffer.h
```diff
@@ -0,0 +1,70 @@
...
+     * preamble is usually 8 bytes and sometimes zero if there is no preamble
+     */
+class data_buffer {
+public:
+  using value_type = unsigned char;
+
+  explicit data_buffer(size_t body_size = 1024);
+
+  // Get a pointer to beginning of preamble
+  value_type *preamble_begin();
```
jackgerrits: nit pointer on type

include/data_buffer.h
```diff
@@ -0,0 +1,70 @@
...
+
+  constexpr size_t preamble_size() const;
```
jackgerrits: 👍

include/data_buffer.h
```diff
@@ -0,0 +1,70 @@
...
+
+  // Return pointer to beginning of buffer, starting at offset from the real beginning.
+  value_type *body_begin();
+
+  // Body size (does not include the preamble.)
```
jackgerrits: nit: remove . from within parentheses

include/data_buffer.h
```diff
@@ -0,0 +1,70 @@
...
+  size_t body_filled_size() const;
+
+  // How large can we make the Body? (does not include the preamble.)
+  size_t body_capacity() const;
+
+  // Offset for beginning of the body region from the beginning of the buffer
+  size_t get_body_beginoffset() const;
```
jackgerrits: it feels like there should be a _ between begin and offset

include/sender.h
```diff
@@ -1,22 +1,21 @@
...
+    int send(buffer &data, api_status *status = nullptr) {
+      return v_send(data, status);
     }
 
     virtual ~i_sender() = default;
-  
+
   protected:
-    virtual int v_send(std::string&& data, api_status* status) = 0;
+    virtual int v_send(buffer &data, api_status *status = nullptr) = 0;
```
jackgerrits: & and * on type

rlclientlib/logger/flatbuffer_allocator.cc
```diff
@@ -0,0 +1,24 @@
...
+    : _buffer(data_buffer)
+  {}
+
+  uint8_t* flatbuffer_allocator::allocate(size_t size) {
+    _buffer.resize_body_region(size);
+    return _buffer.body_begin();
+  }
+
+  void flatbuffer_allocator::deallocate(uint8_t *p, size_t size) {
+
```
jackgerrits: Put a comment for why this is empty

rlclientlib/ranking_event.cc
```diff
@@ -51,108 +54,92 @@ namespace reinforcement_learning {
...
-  void ranking_event::serialize(u::data_buffer& oss) {
-    oss << _body;
-    if (_pass_prob < 1) {
-      oss << R"(,"pdrop":)" << (1 - _pass_prob);
-    }
-    oss << R"(})";
+    string context_str(context);
+    copy(context_str.begin(), context_str.end(), std::back_inserter(_context));
+    _approx_size = _context.size() +
+      _action_ids_vector.size() * sizeof(uint64_t) +
```
jackgerrits: Prefer to get these types from the vector itself
`decltype(_action_ids_vector)::value_type`

rlclientlib/schema/RankingEvent.fbs
```diff
@@ -0,0 +1,20 @@
...
+namespace reinforcement_learning.messages.flatbuff;
+
+table RankingEvent {
+    event_id:string;          // event IDs
+    deferred_action:bool = false;
+    action_ids:[uint64];      // action IDs
+    context:[ubyte];          // context
+    probabilities:[float];    // probabilities
+    model_id:string;          // model ID
+	pass_probability:float;	  // Probability of event surviving throttling operation
```
jackgerrits: mixed spaces and tabs

rlclientlib/serialization/fb_serializer.h
```diff
@@ -0,0 +1,146 @@
...
+  };
+
+  template<>
+  struct fb_event_serializer<outcome_event> {
+    using fb_event_t = OutcomeEventHolder;
+    using offset_vector_t = std::vector<flatbuffers::Offset<fb_event_t>>;
+    using batch_builder_t = OutcomeEventBatchBuilder;
+
+    static size_t size_estimate(const outcome_event& evt) {
+      return  evt.get_event_id().size() +
```
jackgerrits: nit: double space

rlclientlib/utility/data_buffer_streambuf.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
+#include <streambuf>
+namespace reinforcement_learning { namespace utility {
+  class data_buffer;
+  class data_buffer_streambuf :
```
jackgerrits: Can you please explain and add a comment for what streambuf is and used for

rlclientlib/utility/object_pool.h
```diff
@@ -1,130 +1,59 @@
...
-      : _version(0), _factory(factory), _used_objects(0)
-    { }
-
-    object_pool(const object_pool&) = delete;
-    object_pool& operator=(const object_pool& other) = delete;
-    object_pool(object_pool&& other) = delete;
-
-    ~object_pool() {
+    template <typename Object>
+    std::shared_ptr<Object> object_pool<Object>::acquire() {
```
jackgerrits: I like the change to shared_ptr here

rlclientlib/live_model_impl.cc
```diff
@@ -143,16 +144,38 @@ namespace reinforcement_learning {
...
     const auto ranking_sender_impl = _configuration.get(name::INTERACTION_SENDER_IMPLEMENTATION, value::INTERACTION_EH_SENDER);
-    i_sender* ranking_sender;
-    RETURN_IF_FAIL(_sender_factory->create(&ranking_sender, ranking_sender_impl, _configuration, &_error_cb, _trace_logger.get(), status));
-    _ranking_logger.reset(new interaction_logger(_configuration, ranking_sender, _watchdog, &_error_cb));
+    i_sender* ranking_data_sender;
+
+    // Use the name to create an instance of raw data sender for interactions
+    RETURN_IF_FAIL(_sender_factory->create(&ranking_data_sender, ranking_sender_impl, _configuration, &_error_cb, _trace_logger.get(), status));
+    RETURN_IF_FAIL(ranking_data_sender->init(status));
+
```
tyclintw: Why doesn't the factory call init?

rlclientlib/ranking_event.h
```diff
@@ -32,45 +34,61 @@ namespace reinforcement_learning {
...
-    
+    outcome_event(const char* event_id, float pass_prob, const char* outcome, bool _deferred_action);
+    outcome_event(const char* event_id, float pass_prob, float outcome, bool _deferred_action);
+  
   private:
-    std::string _body;
+    std::string _outcome;
+    float _float_outcome;
+    bool _deferred_action;
+    unsigned int outcome_type = 0;
```
tyclintw: >outcome_type [](start = 17, length = 12)

For consistency: "_outcome_type"

rlclientlib/utility/stl_container_adapter.h
```diff
@@ -0,0 +1,27 @@
+﻿#pragma once
+#include "data_buffer.h"
+#include <iterator>
+
+namespace reinforcement_learning {namespace utility {
+  class data_buffer;
+  class stl_container_adapter {
```
tyclintw: forward declaration not necessary if "data_buffer.h" is included.

rlclientlib/model_mgmt/restapi_data_transport.cc
```diff
@@ -128,7 +128,7 @@ namespace reinforcement_learning { namespace model_management {
       RETURN_ERROR_LS(_trace, status, exception_during_http_req) << error_code::unknown_s;
     }
 
-    return error_code::success;
+    return request_task.get();
```
tyclintw: > return request_task.get(); [](start = 3, length = 27)

Shouldn't this still return error_code such line 115?

rlclientlib/serialization/fb_serializer.h
```diff
@@ -0,0 +1,146 @@
...
+      return error_code::success;
+    }
+
+    uint64_t size() const {
+      return _builder.GetSize();
+    }
+
+    void finalize() {
+      auto event_offsets = _builder.CreateVector(_event_offsets);
+      typename serializer_t::batch_builder_t batch_builder(_builder);
```
tyclintw: Can we get rid of the batch builder. Flatbuffers compaction does not work well with our Event Hub batch sizing. One of my main reasons for supporting the preamble was to remove the necessity of flatbuffer batching. Can we have a preamble for each event?

include/data_buffer.h
```diff
@@ -0,0 +1,70 @@
...
+  value_type *raw_begin();
+
+private:
+  std::vector<value_type> _buffer;
+  // Offset for beginning of the body data from the beginning of the buffer
+  size_t _body_beginoffset;
+  // Offset for end of the body data from beginning of the buffer
+  size_t _body_endoffset;
+  // Size in bytes of the preamble region
+  const size_t _preamble_size;
```
tyclintw: IMO this class is very convoluted. There is a lot of duplication of offsets and opportunity for inconsistency. Why are _body_beginoffest and _body_endoffset necessary? Seems like these should be derived from the _buffer and _preamble_size; _body_beginoffset = _buffer.begin() + _preamble_size; and _body_endoffset = _buffer.end(); Likewise, I'm unclear why set_body_beginoffset should be settable. When would the _body_beginoffset ever be more offset from the raw_begin than the _preamble_size?

rlclientlib/serialization/fb_serializer.h
rajan-chari: We added batching for throughput reasons.  Benchmarking EventHub REST API suggests that optimal payload size is around a 100k.  Performance drops off below and above that size.  
Can you elaborate on "Flatbuffers compaction does not work well with our Event Hub batch sizing"?

include/data_buffer.h
rajan-chari: - Buffer class efficiently supports filling the buffer from the start (JSON serialization) or the end (Flatbuffer Serialization) and is used to reduce memory allocations and copies.  
- body_beginoffset and body_endoffset are both necessary for that reason.
- body_beginoffset != buffer.begin() + preamble_size if buffer is filled from the end.
- set_body_beginoffset is needed when filling the buffer from the end

rlclientlib/ranking_event.h
rajan-chari: Make sense

cmake/Modules/BuildFlatBuffers.cmake
```diff
@@ -0,0 +1,148 @@
+# Copyright 2015 Google Inc. All rights reserved.
```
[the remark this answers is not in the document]
rajan-chari: Make sense.  This file is not needed only FindFlatBuffers.cmake

rlclientlib/logger/eventhub_client.cc
```diff
@@ -20,14 +21,14 @@ namespace reinforcement_learning {
     i_http_client* client,
     const std::string& host,
     const std::string& auth,
-    std::string&& post_data,
+    buffer& post_data,
```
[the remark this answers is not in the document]
rajan-chari: Makes sense.

rlclientlib/utility/stl_container_adapter.h
```diff
@@ -0,0 +1,27 @@
...
+  public:
+    using value_type = data_buffer::value_type;
+
+    explicit stl_container_adapter(data_buffer* db);
+    stl_container_adapter(const stl_container_adapter& rhs) = default;
+    stl_container_adapter(stl_container_adapter&& rhs) noexcept = default;
+    size_t size() const;
+    value_type& operator[] (size_t idx);
```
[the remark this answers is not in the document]
rajan-chari: Yep.  const indexer makes sense.

build-linux.sh
rajan-chari: Not needed.  Removed.

examples/override_interface/override_interface.cc
rajan-chari: Ok

include/data_buffer.h
rajan-chari: Ok

include/data_buffer.h
rajan-chari: Had to get rid of constexpr here.  gcc complained that the class does not have a trivial destructor.

include/data_buffer.h
rajan-chari: Ok

rlclientlib/ranking_event.cc
rajan-chari: Makes sense.

rlclientlib/schema/RankingEvent.fbs
rajan-chari: Will Fix.

rlclientlib/serialization/fb_serializer.h
rajan-chari: Will fix

rlclientlib/utility/data_buffer_streambuf.h
rajan-chari: Sure.

rlclientlib/utility/object_pool.h
rajan-chari: 👍

include/data_buffer.h
```diff
@@ -0,0 +1,70 @@
+#pragma once
+#include <string>
+#include <vector>
+
+namespace reinforcement_learning { namespace utility {
+
+/*
+     * Data buffer used for serialized messages.  Data buffer
```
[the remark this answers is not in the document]
rajan-chari: Will fix.

rlclientlib/utility/stl_container_adapter.h
rajan-chari: Makes sense.

rlclientlib/model_mgmt/restapi_data_transport.cc
rajan-chari: No.  In this case the behavior is right.  The async task has the error code we are looking for.

cmake/Modules/BuildFlatBuffers.cmake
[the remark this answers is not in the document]
jackgerrits: Don't you use the build command in the CMakeLists file

rlclientlib/logger/async_batcher.h
```diff
@@ -169,7 +169,7 @@ namespace reinforcement_learning { namespace logger {
     _queue_mode(queue_mode)
   {}
 
-  template<typename TEvent, template<typename> typename TSerializer>
+  template<typename TEvent, template<typename> class TSerializer>
```
jackgerrits: Why change this to class?

rlclientlib/serialization/fb_serializer.h
tyclintw: I completely agree with batching the Event Hub message. The question comes down to batching the events at the level of the FlatBuffers or at the Event Hub message level. My concern was the fact that we couldn't truly know the fb_collection_serializer.size() until after fb_collection_serializer.finalize() is called. This makes the fill_buffer conditional less than perfect. My thought process was to remove the array (RankingEventBatch) , which was necessary due to lack of a preamble, and use multiple preambles to create our own array. Also, being inline with the future log format I see "[preamble][event][preamble][event]....." much simpler than "[preamble][X-events][preamble][Y-events]...". It not lost on me that after your refactor, everything is much cleaner and the  fb_collection_serializer.finalize() really shouldn't change the size much; if at all.

---
In reply to: [237304366](https://github.com/VowpalWabbit/reinforcement_learning/pull/24#discussion_r237304366) [](ancestors = 237304366)

rlclientlib/logger/eventhub_client.cc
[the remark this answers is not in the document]
rajan-chari: Removed move for caller

rlclientlib/utility/stl_container_adapter.h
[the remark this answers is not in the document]
rajan-chari: Changed to const indexer and const data_buffer& in i_sender and i_message_sender

include/data_buffer.h
rajan-chari: I went back and forth on naming this funtion.  The competing elements are 1) body is an entity in itself. begin_offset is an idea pertaining to body.  vs 2) get_body_begin_offset just seems cleaner.  Ultimately the reason I wend with 1 is that this class is particularly confusing around the idea of body and preamble and it made more sense to make that less confusing than make this name seem cleaner.

include/sender.h
rajan-chari: Reformatted sender.h

rlclientlib/ranking_event.cc
rajan-chari: Dead code.  No longer needed.

rlclientlib/schema/RankingEvent.fbs
rajan-chari: Done.

cmake/Modules/BuildFlatBuffers.cmake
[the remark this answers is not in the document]
rajan-chari: You are right.  FindFlatBuffers.cmake does not work with the BuildFlatBuffers.cmake

rlclientlib/serialization/fb_serializer.h
rajan-chari: Will have to revisit this in the future.  Probably a considerable redesign of buffering pipeline to avoid memory reallocation and copies.  However having messgetype in the preamble will make this type of change low impact in the future.  The serverside already setup to handle either individual or batch message.

rlclientlib/logger/async_batcher.h
rajan-chari: Compiler in linux had issues with template template arguments.

rlclientlib/serialization/fb_serializer.h
```diff
@@ -0,0 +1,119 @@
...
+namespace reinforcement_learning { namespace logger {
+  template <typename T>
+  struct fb_event_serializer;
+  template <>
+  struct fb_event_serializer<ranking_event> {
+    using fb_event_t = RankingEvent;
+    using offset_vector_t = typename std::vector<flatbuffers::Offset<fb_event_t>>;
+    using batch_builder_t = RankingEventBatchBuilder;
+
+    static size_t size_estimate(const ranking_event& evt) {
```
ataymano: This size estimation requires for controlling size of queue in memory (before serialization), and there is nothing fb-specific in implementation as well. So, probably it should not be part of serializer<T>? We have static assert that any event have to be inherited from base event class, so probably it is cleaner to add size() method to that interface? In this case we also can get rid of size argument for event_queue::push.

PR #25, opened 2018-11-27
Update readme for clearer instructions

jackgerrits:


sharathmalladi:
> cmake .. -G "Visual Studio 15 2017 Win64" -DCMAKE_TOOLCHAIN_FILE=[vcpkg root]\scripts\buildsystems\vcpkg.cmake -DVCPKG_TARGET_TRIPLET=x64-windows

Could you add a comment on what this command is meant for? Does it output anything? Do we run this every time that we build?

---
Refers to: README.md:78 in 8dc258c. [](commit_id = 8dc258c64be2cd87368f5681aa3d17ef38d709cf, deletion_comment = False)

README.md
```diff
@@ -27,35 +31,47 @@ cd ../../../
...
+
+```
+vcpkg install cpprestsdk:x64-windows
+```
+
+### Build + Test
+
+Open `rl.sln` in Visual Studio 2017.
+
 ### Experimental - CMake on Windows
```
sharathmalladi: >Experimental [](start = 4, length = 12)

Experimental .. as in is this an alternate optional way to build the project? If that is the case, could we call this out in a separate section so it is not confused with the rest of how configure/build works on Windows?

sharathmalladi:
> ## Make targets

Could you add some comment above line 83 such as 'The following are the output targets of each of these projects' .. something that describes what this mapping is about.

---
Refers to: README.md:82 in 8dc258c. [](commit_id = 8dc258c64be2cd87368f5681aa3d17ef38d709cf, deletion_comment = False)

README.md
sharathmalladi: Or maybe it should say ## Experimental instead of ### Experimental on line 63. I am trying to understand if this section is required or some optional way of building on Windows. If it is optional, also call that out explicitly.

---
In reply to: [237568116](https://github.com/VowpalWabbit/reinforcement_learning/pull/25#discussion_r237568116) [](ancestors = 237568116)

jackgerrits: I think it should stay at the same nesting level as there is a section for Windows and Ubuntu. I've added a comment.

README.md
```diff
@@ -27,35 +31,48 @@ cd ../../../
...
+Using CMake is an alternate way to configure and build the project. Currently it only supports the C++ projects.
+All dependencies are managed through Vcpkg:
+```
+vcpkg install cpprestsdk:x64-windows
+vcpkg install zlib:x64-windows
+vcpkg install boost-system:x64-windows
+vcpkg install boost-program-options:x64-windows
+vcpkg install boost-test:x64-windows
+vcpkg install boost-uuid:x64-windows
+```
```
homezcx: This instruction is not up to date, missing flatbuffers. It should align with build-windows.cmd

ras0219-msft: vcpkg supports response files, so it might be easier to put the packages into a file (packages.txt) and install everything as:
```
>vcpkg install @packages.txt
```

jackgerrits: @ras0219-msft That's a handy new feature, thanks for letting us know! I am hesitant to do that right now since the default on Windows is x86, we would then need a specific package requirements file for each platform as we target x64 by default

PR #26, opened 2018-12-06
Enable ControlFlowGuard

tyclintw:
Necessary for the native dll to pass binskim.


jackgerrits:
Could you please enable this in CMake aswell?

rajan-chari:
Reviewed.  Looks good. No comments.

lokitoth:
>           Could you please enable this in CMake aswell?

I think this is a compiler/linker specific feature. I believe it exists in Clang as Control Flow Integrity, but I do not know if GCC has it. There may be third-party libraries that can be used here: https://grsecurity.net/rap_faq.php

I have a question: Does this impact the return callback in the evening layer? I tried finding documentation on CFG and P/Invoke, but could not see any details.

jackgerrits:
We can add compiler flags depending on the compiler being used. So we can check if it is MSVC or Clang etc and act accordingly

PR #29, opened 2018-12-15
Local loop part1

rajan-chari:
This PR includes:
1) Added a new transport for events that writes to files. (event_hub sender via REST already exists)
2) Added a "write to file" option to rl_sim_cpp sample
3) Created a tool to read flatbuffer formatted log file and convert to text and print to console.


rlclientlib/logger/file/file_logger.cc
```diff
@@ -0,0 +1,39 @@
...
+namespace reinforcement_learning { namespace logger { namespace file {
+
+  file_logger::file_logger(const std::string& file_name, i_trace* trace)
+  : _file_name(file_name),
+  _trace(trace)
+  {}
+
+  file_logger::~file_logger()
+  {
+      _file.close();
```
jackgerrits: This is not needed, when the object is destroyed the file will automatically be closed.

rlclientlib/logger/file/file_logger.cc
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  int file_logger::init(api_status* status) {
```
jackgerrits: Can this object be changed to a factory method that returns a valid file logger and the constructor or factory method does the open. It will just help reduce the potential for unitialized objects.

test_tools/joiner/text_converter.cc
```diff
@@ -0,0 +1,147 @@
...
+  void convert_to_text(const std::string& file);
+  void print_ranking_event(void* buff);
+  void print_outcome_event(void* buff);
+  void print_numeric_outcome(const flat::OutcomeEventHolder* evt);
+  void print_string_outcome(const flat::OutcomeEventHolder* evt);
+  void print_action_outcome(const flat::OutcomeEventHolder* evt);
+  ////
+
+  void convert_to_text(const std::vector<std::string>& files) {
+    for(const auto& file:files) {
```
jackgerrits:
```suggestion
    for(const auto& file : files) {
```

test_tools/joiner/text_converter.cc
```diff
@@ -0,0 +1,147 @@
...
+// namespace aliases
+namespace rlog = reinforcement_learning::logger;
+namespace flat = reinforcement_learning::messages::flatbuff;
+////
+
+namespace reinforcement_learning { namespace joiner {
+
+  // forward declarations 
+  void convert_to_text(const std::string& file);
+  void print_ranking_event(void* buff);
```
jackgerrits: consider passing this an `ostream` so that it can be reused for printing to stdout and a file

test_tools/joiner/text_converter.cc
```diff
@@ -0,0 +1,147 @@
...
+    for(const auto& file:files) {
+      convert_to_text(file);
+    }
+  }
+
+  void convert_to_text(const std::string& file){
+    std::ifstream infile;
+    infile.open(file, std::ios_base::binary);
+    if (infile.fail() || infile.bad()){
+      std::cout << "Unable to open file: " << file << std::endl;
```
jackgerrits:
```suggestion
      std::cerr<< "Unable to open file: " << file << std::endl;
```

test_tools/joiner/text_converter.cc
```diff
@@ -0,0 +1,147 @@
...
+    }
+    std::cout << "File:" << file << std::endl;
+    do {
+      char raw_preamble[8];
+      if (infile.fail() || infile.bad())
+        return;
```
jackgerrits: throw?

test_tools/joiner/text_converter.cc
```diff
@@ -0,0 +1,147 @@
...
+      infile.read(raw_preamble, 8);
+      rlog::preamble p;
+      p.read_from_bytes(reinterpret_cast<uint8_t*>(raw_preamble), 8);
+      std::unique_ptr<char> msg_data(new char[p.msg_size]);
+      infile.read(msg_data.get(), p.msg_size);
+      if (infile.fail() || infile.bad())
+        return;
```
jackgerrits: throw?

rlclientlib/logger/file/file_logger.cc
rajan-chari: Removed.

rlclientlib/logger/file/file_logger.cc
rajan-chari: Since we have restriction on throw we construct and init() everywhere in the library.   This class is instantiated by a factory as are all other loggers.  But the factory itself is generic across many object types.  Does not seem appropriate to call init() from the factory.  live_model_impl ensures that init() will always be called.

test_tools/joiner/text_converter.cc
rajan-chari: Done.

test_tools/joiner/text_converter.cc
rajan-chari: Don't want to stop processing.  Other files might be fine.  Writing to cerr makes sense.

test_tools/joiner/text_converter.cc
rajan-chari: Don't want to stop processing.  Other files might be fine.  Writing to cerr makes sense.

PR #30, opened 2018-12-18
Ataymano/fb polishing

ataymano:
DeferredAction / Action taken cleanup (after FB changes)

### DIFF
--- a/rlclientlib/live_model_impl.cc
+++ b/rlclientlib/live_model_impl.cc
@@ -60,7 +60,7 @@ namespace reinforcement_learning {
     RETURN_IF_FAIL(check_null_or_empty(event_id, context, status));
     if (!_model_data_received) {
       RETURN_IF_FAIL(explore_only(event_id, context, response, status));
-      response.set_model_id("N / A");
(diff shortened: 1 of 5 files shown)


rlclientlib/live_model_impl.cc
```diff
@@ -60,7 +60,7 @@ namespace reinforcement_learning {
     RETURN_IF_FAIL(check_null_or_empty(event_id, context, status));
     if (!_model_data_received) {
       RETURN_IF_FAIL(explore_only(event_id, context, response, status));
-      response.set_model_id("N/A");
+      response.set_model_id("N / A");
```
yannstad:
```suggestion
       response.set_model_id("N/A");
```

ataymano: I am ok with this, but all old logs (before flatbuffers change) are having "N / A" - not sure if there is anything (I. e. offline scripts) dependent on this value

ataymano: Oh no, I am wrong - will double check

PR #31, opened 2018-12-21
fix memory leak in Rank

homezcx:


rlclientlib/vw_model/safe_vw.cc
```diff
@@ -142,6 +142,10 @@ namespace reinforcement_learning {
     for (auto&& ex : examples)
       _example_pool.emplace_back(ex);
 
+    for (auto e : examples) {
+      e->pred.a_s.delete_v();
+    }
```
rajan-chari: Why not combine this into one loop?

homezcx: :( didn't notice that. Will update.

rlclientlib/vw_model/safe_vw.cc
```diff
@@ -138,9 +138,11 @@ namespace reinforcement_learning {
       scores[i] = predictions[i].score;
     }
 
-    // push examples back into pool for re-use
-    for (auto&& ex : examples)
+    // push examples back into pool for re-use and clean up examples
+    for (auto&& ex : examples) {
       _example_pool.emplace_back(ex);
+      ex->pred.a_s.delete_v();
```
rajan-chari: Clean up the example before putting it back in the pool?

homezcx: code and comment are updated.

homezcx:
@rajan-chari  @ataymano @jackgerrits ping to see if we are good to merge.

PR #33, opened 2019-01-15
i_sender ownership fix

ataymano:


lokitoth:
Yep, thanks @homezcx for finding it

PR #41, opened 2019-02-14
Implement PDF Passthrough (Sample Mode)

lokitoth:
Enable passing a pre-scored PDF into live_model to be sampled from directly.

This enables prototyping with additional model types without implementing the full contract, while taking advantage of standardized sampling and logging.

To enable this, we also create a new model "source": "No Model Data", to simplify the API of initializing the model (since the AzureBlob source requires an HTTP GET-able resource; the workaround would be to create an empty blob somewhere).

Also contains fixes:
* Fixes an issue where refresh_model was not setting _model_data_received, and getting stuck in "initial exploration"
* Expose refresh_model in Python
* Expose Configuration get/set and RefreshModel in C#


bindings/cs/rl.net/Configuration.cs
```diff
@@ -27,5 +33,24 @@ public static bool TryLoadConfigurationFromJson(string json, out Configuration c
...
         }
+
+        public string this[string key]
+        {
+            get
+            {
+                IntPtr result = ConfigurationGet(this.NativeHandle, key, string.Empty);
+                if (result == IntPtr.Zero)
+                {
+                    // TODO: Should we use string.Empty?
```
tyclintw: >// TODO: Should we use string.Empty? [](start = 20, length = 36)

I think we should at least be consistent. If the default value of `ConfigurationGet` is `string.Empty` then I would suggest returning the same value here.

rlclientlib/live_model_impl.cc
```diff
@@ -112,6 +112,8 @@ namespace reinforcement_learning {
     RETURN_IF_FAIL(_transport->get_data(md, status));
     RETURN_IF_FAIL(_model->update(md, status));
 
+	  _model_data_received = true;
```
tyclintw: >   [](start = 1, length = 2)

[supernit] spacing

rlclientlib/vw_model/pdf_model.cc
```diff
@@ -0,0 +1,45 @@
...
+      // TODO Should we provide some mechanism for the user to inject the model ID?
+      // response.set_model_id(vw->id());
+
+      return error_code::success;
+    }
+    catch ( const std::exception& e) {
+      RETURN_ERROR_LS(_trace_logger, status, model_rank_error) << e.what();
+    }
+    catch ( ... ) {
+      RETURN_ERROR_LS(_trace_logger, status, model_rank_error) << "Unkown error";
```
tyclintw: >Unkown [](start = 67, length = 6)

Unknown

unit_test/live_model_test.cc
```diff
@@ -116,6 +117,29 @@ BOOST_AUTO_TEST_CASE(live_model_ranking_request) {
   BOOST_CHECK_EQUAL(status.get_error_msg(), "");
 }
 
+BOOST_AUTO_TEST_CASE(live_model_ranking_request_pdf_passthrough) {
```
tyclintw: >live_model_ranking_request_pdf_passthrough [](start = 21, length = 42)

Can we add a sanity check that at least verifies the probability responses match the expected PASSTHROUGH_PDF_MODEL probabilities?

bindings/cs/rl.net.cli/EntryPoints.cs
```diff
@@ -9,8 +9,9 @@ static class EntryPoints
         public static void Main(string [] args)
         {
             //BasicUsageExample(args);
+            PdfPassthroughExample(args);
```
tyclintw: >PdfPassthroughExample [](start = 12, length = 21)

This is a bit out of scope for this PR but instead of modifying the code to run the examples can we just add an `args` value to trigger the appropriate example?

dwaijam: Instead of adding another arg maybe the it can be offered as an option to select once the program starts ?

bindings/cs/rl.net.cli/EntryPoints.cs
```diff
@@ -109,6 +110,37 @@ public static void BasicUsageExample(string [] args)
             }
         }
 
+        public static void PdfPassthroughExample(string [] args)
+        {
+            const float outcome = 1.0f;
```
dwaijam: Add args length check

rlclientlib/vw_model/vw_model.cc
```diff
@@ -41,27 +40,9 @@ namespace reinforcement_learning { namespace model_management {
...
-
-      // Swap values in first position with values in chosen index
-      scode = e::swap_chosen(std::begin(response), std::end(response), chosen_index);
-
-      if ( S_EXPLORATION_OK != scode ) {
-        RETURN_ERROR_LS(_trace_logger, status, exploration_error) << "Exploration error code: " << scode;
-      }
-
-      response.set_chosen_action_id(action_ids[chosen_index]);
+      // TODO: Should there be an i_pdf_model? It seems like the ability to integrate at the PDF level easily would be nice.
```
rajan-chari: Agree.  

I think current i_model does not work well.  The following change makes sense to me.

1) i_model that only returns collection of (actions, probabilities)
2) not sure if an i_sampler is needed simply the sampling functionality moved out of i_model.

I meant to go back and fix this debt.

I would be all for changing i_model now with this PR since no one else is using it and we are promoting a bad pattern that others are likely to mimic.

rlclientlib/vw_model/pdf_model.cc
```diff
@@ -0,0 +1,45 @@
...
+    _trace_logger(trace_logger), _vw(new safe_vw("--json --quiet --cb_adf")) {
+  }
+
+  int pdf_model::update(const model_data& data, api_status* status) {
+    return error_code::success;
+  }
+
+  int pdf_model::choose_rank(uint64_t rnd_seed, const char* features, ranking_response& response, api_status* status) {
+    try {
+      // TODO: is there a simpler way to get a vw object if we do not actually need to deal with model mgmt?
```
rajan-chari: I am not aware of any other way.  We talked about creating a library of example parsers that lives independently of vw.  Will add an issue in vw

rlclientlib/vw_model/pdf_model.cc
```diff
@@ -0,0 +1,45 @@
...
+#include "object_factory.h"
+#include "sampling.h"
+#include "ranking_response.h"
+#include "trace_logger.h"
+#include "str_util.h"
+
+namespace reinforcement_learning { namespace model_management {
+
+  pdf_model::pdf_model(i_trace* trace_logger) :
+    _trace_logger(trace_logger), _vw(new safe_vw("--json --quiet --cb_adf")) {
```
jackgerrits: why is json passed?

bindings/cs/rl.net.cli/EntryPoints.cs
lokitoth: Added #48 and #49 to track this (and related work) to clean up the usage/purpose of RL.Net.Cli

rlclientlib/vw_model/pdf_model.cc
lokitoth: To be able to use the json parser to parse the joined event and grab the PDF out of it.

bindings/cs/rl.net.cli/EntryPoints.cs
lokitoth: Will do this as part of #48 and #49

rlclientlib/model_mgmt/model_downloader.cc
```diff
@@ -30,11 +30,6 @@ namespace reinforcement_learning { namespace model_management {
     model_data md;
     RETURN_IF_FAIL(_ptrans->get_data(md, status));
 
-    // If the data size is zero, it's not a valid model
```
rajan-chari: Today when the loop is created a zero length model file is also created.  We test for the existence of this model file during init as an additional guard to make sure that the loop is created properly.  However, we don't hydrate a vw model with this empty file.  Will that continue to work?

lokitoth: Yes - what happens now is that the i_model implementation is expected to specify whether it is ready to handle inference at update() time. Thus, models that are self-contained can participate in this without the model-downloader having special-case logic.

I think there is more refactoring to be done here, but that would likely require a change in how configuration works.

rajan-chari:
Looks good to me.

Issue #39, opened 2019-01-31
C# Bindings should use "EventId" rather than "ActionId"


lokitoth:
There is a naming inconsistency between our documentation and the C# bindings in the LiveModel TryXYZ APIs.

PR #42, opened 2019-02-14
Create more idiomatic API for .NET

lokitoth:
* Adds non-TryXYZ equivalents which wrap the ApiStatus mechanism into a proper exception
* Fixes an issue where eventId was incorrectly mislabeled as actionId (#39)

### DIFF
--- a/bindings/cs/rl.net/LiveModel.cs
+++ b/bindings/cs/rl.net/LiveModel.cs
@@ -72,46 +72,106 @@ namespace Rl.Net
             return result == NativeMethods.SuccessStatus;
         }
 
-        public bool TryChooseRank(string actionId, string contextJson, out RankingResponse response, ApiStatus apiStatus = null)
+        public void Init()
+        {
+            ApiStatus apiStatus = new ApiStatus();
+            if (!this.TryInit(apiStatus))
+            {
+                throw new RLException(apiStatus);
+            }
+        }
+
+        public bool TryChooseRank(string eventId, string contextJson, out RankingResponse response, ApiStatus apiStatus = null)
         {
             response = new RankingResponse();
-            return this.TryChooseRank(actionId, contextJson, response, apiStatus);
+            return this.TryChooseRank(eventId, contextJson, response, apiStatus);
         }
 
-        public bool TryChooseRank(string actionId, string contextJson, RankingResponse response, ApiStatus apiStatus = null)
+        public bool TryChooseRank(string eventId, string contextJson, RankingResponse response, ApiStatus apiStatus = null)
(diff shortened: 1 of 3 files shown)


bindings/cs/rl.net/RLException.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            this.ErrorCode = status.ErrorCode;
+        }
+
+        protected RLException(SerializationInfo info, StreamingContext context) : base(info, context)
+        {
+            this.ErrorCode = info.GetInt32(ErrorCodeSerializationKey);
+        }
+
+        [SecurityPermission(SecurityAction.Demand, SerializationFormatter = true)]
```
kumpera: Do we still need this stuff?

bindings/cs/rl.net/LiveModel.cs
```diff
@@ -72,48 +72,108 @@ public bool TryInit(ApiStatus apiStatus = null)
...
+        public bool TryReportActionTaken(string eventId, ApiStatus apiStatus = null)
+        {
+            int result = LiveModelReportActionTaken(this.NativeHandle, eventId, apiStatus.ToNativeHandleOrNullptr());
             return result == NativeMethods.SuccessStatus;
         }
 
-        public bool TryReportOutcome(string actionId, float outcome, ApiStatus apiStatus = null)
+        public void ReportActionTaken(string eventId)
+        {
+            ApiStatus apiStatus = new ApiStatus();
```
kumpera: We should be disposing the temporary ApiStatus objects.

bindings/cs/rl.net/RLException.cs
lokitoth: If we want to support scenarios like remoting (or serializing for some other reason) exceptions, we need this. Is there a good reason to remove it?

bindings/cs/rl.net/LiveModel.cs
lokitoth: Good call.

lokitoth:
Simple rebase on top of master

bindings/cs/rl.net/RLException.cs
kumpera: dotnet code doesn't support remoting, but has most of the serialization stack.
It's harmless to have it, OTOH.

lokitoth: Fair enough. I'll keep this since we technically do support .Net Framework.

lokitoth:
Will be merging this in EOD today. Please let me know if there are any more comments/suggestions.

Issue #49, opened 2019-03-01
Split out example code from RL.Net.Cli


lokitoth:
RL.Net.Cli started out as a project to test and debug the RL.Net bindings. Over time we added simulator and replay functionality to it. Long-term, we should probably keep the samples grouped together, and out of the simulation/replay tools.

PR #58, opened 2019-03-27
Cs Cli perf test

ataymano:
cs client arg parsing refactoring + perf command


lokitoth:
Something to consider here is to avoid changing the BasicUsageExample and PdfPassthroughExample (and make them inaccessible when running .Cli), in preparation for #49. 

It is a bit of a bad legacy of my using it as a testing harness that they were in RL.Net.Cli in the first place, and we want to keep the code simple since they are being used as examples.

PR #59, opened 2019-04-01
Implement Unicode support for RL.Net

lokitoth:
When we first built the .Net bindings for rlclientlib we make a decision to only support ASCII. (VowpalWabbit/vowpal_wabbit#1601). This will behave incorrectly when sending text with code points outside of ASCII.

This change implements proper marshalling using UTF8.


bindings/cs/rl.net.native/rl.net.live_model.cc
```diff
@@ -65,6 +65,11 @@ API int LiveModelInit(livemodel_context_t* context, reinforcement_learning::api_
 
 API int LiveModelChooseRank(livemodel_context_t* context, const char * event_id, const char * context_json, reinforcement_learning::ranking_response* resp, reinforcement_learning::api_status* status)
 {
+	if (event_id == nullptr)
```
dwaijam: This check looks unnecessary

lokitoth: Ah, good catch. Under the check, it was chaining to the wrong method in the nullptr case. It is supposed to go to the auto-generate EventId code.

PR #60, opened 2019-04-05
Ataymano/logging improvement

ataymano:
Introducing debug logging for events tracking (what was sent / dropped)
Default logging: null -> console


rlclientlib/console_tracer.cc
```diff
@@ -3,7 +3,13 @@
 #include "str_util.h"
 
 namespace reinforcement_learning {
+  console_tracer::console_tracer(int _min_level) 
+    : min_level(_min_level) {
+  }
   void console_tracer::log(int log_level, const std::string& msg) {
-    std::cout << get_log_level_string(log_level) << ": " << msg << std::endl;
+    if (log_level >= min_level) {
+      std::unique_lock<std::mutex> lock(_mutex);
```
lokitoth: [Perf?] It worries me a little that we are grabbing locks when doing logging. What are the throughput consequences of this, especially since we are making this the default logger.

ataymano: The idea is that people are probably not using any library in high-performance scenario without testing in hello-worlds before, where it is more important to be user friendly than performant. Also, given the current config schema implementation, adding "trace.logger.implementation":"CONSOLE_TRACE_LOGGER" line seems impossible without guidance from devteam, which does not seem acceptable for getting-started scenarios (but ok for advanced modes).

lokitoth: I am not sure what you mean by:

> Also, given the current config schema implementation, adding "trace.logger.implementation":"CONSOLE_TRACE_LOGGER" line seems impossible without guidance from devteam

Why not just do how we do it in test:
config.set(r::name::MODEL_IMPLEMENTATION, r::value::PASSTHROUGH_PDF_MODEL);

Or do you mean that there is a discoverability issue?

PR #62, opened 2019-04-10
[http] Introduce two new config properties to modify http behavior an…

kumpera:
…d change the defaults.

Introduce http.timeout and http.certvalidation.

The first controls request timeout. Default value changes from 2 seconds to 30 seconds.
The second controls whether to check for server certificate errors. Default now is to always check.

There are few things to consider here.

- Whether we should be explicit about the default timeout or leave it up to the http client.
- If 30 seconds is a good or bad default

In addition to that, we should consider exposing the more http config knobs.


include/constants.h
```diff
@@ -34,6 +34,8 @@ namespace reinforcement_learning {  namespace name {
       const char *const  QUEUE_MODE = "queue.mode";
       const char *const  INTERACTION_FILE_NAME = "interaction.file.name";
       const char *const  OBSERVATION_FILE_NAME = "observation.file.name";
+      const char *const  HTTP_CLIENT_DISABLE_CERT_VALIDATION  = "http.certvalidation";
```
lokitoth: We should keep the negations of the constant name and the constant value similar. In other words, I would expect false to mean that we do cert validation when I read `HTTP_CLIENT_DISABLE_CERT_VALIDATION`, but the opposite when I read `http.certvalidation`.

include/constants.h
```diff
@@ -34,6 +34,8 @@ namespace reinforcement_learning {  namespace name {
...
+      const char *const  HTTP_CLIENT_TIMEOUT   = "http.timeout"; //timeout is in seconds, default is 30
```
jackgerrits: nit: comment style
```suggestion
      const char *const  HTTP_CLIENT_TIMEOUT   = "http.timeout"; // Timeout is in seconds, default is 30.
```

rlclientlib/utility/http_client.cc
```diff
@@ -4,9 +4,9 @@
 #include <chrono>
 
 namespace reinforcement_learning {
-  http_client::http_client(const char* url)
+  http_client::http_client(const char* url, const utility::configuration &cfg)
```
jackgerrits: nit: ref on type
```suggestion
  http_client::http_client(const char* url, const utility::configuration& cfg)
```

rlclientlib/utility/http_client.h
```diff
@@ -19,7 +20,7 @@ namespace reinforcement_learning {
 
   class http_client : public i_http_client {
   public:
-    http_client(const char* url);
+    http_client(const char* url, const utility::configuration &cfg);
```
jackgerrits: nit: ref on type
```suggestion
    http_client(const char* url, const utility::configuration& cfg);
```

rlclientlib/utility/http_helper.cc
```diff
@@ -1,15 +1,19 @@
 #include "http_helper.h"
+#include "constants.h"
 
 #include <chrono>
 
 namespace reinforcement_learning {
   namespace utility {
-    web::http::client::http_client_config get_http_config() {
+    web::http::client::http_client_config get_http_config(const utility::configuration &cfg) {
```
jackgerrits: nit: ref on typ
```suggestion
    web::http::client::http_client_config get_http_config(const utility::configuration& cfg) {
```

rlclientlib/utility/http_helper.cc
```diff
@@ -1,15 +1,19 @@
...
       web::http::client::http_client_config config;
 
-      config.set_validate_certificates(false);
-      // Set a timeout for network requests to avoid hanging too long because of network issues.
-      config.set_timeout(std::chrono::seconds(2));
+      config.set_validate_certificates(cfg.get_bool(name::HTTP_CLIENT_DISABLE_CERT_VALIDATION, false));
+      int timeout = cfg.get_int(name::HTTP_CLIENT_TIMEOUT, 30);
```
jackgerrits:
```suggestion
      const auto timeout = cfg.get_int(name::HTTP_CLIENT_TIMEOUT, 30);
```

rlclientlib/utility/http_helper.cc
```diff
@@ -1,15 +1,19 @@
...
+	    //Valid values are 1-30
+      if (timeout < 1 || timeout > 30)
+		    timeout = 30;
```
jackgerrits: tabs and spaces are mixed here

rlclientlib/utility/http_helper.h
```diff
@@ -1,8 +1,9 @@
 #pragma once
 #include <cpprest/http_client.h>
+#include "configuration.h"
 
 namespace reinforcement_learning { namespace utility {
 
-    web::http::client::http_client_config get_http_config();
+    web::http::client::http_client_config get_http_config(const utility::configuration &cfg);
```
jackgerrits:
```suggestion
    web::http::client::http_client_config get_http_config(const utility::configuration& cfg);
```

rlclientlib/utility/http_helper.cc
```diff
@@ -1,15 +1,19 @@
...
 namespace reinforcement_learning {
   namespace utility {
-    web::http::client::http_client_config get_http_config() {
+    web::http::client::http_client_config get_http_config(const utility::configuration &cfg) {
       web::http::client::http_client_config config;
 
-      config.set_validate_certificates(false);
-      // Set a timeout for network requests to avoid hanging too long because of network issues.
-      config.set_timeout(std::chrono::seconds(2));
+      config.set_validate_certificates(cfg.get_bool(name::HTTP_CLIENT_DISABLE_CERT_VALIDATION, false));
```
rajan-chari: Based on comments above, should this be HTTP_CLIENT_ENABLE_CERT_VALIDATION?

rajan-chari:
Looks good.  Would like to see response to HTTP_CLIENT_DISABLE_CERT_VALIDATION question/comments before merging.

rlclientlib/utility/http_helper.cc
kumpera: can't use const here as I change it down

rlclientlib/utility/http_helper.cc
kumpera: 🤦‍♂️

rlclientlib/utility/http_helper.cc
kumpera: No, it should be that, the logic in http_helper.cc is wrong

kumpera:
The last commit addresses the issues raised here.

PR #64, opened 2019-04-12
Adding the submodule update command to the readme to ensure everything is in sync.

QuinnDamerell-MS:
Currently the readme doesn't include a command to update the submodules in the cpprestsdk project after the checkout command. This should be in ran to make sure all of the submodules are in sync.


rajan-chari:
Merged.  Thanks

PR #67, opened 2019-04-17
Added client timestamp

rajan-chari:


rlclientlib/logger/event_logger.cc
```diff
@@ -1,15 +1,16 @@
...
 #include "ranking_event.h"
 #include "err_constants.h"
-
+#include "time_helper.h"
 namespace reinforcement_learning { namespace logger {
-  namespace u = utility;
 
   int interaction_logger::log(const char* event_id, const char* context, unsigned int flags, const ranking_response& response, api_status* status) {
-    return append(std::move(ranking_event::choose_rank(event_id, context, flags, response)), status);
+    const auto now = _time_provider != nullptr ? _time_provider->gmt_now() : timestamp();
```
ataymano: Isn't is better to have this logic inside default time_provider implementation? It reduces amount of copy-paste and allows us to make time_provider non-nullable, which seems to be safer.

rajan-chari: Yeah.  Good point.  I thought about that when implementing the null time provider factory.  That adds the cost of a virtual call to every rank/reward call.  Since we don't have a lot of places that need timestamps, I though this would be better.  I  am open to revisiting this in the future if this becomes an issue.

PR #73, opened 2019-05-08
Rajan/model file system loader

rajanjedi:
Load models from local file system.


rlclientlib/model_mgmt/file_model_loader.cc
```diff
@@ -0,0 +1,79 @@
...
+      if (!in_strm.good()) {
+        RETURN_ERROR_LS(_trace, status, file_open_error) << " file_name = " << _file_name;
+      }
+    }
+    return error_code::success;
+  }
+
+  int file_model_loader::get_data(model_data& data, api_status* status) {
+
+    std::ifstream in_strm(_file_name.c_str(), std::ios::in|std::ios::binary|std::ios::ate);
```
jackgerrits: How costly is opening the file? Perhaps we should check if it was modified prior to opening it?

ataymano: it is done on higher layer - see get_file_modified_time above,   which is called before

jackgerrits: But in the sequence of the get_data function, it is called after creating the ifstream object though?

ataymano: oh yes you right

yannstad:
Can this be merged?

rlclientlib/model_mgmt/file_model_loader.cc
rajan-chari: Does not happen in the hot path.   Not clear what the perf impact is.

rajan-chari:
@yannstad  Merged

PR #74, opened 2019-05-17
strong name signing changes

rajan-chari:
Changes to allow strong name signing from a trusted build source.


bindings/cs/rl.net/rl.net.csproj
```diff
@@ -26,4 +28,37 @@
...
+      <LastGenOutput>InternalsVisibleToTest.cs</LastGenOutput>
+    </None>
+    <Compile Include="InternalsVisibleToTest.cs" Exclude="@(Compile)">
+      <DesignTime>True</DesignTime>
+      <AutoGen>True</AutoGen>
+      <DependentUpon>InternalsVisibleToTest.tt</DependentUpon>
+    </Compile>
+  </ItemGroup>
+
+  <Target Name="OutputVars" BeforeTargets="Build">
```
lokitoth: Are these output vars still needed?

appveyor.yml
```diff
@@ -22,3 +22,6 @@ build_script:
   - .\build-windows.cmd
 test_script:
   - .scripts\test.cmd
+# Debug appveyor builds
```
lokitoth: Is this still needed?

bindings/cs/rl.net/rl.net.csproj
rajan-chari: Still have a number of steps to finish.  I'd like to merge this to release with this on.  Can get rid of if eventually.

appveyor.yml
rajan-chari: removed.

PR #77, opened 2019-05-24
Fix python bindings and update version to 0.1.2

yannstad:


ataymano:
Why do we have not latest version in docs?

yannstad:
> Why do we have not latest version in docs (0.1.3 vs 0.1.2)?

From now, the new current version is 0.1.2, and the wheel `rl_client-0.1.2-cp36-cp36m-linux_x86_64.whl` is publicly available.
So from now, if you generate new python bindings, the version will 0.1.3 as defined in setup.py.

PR #78, opened 2019-05-24
Provide troubleshooting instructions for OpenSSL on MacOS

jackgerrits:


kumpera:
This looks ok, but can't we search `brew` instead? https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/CMakeLists.txt#L112

jackgerrits:
We could, I looked into it briefly but would need to spend more time working out the right search path to put in. Merging this in the mean time as an improvement

PR #84, opened 2019-06-26
Conditional contextual bandit

### LINKED ISSUE #93: Add test for CCB no model exploration
It seems to be doing the incorrect thing in C# currently. Two 0 actions were witnessed

jackgerrits:
Fixes #93


rlclientlib/vw_model/vw_model.h
```diff
@@ -11,8 +11,10 @@ namespace reinforcement_learning { namespace model_management {
   class vw_model : public i_model {
   public:
     vw_model(i_trace* trace_logger);
+    vw_model(i_trace* trace_logger, std::string& initial_command_line);
     int update(const model_data& data, bool& model_ready, api_status* status = nullptr) override;
     int choose_rank(uint64_t rnd_seed, const char* features, std::vector<int>& action_ids, std::vector<float>& action_pdf, std::string& model_version, api_status* status = nullptr) override;
+    int request_decision(std::vector<const char*>& event_ids, const char* features, std::vector<std::vector<size_t>>& actions_ids, std::vector<std::vector<float>>& action_pdfs, std::string& model_version, api_status* status = nullptr) override;
```
ataymano: const vector<const char*>& event_ids?

rlclientlib/vw_model/vw_model.cc
```diff
@@ -11,10 +11,14 @@ namespace reinforcement_learning { namespace model_management {
     _vw_pool(nullptr), _trace_logger(trace_logger) {
   }
 
+  vw_model::vw_model(i_trace* trace_logger, std::string& initial_command_line) :
```
ataymano: const string&?

rlclientlib/sampling.cc
```diff
@@ -1,12 +1,50 @@
...
+    response.push_back(action_ids[idx], pdf[idx]);
+  }
+
+  RETURN_IF_FAIL(response.set_chosen_action_id(action_ids[chosen_action_index]));
+  response.set_model_id(std::move(model_id));
+  return error_code::success;
+}
+
+// TODO: write test for this
+int populate_response(std::vector<std::vector<size_t>>& action_ids, std::vector<std::vector<float>>& pdfs, std::vector<const char*>& event_ids, std::string&& model_id, decision_response& response, i_trace* trace_logger, api_status* status) {
```
ataymano: const action_ids, const pdfs

rlclientlib/ranking_event.h
```diff
@@ -46,13 +50,15 @@ namespace reinforcement_learning {
     const std::vector<float>& get_probabilities() const;
     const std::string& get_model_id() const;
     bool get_defered_action() const;
-  
+    const std::string& get_event_id() const {return get_seed_id();}
```
ataymano: It is weird that get_event_id is re-using get_seed_id - they accidentally share implementation right now, but there is no real dependency between them.

rlclientlib/ranking_event.cc
```diff
@@ -48,6 +46,40 @@ namespace reinforcement_learning {
     return ranking_event(event_id, flags & action_flags::DEFERRED, pass_prob, context, resp, ts);
   }
 
+  decision_ranking_event::decision_ranking_event() { }
+
+  decision_ranking_event::decision_ranking_event(std::vector<const char*> event_ids, bool deferred_action, float pass_prob, const char* context, const decision_response& response, const timestamp& ts)
```
ataymano: const vector&?

rlclientlib/ranking_event.cc
```diff
@@ -48,6 +46,40 @@ namespace reinforcement_learning {
...
+  }
+
+  const std::vector<unsigned char>& decision_ranking_event::get_context() const { return _context; }
+  const std::vector<std::vector<uint32_t>>& decision_ranking_event::get_actions_ids() const { return _action_ids_vector; }
+  const std::vector<std::vector<float>>& decision_ranking_event::get_probabilities() const { return _probilities_vector; }
+  const std::string& decision_ranking_event::get_model_id() const { return _model_id; }
+  bool decision_ranking_event::get_defered_action() const { return _deferred_action; }
+  const std::vector<std::string>& decision_ranking_event::get_event_ids() const { return _event_ids; }
+
+  decision_ranking_event decision_ranking_event::request_decision(std::vector<const char*> event_ids, const char* context, unsigned int flags,
```
ataymano: const vector&?

rlclientlib/utility/context_helper.cc
```diff
@@ -41,4 +44,64 @@ namespace reinforcement_learning { namespace utility {
       RETURN_ERROR_LS(trace, status, json_parse_error) << error_code::unknown_s;
     }
   }
+
+  int get_event_ids(std::map<size_t, std::string>& event_ids, const char *context, i_trace* trace, api_status* status) {
```
ataymano: Can we move output (and non-const) variables after input (and const) ones? It improves readability.

rlclientlib/live_model_impl.cc
```diff
@@ -83,6 +85,59 @@ namespace reinforcement_learning {
...
+  {
+    resp.clear();
+    //clear previous errors if any
+    api_status::try_clear(status);
+
+    //check arguments
+    RETURN_IF_FAIL(check_null_or_empty(context_json, status));
+
+    // Ensure multi comes before slots, this is a current limitation of the parser.
+    RETURN_IF_FAIL(utility::validate_multi_before_slots(context_json, nullptr, status));
```
ataymano: Why we do not want to provide trace_logger here (and few below calls - get_slot_count, get_event_ids)?

rlclientlib/ranking_event.h
jackgerrits: Spoke offline - the overload of this value is something to keep in mind as we continue to work on these objects

rlclientlib/live_model_impl.cc
jackgerrits: Good catch, thanks!

include/decision_response.h
```diff
@@ -0,0 +1,44 @@
...
+#include <vector>
+
+#include "ranking_response.h"
+
+namespace reinforcement_learning {
+  class api_status;
+
+  class decision_response {
+  public:
+    using coll_t = std::vector<ranking_response>;
```
rajan-chari: In ranking_response, we moved away from stl collections in the public interface due to concerns that collections that might be compiled with different flags.  This resulted in defining iterators independely of stl.

include/decision_response.h
```diff
@@ -0,0 +1,44 @@
...
+    using iterator_t = coll_t::iterator;
```
rajan-chari: ranking_response was a compromise for CB where we had to expose multiple actions although the underlying algorithm only dealt properly with one action.
With CCB now we can deal with multiple actions.  Given that, isn't a more natural interface collection of <actionid,eventid>?

include/live_model.h
```diff
@@ -136,7 +137,34 @@ namespace reinforcement_learning {
...
+    * per slot by sampling the probability distribution produced per slot. A unique event_id can be
+    * supplied for each slot using the `_id` json field. The corresponding event_id should be used
+    * when reporting the outcome for each slot.
+    * @param context_json Contains slots, slot_features, slot ids, actions, action features and context features in json format
+    * @param flags Action flags (see action_flags.h)
+    * @param resp Decision response contains the chosen action per slot, probability distribution used for sampling actions and ranked actions.
+    * @param status  Optional field with detailed string description if there is an error
+    * @return int Return error code.  This will also be returned in the api_status object
+    */
+    int request_decision(const char * context_json, unsigned int flags, decision_response& resp, api_status* status = nullptr);
```
rajan-chari: Given that we want to move to a FB based interface, should we indicate somehow that this interface is deprecated?  A more natural example_builder would be the right interface going forward.

include/live_model.h
```diff
@@ -136,7 +137,34 @@ namespace reinforcement_learning {
...
+    * action features, slot feautres and context features. The inference library chooses an action
+    * per slot by sampling the probability distribution produced per slot. A unique event_id can be
+    * supplied for each slot using the `_id` json field. The corresponding event_id should be used
+    * when reporting the outcome for each slot.
+    * @param context_json Contains slots, slot_features, slot ids, actions, action features and context features in json format
+    * @param resp Decision response contains the chosen action per slot, probability distribution used for sampling actions and ranked actions.
+    * @param status  Optional field with detailed string description if there is an error
+    * @return int Return error code.  This will also be returned in the api_status object
+    */
+    int request_decision(const char * context_json, decision_response& resp, api_status* status = nullptr);
```
rajan-chari: Deprecated?

include/live_model.h
```diff
@@ -153,7 +153,7 @@ namespace reinforcement_learning {
     int request_decision(const char * context_json, unsigned int flags, decision_response& resp, api_status* status = nullptr);
```
jackgerrits: Can you also add the deprecated attribute similar to what we do in VW? https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/vowpalwabbit/vw.h#L42

jackgerrits:
LGTM - thanks @ataymano, the container iterator looks good

bindings/cs/rl.net/DecisionResponse.cs
```diff
@@ -0,0 +1,190 @@
...
+namespace Rl.Net {
+    namespace Native
+    {
+        internal partial class NativeMethods
+        {
+            [DllImport("rl.net.native.dll")]
+            public static extern IntPtr CreateSlotResponse();
+
+            [DllImport("rl.net.native.dll")]
+            public static extern void DeleteSlotResponse(IntPtr slotResponse);
```
lokitoth: This is not used, since the NativeObject never owns the handle; we should consider removing it.

rlclientlib/factory_resolver.cc
```diff
@@ -68,6 +68,13 @@ namespace reinforcement_learning {
     return error_code::success;
   }
 
+  template <>
+  int model_create<m::vw_model>(m::i_model** retval, const u::configuration& c, i_trace* trace_logger, api_status* status)
```
lokitoth: Why do we need this explicit special-case? It looks identical to the generic template above it.

PR #88, opened 2019-07-16
Ataymano/empty events fix

ataymano:
No empty events and no concurrent prunings.


rlclientlib/logger/event_queue.h
```diff
@@ -52,6 +56,7 @@ namespace reinforcement_learning {
     void prune(float pass_prob)
     {
       std::unique_lock<std::mutex> mlock(_mutex);
+      if (!is_full()) return;
```
lokitoth: When you merge this, please make sure to explain the broken behaviour and the fix in the commit message.

PR #89, opened 2019-07-18
Ataymano/multiinstance events send

ataymano:
Custom reward ratio + Multiinstance mode for test_cpp


examples/test_cpp/main.cc
```diff
@@ -34,18 +36,29 @@ po::variables_map process_cmd_line(const int argc, char** argv) {
...
-    if (!loop.init()) {
-      std::cerr << "Test loop haven't initialized properly." << std::endl;
-      return -1;
+    const size_t instancees = vm["instances"].as<size_t>();
+    std::vector<std::thread> _threads;
+    for (size_t i = 0; i < instancees; ++i) {
+      _threads.push_back(std::thread(&run_test_instance, i, vm));
+    }
+    for (size_t i = 0; i < instancees; ++i) {
+      _threads[i].join();
```
lokitoth: If I understand correctly, what is going on here is that it will join into the execution of each thread, one-at-a-time. Does this impact the results in any way? In other words, does draining one-at-a-time, rather than all-together, change the scheduling behaviour of the threads?

examples/test_cpp/main.cc
```diff
@@ -34,18 +36,29 @@ po::variables_map process_cmd_line(const int argc, char** argv) {
...
 int main(int argc, char** argv) {
   try {
     const auto vm = process_cmd_line(argc, argv);
     if (is_help(vm)) return 0;
-
-    test_loop loop(vm);
-    if (!loop.init()) {
-      std::cerr << "Test loop haven't initialized properly." << std::endl;
-      return -1;
+    const size_t instancees = vm["instances"].as<size_t>();
```
lokitoth: [Minor][Nit] Typo

examples/test_cpp/main.cc
ataymano: no, it is not - they are running simultaneously for sure.

PR #110, opened 2019-10-10
Support ONNX Models via OnnxRuntime on Linux

lokitoth:
Adds support for CB-like ONNX models using OnnxRuntime on Linux

"CB-like" is defined as having at least one tensor output that will be interpreted as a PDF for the purposes of sampling and assigning a "chosen action".

The Context is provided through a JSON-like "TensorNotation". The format is:

```
Example:

{"input1":"BAAAAAAAAAA=;AACAP2ZmBkBmZoZAmpkRwQ=="}

This represents a rank-1 tensor "input1" of shape { 4 } with values { 1.0f, 2.1f, 4.2f, -9.1f }

Details:

<TENSORS> := "{" <TENSOR-LIST> "}"
<TENSOR-LIST> := <TENSOR> ["," [<TENSOR-LIST>]]
<TENSOR> := "\"" <INPUT-NAME> "\":\"" <TENSOR-DATA> "\""
<TENSOR-DATA> := <DIMS-BASE64> ";" <VALUES-BASE64>
<DIMS-BASE64> := { base64 encoding of int64[] representing dimensions of the tensor }
<VALUES-BASE64> := { base64 encoding of float[] representing values of the tensor }
```

No other JSON concepts are allowed. The reason for any relation to JSON is the current semantics of RLClientLib and associated tools expecting to log the context as JSON. Note that whitespace is not allowed.


cmake/Modules/FindOnnxRuntime.cmake
```diff
@@ -0,0 +1,20 @@
+#set(FLATBUFFERS_CMAKE_DIR ${CMAKE_CURRENT_LIST_DIR})
```
jackgerrits: Needed?

cmake/Modules/FindOnnxRuntime.cmake
```diff
@@ -0,0 +1,20 @@
...
+
+find_library(ONNXRUNTIME_LIB
```
jackgerrits: Onnxruntime does not supply a cmake configuration as part of the installation?

rlclientlib/CMakeLists.txt
```diff
@@ -112,6 +112,7 @@ target_include_directories( rlclientlib
                             PRIVATE
                             ${CMAKE_CURRENT_SOURCE_DIR}
                             ${FLATBUFFERS_INCLUDE_DIRS}
+                            ${CPPRESTSDK_INCLUDE_DIRS}
```
jackgerrits: This should not be needed, since it is a linked target on line 118

rlclientlib/extensions/onnx/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+  src/onnx_model.h
+  #src/onnxrt_type_mapping.h
+  src/tensor_notation.h
+)
+
+#source_group("Sources" FILES ${ONNX_EXTENSION_SOURCES})
+#source_group("Public headers" FILES ${ONNX_EXTENSION_PUBLIC_HEADERS})
+#source_group("Private headers" FILES ${ONNX_EXTENSION_HEADERS})
+
+add_library(rlclientlib-onnx STATIC ${ONNX_EXTENSION_SOURCES} ${ONNX_EXTENSION_PUBLIC_HEADERS} ${ONNX_EXTENSION_PRIVATE_HEADERS})
```
jackgerrits: Must it be static?

rlclientlib/extensions/onnx/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+
+set_target_properties(rlclientlib-onnx PROPERTIES POSITION_INDEPENDENT_CODE ON)
+
+add_dependencies(rlclientlib-onnx rlclientlib)
```
jackgerrits: No longer needed now you are linking against the target

rlclientlib/extensions/onnx/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+
+target_include_directories( rlclientlib-onnx
+                            PUBLIC
+                            ${CMAKE_CURRENT_SOURCE_DIR}/include
+                            PRIVATE
+                            $<TARGET_PROPERTY:rlclientlib,INTERFACE_INCLUDE_DIRECTORIES>
+                            ${CMAKE_CURRENT_SOURCE_DIR}/src
+                            ${ONNXRUNTIME_INCLUDE_DIRS}
+                            ${CPPRESTSDK_INCLUDE_DIRS}
```
jackgerrits: Link against target instead?

rlclientlib/extensions/onnx/src/onnx_model.cc
```diff
@@ -0,0 +1,206 @@
...
+// for base64 decoding
+#include <cpprest/http_client.h>
+
+namespace reinforcement_learning { namespace onnx {
+
+  inline void OrtLogCallback(
+    void* param, OrtLoggingLevel severity, const char* category, const char* logid, const char* code_location,
+    const char* message)
+  {
+    i_trace* trace_logger = (i_trace*)param;
```
jackgerrits: static cast here?

rlclientlib/extensions/onnx/src/onnx_model.cc
```diff
@@ -0,0 +1,206 @@
...
+    try {
+      TRACE_INFO(_trace_logger, utility::concat("Received new model data. With size ", data.data_sz()));
+	  
+      Ort::Session* new_session = nullptr;
+      if (data.data_sz() <= 0)
+      {
+        RETURN_ERROR_LS(_trace_logger, status, model_update_error) << "Empty model data.";
+      }
+
+      new_session =  new Ort::Session(_env, data.data(), data.data_sz(), _session_options);
```
jackgerrits:
```suggestion
      new_session = new Ort::Session(_env, data.data(), data.data_sz(), _session_options);
```

rlclientlib/extensions/onnx/src/tensor_notation.cc
```diff
@@ -0,0 +1,312 @@
...
+  if (_has_error) { \
+    return val; \
+  }
+//!define# PARSER_HEADER
+
+#define NO_ERROR_OR_RETURN_DEFAULT(val) \
+  if (_has_error) { \
+    return val; \
+  }
+//!define# PARSER_HEADER
```
jackgerrits: Are these meant to be the same?

rlclientlib/extensions/onnx/src/tensor_notation.cc
```diff
@@ -0,0 +1,312 @@
...
+
+  inline std::string error_message()
+  {
+    return _error_message;
+  }
+
+private:
+  inline void skip_whitespace()
+  {
+    NO_ERROR_OR_RETURN();
```
jackgerrits: Not sure how I feel about this error handling technique. Given that many portions of rlclientlib already throw, should this module just throw and ensure that it is caught and converted to error codes?

rlclientlib/extensions/onnx/src/tensor_notation.cc
```diff
@@ -0,0 +1,312 @@
...
+: _parse_buffer{tensor_notation}, 
+  _reading_head{tensor_notation}, 
+  _token_start{tensor_notation}, 
+  _parse_context{input_context},
+  _has_error{false},
+  _error_message{""}
+{
+  if (_parse_buffer && *_parse_buffer != tokens::NULL_TERMINATOR)
+  {
+    read_tensor_list();
```
jackgerrits: Why start off the parser in the constructor?

rlclientlib/extensions/onnx/src/tensor_notation.h
```diff
@@ -0,0 +1,136 @@
...
+
+#include <algorithm>
+#include <string>
+#include <vector>
+
+#include <core/session/onnxruntime_cxx_api.h>
+
+namespace reinforcement_learning { namespace onnx {
+  using bytes = std::vector<unsigned char>;
+  using tensor_data = std::pair<bytes, bytes>;
```
jackgerrits: Consider bytes_t and tensor_data_t?

rlclientlib/extensions/onnx/src/tensor_notation.h
```diff
@@ -0,0 +1,136 @@
...
+    inline size_t input_count() const
+    {
+      return _inputs.size();
+    }
+
+    inline std::vector<Ort::Value> inputs() const
+    {
+      // TODO: enable non-float tensors (see tensor_notation.cc for hook location)
+      int count = 0;
+      std::vector<Ort::Value> result;
```
jackgerrits: Can you reserve the size to minimize allocations?

rlclientlib/extensions/onnx/src/tensor_notation.h
```diff
@@ -0,0 +1,136 @@
...
+        size_t rank = dimensions_bytes.size() / sizeof(int64_t);
+        if (!check_array_size<int64_t>(dimensions_bytes, rank))
+        {
+          // TODO: error
+          break;
+        }
+
+        size_t expected_values_count = (rank > 0);
+        int64_t* dimensions = (int64_t*)dimensions_bytes.data();
+        for (int i = 0; i < rank; i++)
```
jackgerrits: std::accumulate instead?

rlclientlib/extensions/onnx/src/tensor_notation.h
```diff
@@ -0,0 +1,136 @@
...
+    {
+      _input_names.push_back(input_name);
+      _inputs.push_back(input);
+    }
+
+  private:
+    std::vector<std::string> _input_names{};
+    std::vector<tensor_data> _inputs{};
+
+    const Ort::AllocatorInfo& _allocator_info;
```
jackgerrits: prefer to not use const for class members, as it breaks swap and copy/move assignment/construction

rlclientlib/utility/config_utility.cc
```diff
@@ -97,12 +97,19 @@ namespace reinforcement_learning { namespace utility { namespace config {
...
+      { "sendHighMaterMark"         , name::INTERACTION_SEND_HIGH_WATER_MARK },
       { "SendBatchIntervalMs"       , name::INTERACTION_SEND_BATCH_INTERVAL_MS },
+      { "sendBatchIntervalMs"       , name::INTERACTION_SEND_BATCH_INTERVAL_MS },
       { "InitialExplorationEpsilon" , name::INITIAL_EPSILON },
+      { "initialExplorationEpsilon" , name::INITIAL_EPSILON },
       { "ModelRefreshIntervalMs"    , name::MODEL_REFRESH_INTERVAL_MS },
-      { "QueueMode"                 , name::QUEUE_MODE } // expect either DROP or BLOCK, default is DROP
+      { "modelRefreshIntervalMs"    , name::MODEL_REFRESH_INTERVAL_MS },
+      { "QueueMode"                 , name::QUEUE_MODE }, // expect either DROP or BLOCK, default is DROP
+      { "queueMode"                 , name::QUEUE_MODE } // expect either DROP or BLOCK, default is DROP
```
jackgerrits: Can this be a separate PR? This is super helpful

unit_test/extensions/onnx/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+          ${CMAKE_CURRENT_BINARY_DIR}/mnist_data
+)
+
+# Devirtualization is an optimization that changes the vtable if the compiler decides a function
+# doesn't need to be virtual. This is incompatible with the mocking framework used in testing as it
+# makes the vtable structure unpredictable
+if(MSVC)
+  # MSVC doesn't optimize aggresively enough for this bug to be an issue right now.
+else()
+  target_compile_options(rltest-onnx PRIVATE -fno-devirtualize)
```
jackgerrits: I wonder if gmock would be affected by this?

unit_test/extensions/onnx/main.cc
```diff
@@ -0,0 +1,18 @@
...
+
+struct GlobalConfig {
+  GlobalConfig()
+  {
+    reinforcement_learning::onnx::register_onnx_factory();
+  }
+
+  ~GlobalConfig()
+  {
+  }
```
jackgerrits: Use default destructor

unit_test/extensions/onnx/mnist_data/data_generator.py
```diff
@@ -0,0 +1,65 @@
+#%%
+
+from mnist import MNIST;
+import base64;
+import numpy as np;
```
jackgerrits: Unused import

unit_test/extensions/onnx/mnist_data/data_generator.py
```diff
@@ -0,0 +1,65 @@
...
+import struct;
+
+
+#%%
```
jackgerrits: What are these for?

unit_test/extensions/onnx/mock_helpers.cc
```diff
@@ -0,0 +1,30 @@
+#include "mock_helpers.h"
+
+#include "constants.h"
+
+namespace r = reinforcement_learning;
+namespace u = reinforcement_learning::utility;
+
+//using namespace fakeit;
+//using fakeit::Mock;
```
jackgerrits: remove?

rlclientlib/extensions/onnx/src/tensor_notation.cc
```diff
@@ -0,0 +1,312 @@
...
+    return result;
+  }
+
+  inline void set_error(std::string error_message)
+  {
+    _has_error = true;
+    _error_message = error_message;
+  }
+
+  template <char c>
```
ataymano: why does it have to template argument?

rlclientlib/extensions/onnx/src/tensor_notation.cc
```diff
@@ -0,0 +1,312 @@
...
+  {
+    std::stringstream result("Error parsing TensorNotation at position ");
+
+    // TODO: Error reporting here should be line/col based, rather than position-based.
+    result << (_reading_head - _parse_buffer) << ": ";
+
+    return result;
+  }
+
+  inline void set_error(std::string error_message)
```
ataymano: const std::string&?

rlclientlib/extensions/onnx/src/tensor_notation.h
```diff
@@ -0,0 +1,136 @@
...
+namespace reinforcement_learning { namespace onnx {
+  using bytes = std::vector<unsigned char>;
+  using tensor_data = std::pair<bytes, bytes>;
+
+  /**
+   * Check whether the provided bytes contain enough data to represent an
+   * array of elements of type element_t
+   */
+  template <typename element_t>
+  inline bool check_array_size(bytes bytes, size_t& element_count)
```
ataymano: const bytes&?

rlclientlib/extensions/onnx/src/tensor_notation.h
```diff
@@ -0,0 +1,136 @@
...
+    // array of elements of type element_t
+    return element_count * sizeof(element_t) == bytes.size();
+  }
+
+  /**
+   * Check whether the provided bytes contains exactly expected_element_count
+   * elements of type element_t
+   */
+  template <typename element_t>
+  inline bool check_array_size(bytes bytes, size_t expected_element_count, size_t& element_count)
```
ataymano: const  bytes&?

rlclientlib/extensions/onnx/src/tensor_notation.h
```diff
@@ -0,0 +1,136 @@
...
+      if (!succeeded)
+      {
+        return {};
+      }
+
+      return result;
+    }
+
+  public:
+    inline void push_input(std::string input_name, tensor_data input)
```
ataymano: const std::string&?

ataymano: const tensor_data&?

rlclientlib/vw_model/safe_vw.h
```diff
@@ -19,7 +19,7 @@ namespace reinforcement_learning {
   public:
     safe_vw(const std::shared_ptr<safe_vw>& master);
     safe_vw(const char* model_data, size_t len);
-	safe_vw(std::string vw_parameters);
+    safe_vw(std::string vw_parameters);
```
ataymano: const std::string&

rlclientlib/extensions/onnx/src/tensor_notation.h
lokitoth: Discussed offline - temporary solution is to disable moving OnnxRtInputContext.

unit_test/extensions/onnx/mnist_data/data_generator.py
lokitoth: These represent cell boundaries for the built-in "notebook-like" experience for Python Tools in VS Code

rlclientlib/extensions/onnx/src/tensor_notation.cc
lokitoth: No. There were two of them when I had a more complex expression in the return statement, but I forgot to merge them. Switched to exception internally.

rlclientlib/extensions/onnx/src/tensor_notation.cc
lokitoth: Because the semantics were "parsing can happen once" per Parser instance. This indicates a flawed abstraction, in particular that we pass the OnnxRtInputContext into the parser (to parallel what happens with VW), but it seems like a better design would be to have ORIC to drive the parser inside of its constructor, or a factory method, or similar.

For now pulled this out into `cached_parse()`. I will try to rework this a bit when I give ExampleBuilder another go.

rlclientlib/extensions/onnx/src/tensor_notation.h
```diff
@@ -0,0 +1,137 @@
+#pragma once
+
+#include "api_status.h"
+
+#include <algorithm>
```
ataymano:
```suggestion
#include <algorithm>
#include <numeric>
```

ataymano: Because of std::accumulate (https://en.cppreference.com/w/cpp/algorithm/accumulate).
Required for build in 18.04

cmake/Modules/FindOnnxRuntime.cmake
```diff
@@ -0,0 +1,18 @@
...
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(onnxruntime
+  DEFAULT_MSG ONNXRUNTIME_LIB ONNXRUNTIME_INCLUDE_DIR_ROOT)
+
+if(ONNXRUNTIME_FOUND)
+  add_library(onnxruntime SHARED IMPORTED)
+  set_target_properties(onnxruntime PROPERTIES IMPORTED_LOCATION ${ONNXRUNTIME_LIB})
+
+  set(ONNXRUNTIME_INCLUDE_DIRS ${ONNXRUNTIME_INCLUDE_DIR_ROOT}/include/onnxruntime)
```
jackgerrits: The preferred way of doing this is setting the include dirs on the onnxruntime target itself

rlclientlib/extensions/onnx/src/onnx_extension.cc
```diff
@@ -0,0 +1,38 @@
+
+#include "model_mgmt.h"
+#include "configuration.h"
+#include "factory_resolver.h"
+#include "constants.h"
+
+#include "onnx_extension.h"
```
jackgerrits: Nit: I prefer to have the header that this source file corresponds to as the first line of the cc

rlclientlib/extensions/onnx/src/onnx_extension.cc
```diff
@@ -0,0 +1,38 @@
...
+    const char* output_name = config.get(name::ONNX_OUTPUT_NAME, nullptr);
+    if (output_name == nullptr)
+    {
+      RETURN_ERROR_LS(trace_logger, status, inference_configuration_error) << "Output name is not provided in the configuration.";
+    }
+
+    int onnx_threadpool_size = config.get_int(name::ONNX_THREADPOOL_SIZE, 10);
+    bool parse_feature_string = config.get_bool(name::ONNX_PARSE_FEATURE_STRING, true);
+  
+    *retval = new onnx_model(trace_logger, app_id, output_name, onnx_threadpool_size, parse_feature_string);
```
jackgerrits: Does init need to be called?

rlclientlib/extensions/onnx/src/onnx_model.cc
```diff
@@ -0,0 +1,204 @@
...
+      case ORT_LOGGING_LEVEL_FATAL: // TODO: Should this be a background error?
+      case ORT_LOGGING_LEVEL_ERROR:
+        loglevel = LEVEL_ERROR;
+      break;
+    }
+
+    std::stringstream buf;
+    buf << "[onnxruntime, modelid=" << logid << "]: " << message;
+
+    TRACE_LOG(trace_logger, loglevel, buf.str());
```
jackgerrits: If trace log is not being used we should not pay the cost fo the stringstream above

rlclientlib/extensions/onnx/src/onnx_model.cc
```diff
@@ -0,0 +1,204 @@
...
+
+  int onnx_model::choose_rank(uint64_t rnd_seed, 
+    const char* features, 
+    std::vector<int>& action_ids, 
+    std::vector<float>& action_pdf, 
+    std::string& model_version,
+    api_status* status)
+  {
+    std::shared_ptr<Ort::Session> local_session = _master_session;
+    if (!(bool(local_session)))
```
jackgerrits: Prefer static cast over c_style cast

rlclientlib/extensions/onnx/src/onnx_model.cc
```diff
@@ -0,0 +1,204 @@
...
+
+    std::vector<const char*> input_names = input_context.input_names();
+    std::vector<Ort::Value> inputs = input_context.inputs();
+    if (inputs.size() != input_context.input_count())
+    {
+      // TODO: propagate errors to be more specific about which input(s) is/are bad
+      RETURN_ERROR_LS(_trace_logger, status, model_rank_error) << "Could not interpret input values to match expected inputs.";
+    }
+
+    auto outputs = local_session->Run(Ort::RunOptions{nullptr}, input_names.data(), inputs.data(), input_context.input_count(), (const char* const*)&_output_name, 1);
```
jackgerrits: Can this fail?

unit_test/extensions/onnx/CMakeLists.txt
jackgerrits: As of this PR the flag shouldn't be needed https://github.com/VowpalWabbit/reinforcement_learning/pull/114

rlclientlib/extensions/onnx/src/tensor_parser.cc
```diff
@@ -0,0 +1,402 @@
...
+{
+  enum CONSUME_TYPE : bool { EXCLUSIVE = false, INCLUSIVE = true };
+  enum KNOWN_CHARS : char { END = '\0', DOUBLE_QUOTE = '\"', SEMICOLON = ';', COMMA = ',', OPEN_BRACE = '{', CLOSE_BRACE = '}', BACKSLASH = '\\', COLON = ':' };
+
+  namespace errors
+  {
+    struct error_context
+    {
+    private:
+      std::vector<std::string>& errors;
```
peterychang: nit: See previous comments about ref members and consts

rlclientlib/extensions/onnx/src/tensor_parser.cc
```diff
@@ -0,0 +1,402 @@
...
+      const std::string prefix;
+      tensor_parser::parser_context& parser_context;
+
+    public:
+      inline error_context(std::vector<std::string>& target, const std::string prefix, tensor_parser::parser_context& parser_context)
+        : errors(target), prefix(prefix), parser_context(parser_context)
+      {
+      }
+
+      inline error_context with_prefix(const std::string prefix)
```
peterychang: `const std::string&`

rlclientlib/extensions/onnx/src/tensor_parser.cc
```diff
@@ -0,0 +1,402 @@
...
+  {
+    struct error_context
+    {
+    private:
+      std::vector<std::string>& errors;
+      const std::string prefix;
+      tensor_parser::parser_context& parser_context;
+
+    public:
+      inline error_context(std::vector<std::string>& target, const std::string prefix, tensor_parser::parser_context& parser_context)
```
peterychang: `const std::string&`

rlclientlib/extensions/onnx/src/tensor_parser.cc
```diff
@@ -0,0 +1,402 @@
...
+        : errors(target), prefix(prefix), parser_context(parser_context)
+      {
+      }
+
+      inline error_context with_prefix(const std::string prefix)
+      {
+        return error_context(errors, prefix, parser_context);
+      }
+
+      inline bool append_error(std::string detail)
```
peterychang: `const std::string&`

rlclientlib/extensions/onnx/src/tensor_parser.cc
peterychang: Also, can we prefix these variable names with an underscore?

rlclientlib/extensions/onnx/src/tensor_parser.cc
```diff
@@ -0,0 +1,402 @@
...
+      std::string& value;
+
+    public:
+      parse_context(std::string& target) : value(target), in_escape(false)
+      {
+      }
+
+      inline bool invoke(const char c)
+      {
+        if (in_escape || c != escape)
```
peterychang: nit: swap the order of these checks

rlclientlib/extensions/onnx/src/tensor_parser.cc
```diff
@@ -0,0 +1,402 @@
...
+        {
+          value.push_back(c);
+          in_escape = false;
+        }
+
+        if (c == escape)
```
peterychang: i think this should be an else if?

rlclientlib/extensions/onnx/src/tensor_parser.cc
```diff
@@ -0,0 +1,402 @@
...
+  {
+    if (context._parsed)
+    {
+      return false;
+    }
+
+    const char*& reading_head = context._reading_head;
+
+    // '{' <tensor_name_value> [ ',' <tensor_name_value> ]* 
+    if (!consume_exact<OPEN_BRACE>(reading_head))
```
peterychang: It seems like improper whitespace is an error condition?

rlclientlib/utility/config_utility.cc
peterychang: Would it make sense to just `to_lower()` the keys so we don't need extra entries for capitalization differences?

rlclientlib/extensions/onnx/src/onnx_model.h
```diff
@@ -0,0 +1,34 @@
...
+#include <core/session/onnxruntime_cxx_api.h>
+
+namespace reinforcement_learning {
+  class i_trace;
+}
+
+namespace reinforcement_learning { namespace onnx {
+  class onnx_model : public model_management::i_model {
+  public:
+    onnx_model(onnx_model&&) = delete;
```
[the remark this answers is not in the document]
peterychang: nit: Also, rule of 5. Add defaults for constructor, destructor, copy constructor/assignment as well

rlclientlib/extensions/onnx/src/onnx_model.cc
```diff
@@ -0,0 +1,200 @@
...
+#include "trace_logger.h"
+#include "err_constants.h"
+#include "onnx_model.h"
+#include "str_util.h"
+#include "api_status.h"
+
+#include "factory_resolver.h"
+
+// for base64 decoding
+#include <cpprest/http_client.h>
```
peterychang: Remove this

rlclientlib/extensions/onnx/src/onnx_extension.cc
lokitoth: No - it gets called when live_model::init() gets called, like for the other models.

rlclientlib/extensions/onnx/src/tensor_parser.cc
lokitoth: Yes, this is correct. Right now we do not support improper (any) whitespace.

rlclientlib/utility/config_utility.cc
lokitoth: Turns out making this work with `to_lower()` is a bit challenging because we need to be able to deal with UTF-8 strings, and I do not want to delay this PR for it.

rlclientlib/extensions/onnx/src/tensor_notation.cc
```diff
@@ -0,0 +1,335 @@
...
+    bool in_escape = false;
+    while (true)
+    {
+      switch (*_reading_head)
+      {
+        case tokens::SQUOTE:
+          if (in_escape) break;
+          // If not in an escape sequence, terminate scanning for the name
+        case tokens::NULL_TERMINATOR:
+          goto end_name_scan;
```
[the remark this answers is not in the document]
lokitoth: The rewrite to avoid exceptions removed the need for this.

rlclientlib/extensions/onnx/src/onnx_extension.cc
```diff
@@ -0,0 +1,37 @@
...
+    const char* app_id = config.get(name::APP_ID, "");
+    const char* output_name = config.get(name::ONNX_OUTPUT_NAME, nullptr);
+    if (output_name == nullptr)
+    {
+      RETURN_ERROR_LS(trace_logger, status, inference_configuration_error) << "Output name is not provided in the configuration.";
+    }
+
+    bool parse_feature_string = config.get_bool(name::ONNX_PARSE_FEATURE_STRING, true);
+  
+    *retval = new onnx_model(trace_logger, app_id, output_name, parse_feature_string);
```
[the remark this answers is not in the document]
lokitoth: No. This behaves the same way as other model factories, see [factory_resolver.cc#L68-L73](https://github.com/VowpalWabbit/reinforcement_learning/blob/master/rlclientlib/factory_resolver.cc#L68-L73)

rlclientlib/extensions/onnx/src/onnx_input.cc
```diff
@@ -0,0 +1,101 @@
+#include "onnx_input.h"
+#include "err_constants.h"
+#include "api_status.h"
+#include "tensor_parser.h"
+
+#include <sstream>
+#include <assert.h>
```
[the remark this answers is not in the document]
lokitoth: We are not, good catch.

rlclientlib/extensions/onnx/src/onnx_input.cc
```diff
@@ -0,0 +1,101 @@
...
+      break;
+    }
+
+    value_t* values = (value_t*)values_bytes.data();
+    result.push_back(std::move(Ort::Value::CreateTensor<value_t>(this->_memory_info, values, values_count, dimensions, rank)));
+
+    succeeded = true;
+  }
+
+  if (!succeeded)
```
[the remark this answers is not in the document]
lokitoth: Good catch. Fixed.

rlclientlib/extensions/onnx/src/tensor_parser.cc
lokitoth: Yes. Good catch.

.scripts/linux/build.sh
```diff
@@ -10,6 +10,6 @@ cd $REPO_DIR
 # Build reinforcement_learning
 mkdir -p build
 cd build
-cmake ..
+cmake .. -Drlclientlib_BUILD_ONNXRUNTIME_EXTENSION=On
```
lalo, line 13: should we have two separate ci's building with and without?

lokitoth: Good idea, but would prefer to avoid expanding the scope of this PR (and there may a bit of a chicken-and-egg problem with setting it up so that it works as part of a PR that includes this change)

rlclientlib/extensions/onnx/src/onnx_model.cc
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  int onnx_model::choose_rank(uint64_t rnd_seed, 
+    const char* features, 
+    std::vector<int>& action_ids, 
+    std::vector<float>& action_pdf, 
+    std::string& model_version,
+    api_status* status)
+  {
+    std::shared_ptr<Ort::Session> local_session = _master_session;
```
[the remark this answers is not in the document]
lokitoth: This prevents race conditions between the inference run and cleanup of the `Ort::Session` - this can happen during a model reload, for example.

rlclientlib/extensions/onnx/src/onnx_input.cc
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    value_t* values = (value_t*)values_bytes.data();
+    result.push_back(std::move(Ort::Value::CreateTensor<value_t>(this->_memory_info, values, values_count, dimensions, rank)));
+  }
+
+  if (failed)
+  {
+    // TODO: report error
+    return {};
```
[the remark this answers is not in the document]
lokitoth: Hmm, I seem to have lost a commit somewhere during rebase. :-(

rlclientlib/extensions/onnx/src/onnx_input.h
```diff
@@ -0,0 +1,79 @@
...
+  using tensor_data_t = std::pair<bytes_t, bytes_t>;
+
+  /**
+   * Check whether the provided bytes contain enough data to represent an
+   * array of elements of type element_t
+   */
+  template <typename element_t>
+  inline bool check_array_size(const bytes_t& bytes, size_t& element_count)
+  {
+    element_count = bytes.size() / sizeof(element_t);
```
ataymano, line 22: shouldn't we have bytes.size() * sizeof(byte_t) instead of bytes.size() here? (probably not critical right now)

rlclientlib/extensions/onnx/src/onnx_input.h
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  /**
+   * Check whether the provided bytes contains exactly expected_element_count
+   * elements of type element_t
+   */
+  template <typename element_t>
+  inline bool check_array_size(const bytes_t& bytes, size_t expected_element_count, size_t& element_count)
+  {
+    check_array_size<element_t>(bytes, element_count);
```
ataymano, line 36: don't we care if it returns false here?

rlclientlib/extensions/onnx/src/onnx_model.cc
```diff
@@ -0,0 +1,221 @@
...
+      // Validate that the model makes sense
+      // Rules: 
+      // 1. There are N inputs, which are all tensors of floats
+      // 2. There is an output with the provided name, which is a tensor of floats
+
+      size_t input_count = new_session->GetInputCount();
+      for (int i = 0; i < input_count; i++)
+      {
+        // TODO: Support more input types (by making the input interface richer)
+        Ort::TypeInfo input_type_info = new_session->GetInputTypeInfo(i);
```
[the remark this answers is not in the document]
lokitoth: No. Ort::TypeInfo is an owning container around the underlying C handle. The consumer of the API is responsible for deallocating it.

rlclientlib/extensions/onnx/src/onnx_input.h
lokitoth: Not critical, but might as well make it safe.

rlclientlib/extensions/onnx/src/onnx_input.h
lokitoth: We do, good catch.

include/future_compat.h
```diff
@@ -2,13 +2,8 @@
...
 
-#if __cplusplus >= 201402L || defined(_MSC_VER) && (_MSC_VER >= 1910) && (_MSVC_LANG >= 201402L)
-#define HAS_STD14
-#endif
-
-#if __cplusplus >= 201703L  || defined(_MSC_VER) && (_MSC_VER >= 1914) && (_MSVC_LANG >= 201703L)
-#define HAS_STD17
-#endif
+// Rely on the VW implementation of future_compat to set up HAS_STD(NN)
+#include "../ext_libs/vowpal_wabbit/explore/future_compat.h"
```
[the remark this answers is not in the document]
lokitoth: Discussed offline; I was running into issues here around tooling and Windows when not using a relative path. The solution here will be to remove the RL version of future_compat.h and replatform reinforcement_learning to use the version in vowpal_wabbit, but like with the CI suggestion below, it probably makes sense to do that as a separate PR.

rlclientlib/extensions/onnx/src/onnx_model.cc
```diff
@@ -0,0 +1,200 @@
...
+    try {
+      TRACE_INFO(_trace_logger, utility::concat("Received new model data. With size ", data.data_sz()));
+	  
+      Ort::Session* new_session = nullptr;
+      if (data.data_sz() <= 0)
+      {
+        RETURN_ERROR_LS(_trace_logger, status, model_update_error) << "Empty model data.";
+      }
+
+      new_session = new Ort::Session(_env, data.data(), data.data_sz(), _session_options);
```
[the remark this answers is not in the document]
lokitoth: Yes, that is a good catch and idea.

rlclientlib/extensions/onnx/src/onnx_model.h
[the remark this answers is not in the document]
lokitoth: I do not recall the exact reason why they were added, but does not seem relevant with current implementation. Removing these and using defaults for everything.

rlclientlib/extensions/onnx/src/tensor_parser.cc
lokitoth: I am using references here to ensure that these are always non-null. Would prefer to avoid changing it so I do not need to add nullchecks throughout.

PR #115, opened 2019-10-29
Fix openssl home on macos

jackgerrits:


jackgerrits:
MacOS CI is blocked on Cpprestsdk fixing the homebrew install as tracked by this issue: https://github.com/microsoft/cpprestsdk/issues/1264

The cause is explained in this issue: https://gitlab.kitware.com/cmake/cmake/issues/19824

PR #122, opened 2019-12-23
Ataymano/vw88 upgrade

ataymano:
VW 8.7 -> 8.8
Boost 1.63 -> 1.70
ToolsVersion 14 -> 15
Toolset 140 -> 141

### DIFF
--- a/rlclientlib/CMakeLists.txt
+++ b/rlclientlib/CMakeLists.txt
@@ -120,6 +120,7 @@ target_include_directories( rlclientlib
                             ${CMAKE_CURRENT_SOURCE_DIR}
                             ${FLATBUFFERS_INCLUDE_DIRS}
                             ${CMAKE_CURRENT_SOURCE_DIR}/../ext_libs/date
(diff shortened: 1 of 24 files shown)


rlclientlib/CMakeLists.txt
```diff
@@ -120,6 +120,7 @@ target_include_directories( rlclientlib
                             ${CMAKE_CURRENT_SOURCE_DIR}
                             ${FLATBUFFERS_INCLUDE_DIRS}
                             ${CMAKE_CURRENT_SOURCE_DIR}/../ext_libs/date
+                            ${CMAKE_CURRENT_SOURCE_DIR}/../ext_libs/vowpal_wabbit/rapidjson/include
```
jackgerrits: This should not be necessary. Where does rlclientlib require the rapidjson headers?

ataymano: agree it looks like hack.
safe_vw.cc -> parse_example_json.h -> rapidjson

PR #123, opened 2020-01-06
Create safe_vw shared instance on initialization of safe vw factory.

slahabar:
Currently, if the model data is corrupt, safe vw factory throws an exception when trying to create the safe vw instance. 

Now, moving the user input validation in the constructor by creating a shared safe_vw instance on safe vw factory initialization method. Shared pointer will get auto deleted when no longer in use by the vw pool.

### DIFF
--- a/rlclientlib/vw_model/safe_vw.cc
+++ b/rlclientlib/vw_model/safe_vw.cc
@@ -251,27 +251,23 @@ bool safe_vw::is_compatible(const std::string& args) const {
 }
 
 safe_vw_factory::safe_vw_factory(const std::string& command_line)
-  : _command_line(command_line)
-{}
+{
+	_safe_vw = std::make_shared<safe_vw>(command_line);
+}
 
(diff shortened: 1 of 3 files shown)


jackgerrits:
So we originally had this code similar to how you have it now, but there are issues when using a seeded model (what the constructor being used here does) and marginal. That's why it was changed to creating new entire vw instances from the model data. See here for the commit where this was originally mitigated: https://github.com/VowpalWabbit/reinforcement_learning/commit/e159effa006a019d127ad1dd4e281d40b216650b.

Can the model data be tested for corruption when the factory is given the updated payload?

slahabar:
Thanks Jack for the feedback. Right now the issue is already mitigated as a side effect of the ccb PR. It creates an instance to check ML args compatibility . I started working on this before that PR was merged and wanted to get the idea out. I don't understand what marginal issue is. I'll message you to discuss about that.

PR #129, opened 2020-02-26
Introduce imitation mode in Rank call

homezcx:
Introduce imitation mode in Rank call:

**Goal**
Help customers adopt model produced by VowpalWabbit, by de-risking an assessment of “if” and “how long it will take” for the model to match the business performance (rewards) of the current business logic. 

**How it works**
The feature will be able to turn on/off through configuration file "config.json" by adding / removing "LearningMode: "IMITATION""

**Change in this PR**
The imitation mode is assuming customers always put baseline action as action[0]. Regardless of what VowpalWabbit returns for Rank action and probability array, it will always be sorted by the order of action index. The suggestion will be always action[0].


include/constants.h
```diff
@@ -47,6 +47,7 @@ namespace reinforcement_learning {  namespace name {
       const char *const  DECISION_SEND_QUEUE_MAX_CAPACITY_KB    = "decisions.send.queue.maxcapacity.kb";
       const char *const  DECISION_SEND_BATCH_INTERVAL_MS   = "decisions.send.batchintervalms";
       const char *const  DECISION_SENDER_IMPLEMENTATION    = "decisions.sender.implementation";
+      const char *const  DECISION_RANK_MODE = "decisions.rank.mode";
```
yannstad: Should we fix the naming consistency in a separate PR?
* `DECISION_FOO_BAR = decisions.foo.bar` => `DECISION_FOO_BAR = decision.foo.bar`
* `maxcapacity` => `max.capacity`
* `batchintervalms` => `batch.interval.ms`

include/constants.h
```diff
@@ -75,6 +76,7 @@ namespace reinforcement_learning {  namespace value {
       const char *const CONSOLE_TRACE_LOGGER = "CONSOLE_TRACE_LOGGER";
       const char *const NULL_TIME_PROVIDER = "NULL_TIME_PROVIDER";
       const char *const CLOCK_TIME_PROVIDER = "CLOCK_TIME_PROVIDER";
+      const char *const DECISION_RANK_MODE = "DEFAULT";
```
yannstad: `DEFAULT` is very vague. Can you find a self descriptive name ?

include/decision_modes.h
```diff
@@ -0,0 +1,15 @@
...
+* @brief decision_modes definition.
+*
+* @file decision_modes.h
+* @author Chenxi Zhao
+* @date 2020-02-13
+*/
+#pragma once
+
+namespace reinforcement_learning {
+  enum decision_modes {
```
yannstad: `decision_modes` => `decision_mode` ?

rlclientlib/schema/DecisionMode.fbs
```diff
@@ -0,0 +1,16 @@
+namespace reinforcement_learning.messages.flatbuff;
+
+union Mode { DefaultMode, ImitationMode }
+
+table DefaultMode {
```
yannstad: I believe we can find something more descriptive than `DefaultMode`.

include/constants.h
ataymano: Also, there is a mismatch in terminology right now.
In client library:
interactions - CB interactions.
decisions - CCB interactions.
Given that your change is only about CB now, renaming "*decision*" to "*interaction*" looks more consistent

homezcx: Renamed them to "Interaction" at this moment.

include/constants.h
homezcx: I talked about it with Ed. Using "Online" to indicate the default state at this moment.

include/decision_modes.h
yannstad: Not sure it is a pattern. For instance there is also [`queue_mode_enum`](https://github.com/VowpalWabbit/reinforcement_learning/blob/6f3ac9044bd0fdea5d148eb31bccd5101a5987dc/rlclientlib/logger/async_batcher.h#L17.

homezcx: Fixed.

rlclientlib/schema/DecisionMode.fbs
homezcx: Fixed. Using OnlineMode

rlclientlib/live_model_impl.cc
```diff
@@ -69,13 +71,15 @@ namespace reinforcement_learning {
...
     response.set_event_id(event_id);
-    RETURN_IF_FAIL(_ranking_logger->log(event_id, context, flags, response, status));
+    RETURN_IF_FAIL(_ranking_logger->log(event_id, context, flags, response, status, _decision_mode));
 
     // Check watchdog for any background errors. Do this at the end of function so that the work is still done.
     if (_watchdog.has_background_error_been_reported()) {
       RETURN_ERROR_LS(_trace_logger.get(), status, unhandled_background_error_occurred);
     }
 
+    RETURN_IF_FAIL(post_process_rank(response, _decision_mode));
```
rajan-chari: This should be moved before _watchdog.  _watchdog errors are the lowest priority errors.

rlclientlib/live_model_impl.h
```diff
@@ -73,6 +74,7 @@ namespace reinforcement_learning
     error_callback_fn _error_cb;
     model_management::data_callback_fn _data_cb;
     utility::watchdog _watchdog;
+    decision_mode _decision_mode;
```
rajan-chari: 1) Can we make this a bool like _bg_model_proc(essing)?
Making this an enum implies there are many behaviors.

2) The name decision_mode seems non specific.  Some thing more specific to describe the behavior would be better.  Will review the rest of the code to see if I have a better suggestion.

rajan-chari:
>     resp.clear();

What is the change in behavior in the CCB case?

---
Refers to: rlclientlib/live_model_impl.cc:94 in 36cc99e. [](commit_id = 36cc99e2c0277a987dacc38447911951b2798938, deletion_comment = False)

rlclientlib/live_model_impl.cc
```diff
@@ -417,4 +423,35 @@ namespace reinforcement_learning {
     return error_code::success;
   }
 
+  int post_process_rank(ranking_response& response, decision_mode decision_mode) {
+    if (decision_mode == IMITATION_MODE) {
```
rajan-chari: I prefer a bool here but if its not, the enum should be in a switch and non explicitly handled cases should return error.

rlclientlib/live_model_impl.cc
```diff
@@ -417,4 +423,35 @@ namespace reinforcement_learning {
...
+      std::vector<action_prob> copied_action_prob;
```
rajan-chari: 1) This will allocate memory for every call.  
2) it also extra copies of memory
response already has a vector.  please reuse this instead.

rlclientlib/live_model_impl.cc
rajan-chari: >post_process_rank [](start = 19, length = 17)

Trainer need to train using following tuple for unbiased training
[a,p(a),c(a),x]
a = user chosen
p(a) = 1
c(a)=feedback from env
x = features

In this case why is local inference even needed?

rlclientlib/schema/DecisionMode.fbs
```diff
@@ -0,0 +1,16 @@
...
+union Mode { OnlineMode, ImitationMode }
+
+table OnlineMode {
+}
+
+table ImitationMode {
+}
+
+table DecisionMode {
+	type:string;                      // name of the learning mode
```
rajan-chari: Why is this needed?

rlclientlib/serialization/fb_serializer.h
```diff
@@ -35,9 +35,28 @@ namespace reinforcement_learning { namespace logger {
...
+        break;
+      }
+      case ONLINE_MODE: {
+        const auto default_mode = CreateOnlineMode(builder).Union();
+        const auto default_mode_name = builder.CreateString("online_mode");
+        decision_mode_offset = CreateDecisionMode(builder, default_mode_name, Mode_OnlineMode, default_mode);
+        break;
+      }
+      default:
+        break;
```
rajan-chari: Is this not an error?

rlclientlib/live_model_impl.h
homezcx: From what I heard from Ed, more modes may come (e.g. logging mode). So a boolean may not be suffient for the design.

rlclientlib/schema/RankingEvent.fbs
```diff
@@ -11,7 +13,8 @@ table RankingEvent {
     probabilities:[float];    // probabilities
     model_id:string;          // model ID
     pass_probability:float;	  // Probability of event surviving throttling operation
-	meta:Metadata;
+    meta:Metadata;
+    decision_mode:DecisionMode; // decision mode used to determine rank behavior
```
rajan-chari: For reasons of backward/forward compatibility, 

1) use a uint/ubyte here.
2)  declare an enum for the value:

example: 
enum DecisionModeValues : ubyte { Online, Imitation }

This allows one to 
1) use enums in code 
2) enums are not enforced during deserialization so will not cause an error.
3) do proper min, max values checks in code and report error when deserialized value is outside min/max range

rlclientlib/schema/DecisionMode.fbs
```diff
@@ -0,0 +1,16 @@
+namespace reinforcement_learning.messages.flatbuff;
+
+union Mode { OnlineMode, ImitationMode }
```
rajan-chari: Instead of empty structs use enum + int.  
Please check comment in RankingEvent.fbs

homezcx:
>     resp.clear();

At this moment, CCB is not supported. There is one ongoing discussion among John, Marco, Paul - Definition of CCB baseline1

---
In reply to: [592561744](https://github.com/VowpalWabbit/reinforcement_learning/pull/129#issuecomment-592561744) [](ancestors = 592561744)

---
Refers to: rlclientlib/live_model_impl.cc:94 in 36cc99e. [](commit_id = 36cc99e2c0277a987dacc38447911951b2798938, deletion_comment = False)

rlclientlib/live_model_impl.cc
homezcx: The problem there is that response's vector has a constant iterator - I cannot shift the order unless I clear them.

Checking to remove the constant iterator but need to confirm the reason when adding it



---
In reply to: [385784544](https://github.com/VowpalWabbit/reinforcement_learning/pull/129#discussion_r385784544) [](ancestors = 385784544)

rlclientlib/live_model_impl.cc
```diff
@@ -417,4 +423,35 @@ namespace reinforcement_learning {
...
+      for (auto it = copied_action_prob.begin(); it != copied_action_prob.end(); ++it) {
+        action_prob cur = *it;
+        response.push_back(cur.action_id, cur.probability);
+      }
+    }
+    return error_code::success;
+  }
+
+  decision_mode to_decision_modes(const char* decision_mode) {
+    if (std::strcmp(decision_mode, "IMITATION_MODE") == 0) {
```
rajan-chari: >IMITATION_MODE [](start = 36, length = 14)

move this to constants.h?

rlclientlib/live_model_impl.cc
homezcx: As discussed offline, VW could not learn from 0 probability. This is by design.

rlclientlib/live_model_impl.h
rajan-chari: should we call this a bool flag "logging_only"  because that's what is happening ... no?  The config file generation can set this as a fine grained behavior desired from rlclientlib.  Then we can mix and match fine grained behaviors to achieve different modes.

---
In reply to: [386531155](https://github.com/VowpalWabbit/reinforcement_learning/pull/129#discussion_r386531155) [](ancestors = 386531155)

rlclientlib/serialization/fb_serializer.h
homezcx: No. It's not.

This is to be back-compatible with the config not setting learning mode. I will add a comment in the next push.

rajan-chari:
>     resp.clear();

1) I thought that discussion was about counter factual eval?
2) Return an error if CCB && imitation mode

---
In reply to: [593523356](https://github.com/VowpalWabbit/reinforcement_learning/pull/129#issuecomment-593523356) [](ancestors = 593523356,592561744)

---
Refers to: rlclientlib/live_model_impl.cc:94 in 36cc99e. [](commit_id = 36cc99e2c0277a987dacc38447911951b2798938, deletion_comment = False)

rlclientlib/live_model_impl.cc
homezcx: Fixed.

homezcx: Fixed.

---
In reply to: [385751901](https://github.com/VowpalWabbit/reinforcement_learning/pull/129#discussion_r385751901) [](ancestors = 385751901)

rlclientlib/live_model_impl.cc
homezcx: Changed to switch-case statement.

rlclientlib/live_model_impl.cc
homezcx: Now it has changed to in-memory replacement. Please see the change in `container_iterator.h`

rlclientlib/schema/DecisionMode.fbs
homezcx: Removed.

rlclientlib/schema/RankingEvent.fbs
homezcx: Tyler suggested me adding the original array of actionId and probabilities in the ImitationMode structure. So We probably won't be able to use an enum.
 
From what I understand, the table is optional and struct is mandatory. We should be compatible.

I am working on this change.

rlclientlib/schema/DecisionMode.fbs
homezcx: Please check comment in RankingEvent.fbs

homezcx:
>     resp.clear();

1) I think you are right. I might have misinterpreted the email
2) I was following some suggestion from Alexey that using "InteractionMode: "IMITATION_MODE"" in config.json mean CB only. For CCB, we should implement "DecisionMode" in config.json. What do you think of this approach?


---
In reply to: [593692574](https://github.com/VowpalWabbit/reinforcement_learning/pull/129#issuecomment-593692574) [](ancestors = 593692574,593523356,592561744)

---
Refers to: rlclientlib/live_model_impl.cc:94 in 36cc99e. [](commit_id = 36cc99e2c0277a987dacc38447911951b2798938, deletion_comment = False)

homezcx:
>     resp.clear();

We synced up internally @rajan.chari@yahoo.com and we prefer your approach - Return an error if CCB & imitation mode.

---
In reply to: [593712378](https://github.com/VowpalWabbit/reinforcement_learning/pull/129#issuecomment-593712378) [](ancestors = 593712378,593692574,593523356,592561744)

---
Refers to: rlclientlib/live_model_impl.cc:94 in 36cc99e. [](commit_id = 36cc99e2c0277a987dacc38447911951b2798938, deletion_comment = False)

rlclientlib/schema/RankingEvent.fbs
homezcx: We had the discussion yesterday and decided to go with what you suggested here.

---
In reply to: [386537263](https://github.com/VowpalWabbit/reinforcement_learning/pull/129#discussion_r386537263) [](ancestors = 386537263)

rlclientlib/live_model_impl.cc
homezcx: Fixed.

---
In reply to: [386552274](https://github.com/VowpalWabbit/reinforcement_learning/pull/129#discussion_r386552274) [](ancestors = 386552274)

include/constants.h
```diff
@@ -11,6 +11,9 @@ namespace reinforcement_learning {  namespace name {
       const char *const  VW_CMDLINE              = "vw.commandline";
       const char *const  VW_POOL_INIT_SIZE       = "vw.pool.init.size";
       const char *const  INITIAL_EPSILON         = "initial_exploration.epsilon";
+      const char *const  LEARNING_MODE = "rank.learning.mode";
+      const char *const  LEARNING_MODE_ONLINE = "rank.learning.mode.online";
+      const char *const  LEARNING_MODE_IMITATION = "rank.learning.mode.imitation";
```
tyclintw: [nit] spacing

homezcx: yes. I just fixed the constants in the tests as well. It was missed from pushing.

---
In reply to: [387881109](https://github.com/VowpalWabbit/reinforcement_learning/pull/129#discussion_r387881109) [](ancestors = 387881109)

rlclientlib/live_model_impl.cc
```diff
@@ -417,4 +423,35 @@ namespace reinforcement_learning {
...
+      default:
+      {
+        // This is to be back-compatible with the config not setting learning mode.
+        break;
+      }
+    }
+    return error_code::success;
+  }
+
+  learning_mode to_learning_mode(const char* learning_mode) {
```
tyclintw: >to_learning_mode [](start = 16, length = 16)

move this to learning_mode.h/.cc

rlclientlib/ranking_response.cc
```diff
@@ -64,6 +64,11 @@ namespace reinforcement_learning {
     _ranking.clear();
   }
 
+  void ranking_response::clear_ranking() {
+    _ranking.clear();
+    _chosen_action_id = 0;
```
homezcx: note to myself: this probably need to reset to -1

include/container_iterator.h
```diff
@@ -47,10 +57,20 @@ namespace reinforcement_learning {
...
+    container_iterator operator-(const int idx) const {
+      return { _coll, _idx - idx };
+    }
     //! Increment the index
-    container_iterator operator+(const uint32_t idx) const {
+    container_iterator operator+(const int idx) const {
       return { _coll, _idx + idx };
     }
+    //! Assign the values for the iterator
+    container_iterator operator=(const container_iterator& other) {
```
ataymano: container_iterator& as return type

include/container_iterator.h
```diff
@@ -47,10 +57,20 @@ namespace reinforcement_learning {
     int64_t operator-(const container_iterator& rhs) const {
       return _idx - rhs._idx;
     }
+    //! Allow distance measurement
+    container_iterator operator-(const int idx) const {
```
ataymano: Do we need this?
Also, it is outside of scope of this PR, but I believe there is a size_t overflowing issue on lines 57-59 which we need to fix.

homezcx: Yes. There are required to sort the ranking array (action, probability). Check live_modeL_impl.cc post_process_rank.

Here I only added the minimum number of operator I need

ataymano: oh I see.

rlclientlib/ranking_event.cc
```diff
@@ -33,17 +33,19 @@ namespace reinforcement_learning {
...
     copy(context_str.begin(), context_str.end(), std::back_inserter(_context));
+    _learning_mode = decision_mode;
   }
 
   const std::vector<unsigned char>& ranking_event::get_context() const { return _context; }
   const std::vector<uint64_t>& ranking_event::get_action_ids() const { return _action_ids_vector; }
   const std::vector<float>& ranking_event::get_probabilities() const { return _probilities_vector; }
   const std::string& ranking_event::get_model_id() const { return _model_id; }
   bool ranking_event::get_defered_action() const { return _deferred_action; }
+  const learning_mode ranking_event::get_learning_mode() const { return _learning_mode; }
```
ataymano: const is redundant here

rlclientlib/ranking_response.cc
```diff
@@ -64,6 +64,11 @@ namespace reinforcement_learning {
     _ranking.clear();
   }
 
+  void ranking_response::clear_ranking() {
```
ataymano: are we still using it?

rlclientlib/ranking_event.cc
ataymano: I mean the one associated with returned value

unit_test/live_model_test.cc
```diff
@@ -131,6 +131,56 @@ BOOST_AUTO_TEST_CASE(live_model_ranking_request) {
...
+  //create the ds live_model, and initialize it with the config
+  r::live_model ds = create_mock_live_model(config);
+  BOOST_CHECK_EQUAL(ds.init(&status), err::success);
+
+  const auto event_id = "event_id";
+  r::ranking_response response;
+  ds.choose_rank(event_id, JSON_CONTEXT, response, &status);
+  BOOST_CHECK_EQUAL(status.get_error_code(), 0);
+  BOOST_CHECK_EQUAL(status.get_error_msg(), "");
+}
```
tyclintw: [nit] newline

unit_test/live_model_test.cc
```diff
@@ -131,6 +131,56 @@ BOOST_AUTO_TEST_CASE(live_model_ranking_request) {
...
+
+  //create the ds live_model, and initialize it with the config
+  r::live_model ds = create_mock_live_model(config);
+  BOOST_CHECK_EQUAL(ds.init(&status), err::success);
+
+  const auto event_id = "event_id";
+
+  r::ranking_response response;
+
+  ds.choose_rank(event_id, JSON_CONTEXT, response, &status);
```
tyclintw: >JSON_CONTEXT [](start = 27, length = 12)

JSON_CONTEXT has no actions, so are we testing anything here?

rlclientlib/ranking_event.cc
```diff
@@ -33,17 +33,19 @@ namespace reinforcement_learning {
     }
     string context_str(context);
     copy(context_str.begin(), context_str.end(), std::back_inserter(_context));
+    _learning_mode = decision_mode;
```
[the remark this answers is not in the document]
homezcx: I will move it to the initializer list. Making it const will require all constructor changes. I prefer not changing this part, as we don't have any const members in this class (except vectors)

include/container_iterator.h
homezcx: Fixed.

include/container_iterator.h
homezcx: Fixed line 57-59 to use size_t

rlclientlib/ranking_event.cc
homezcx: Fixed.

rlclientlib/ranking_response.cc
homezcx: Removed.

unit_test/live_model_test.cc
homezcx: Fixed.

unit_test/live_model_test.cc
homezcx: Added a new JSON_CONTEXT_LEARNING and I verified online mode and imitation mode could give different suggestions.

Due to exploration, the selected action in online mode is not asserted.

rlclientlib/live_model_impl.cc
homezcx: Fixed.

include/learning_mode.h
```diff
@@ -0,0 +1,20 @@
...
+* @file decision_modes.h
+* @author Chenxi Zhao
+* @date 2020-02-13
+*/
+#pragma once
+
+namespace reinforcement_learning {
+
+  enum learning_mode {
+    ONLINE_MODE = 0,
```
dwaijam: Change to ONLINE and IMITATION

homezcx: Please check constants.h - config values constants are defined there.

Here I would prefer using ONLINE_MODE as enum reference so the code would be easier to read.

E.g live_model_impl.cc line 93

`    if (_learning_mode == IMITATION_MODE) {
      // Imitaion mode is not supported here at this moment
      return error_code::not_supported;
    }
`

include/container_iterator.h
ataymano: counterexample:
_idx = 1
rhs._idx = 2

_idx - rhs._idx will be first evaluated as size_t (to max(size_t) - 1) and then converted to return type.
So return type insignature should be left as int64_t, but return expression changed to:
static_cast<int64_t>(_idx) - static_cast<int64_t>(rhs._idx)

homezcx: Fixed.

Issue #45, opened 2019-02-15
Support for converting rlclient lib error codes to string descriptions


rajan-chari:
Currently all API methods take an apistatus object if error details are needed as strings.  It would be good to have a utility method to convert the error codes into strings.

lokitoth:
One way to do something like this is to set up a single "ground truth" source of constants (config keys/values, error numbers, error strings) to be shared between the different bindings and RLClientLib. Then rather than opening up an API, we simply generate the language-appropriate constants code.

jackgerrits:
We actually currently have this and it is how I generated the relevant exceptions in Python and error constants in C++.

See this file: https://github.com/VowpalWabbit/reinforcement_learning/blob/master/include/errors_data.h

PR #131, opened 2020-02-28
Support for converting rlclient lib error codes to string descriptions(reinforcement_learning#45)

mehulrastogi:
Added a utility function in rlclientlib/utility/config_utility.cc

Issue#45 mentioned here : https://github.com/VowpalWabbit/reinforcement_learning/issues/45

Also has minor fixes:

 - err_data.h had duplicate error codes


rajan-chari:
It would be good if we don't have to add error information in ERROR_CODE_DEFINITION (errors_data.h) and in error_code_string() function

prateekiiest:
Hi is this PR currently worked upon?
I guess from @kumpera 's comments thing that we can improve is to have error code numbers being matched under the switch case rather than error code literals. 

Also as I understand @rajan-chari that we dont want to have error information (2nd args) in the error code definition just the error code number and the error description right?

jackgerrits:
Closing this as the PR is stale and has not been looked at in a long time. The preferred method to handle returning the error string would be to utilize the error definition macro. If you wanted to do a different PR using that please feel free

PR #135, opened 2020-03-20
Accumulated update on Imitation Mode

homezcx:
1. LearningMode config should not be case sensitive
2. Original ranking Result should be sent to Eventhub in Imitation Mode

### DIFF
--- a/unit_test/live_model_test.cc
+++ b/unit_test/live_model_test.cc
@@ -152,26 +152,6 @@ BOOST_AUTO_TEST_CASE(live_model_ranking_request_online_mode) {
   BOOST_CHECK_EQUAL(status.get_error_msg(), "");
 }
 
-BOOST_AUTO_TEST_CASE(live_model_ranking_request_online_mode_lower_case) {
-  //create a simple ds configuration
-  u::configuration config;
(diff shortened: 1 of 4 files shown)


unit_test/live_model_test.cc
```diff
@@ -152,6 +152,26 @@ BOOST_AUTO_TEST_CASE(live_model_ranking_request_online_mode) {
...
+  //create the ds live_model, and initialize it with the config
+  r::live_model ds = create_mock_live_model(config);
+  BOOST_CHECK_EQUAL(ds.init(&status), err::success);
+
+  const auto event_id = "event_id";
+  r::ranking_response response;
+  ds.choose_rank(event_id, JSON_CONTEXT_LEARNING, response, &status);
+  BOOST_CHECK_EQUAL(status.get_error_code(), 0);
+  BOOST_CHECK_EQUAL(status.get_error_msg(), "");
+}
```
tyclintw: There is no reason to test an entire rank request when all you want to test is case sensitivity. Please isolate the functionality to a unit test, i.e. testing the function to_learning_mode explicitly.

homezcx: Fixed.

PR #160, opened 2020-07-07
Add LoggingOnly learning mode to rlclientlib

zli12:
Add support for LoggingOnly mode:
1. Add LoggingOnly to LearningMode enum
2. Return response with action order as is from input
3. Use input action order and add LoggingOnly tag to events sent to eventhub. 
4. Rename Imitation to Apprentice
5. Update existing and add new unit tests

### DIFF
--- a/rlclientlib/live_model_impl.cc
+++ b/rlclientlib/live_model_impl.cc
@@ -90,8 +90,8 @@ namespace reinforcement_learning {
 
   int live_model_impl::request_decision(const char* context_json, unsigned int flags, decision_response& resp, api_status* status)
   {
-    if (_learning_mode == IMITATION) {
-      // Imitaion mode is not supported here at this moment
+    if (_learning_mode == APPRENTICE) {
+      // Apprentice mode is not supported here at this moment
       return error_code::not_supported;
     }
 
@@ -429,8 +429,10 @@ namespace reinforcement_learning {
 
   int post_process_rank(ranking_response& response, learning_mode learning_mode) {
     switch (learning_mode) {
-    case IMITATION:
+    case APPRENTICE:
+    case LOGGINGONLY:
       {
+        // In Apprentice or LoggingOnly mode, we keep the order of actions from input
 #ifdef __clang__
         std::vector<action_prob> tmp;
         std::copy(response.begin(), response.end(), std::back_inserter(tmp));
--- a/rlclientlib/live_model_impl.cc
+++ b/rlclientlib/live_model_impl.cc
@@ -36,7 +36,7 @@ namespace reinforcement_learning {
 
   int check_null_or_empty(const char* arg1, const char* arg2, api_status* status);
   int check_null_or_empty(const char* arg1, api_status* status);
-  int post_process_rank(ranking_response& response, learning_mode learning_mode);
+  int reset_action_order(ranking_response& response);
 
   void default_error_callback(const api_status& status, void* watchdog_context) {
     auto watchdog = static_cast<utility::watchdog*>(watchdog_context);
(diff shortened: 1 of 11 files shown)


unit_test/fb_serializer_test.cc
```diff
@@ -71,7 +71,7 @@ BOOST_AUTO_TEST_CASE(fb_serializer_ranking_event) {
   const timestamp ts;
   const size_t events_count = 10;
   for (size_t i = 0; i < events_count; ++i) {
-    mode = static_cast<learning_mode>(i % 2);
+    mode = static_cast<learning_mode>(i % 3);
```
lalo, line 74: might be good code hygiene to have this 3 as it's own const since it's being referenced multiple times

unit_test/fb_serializer_test.cc
```diff
@@ -94,11 +94,14 @@ BOOST_AUTO_TEST_CASE(fb_serializer_ranking_event) {
     BOOST_CHECK_EQUAL_COLLECTIONS(event.probabilities()->begin(), event.probabilities()->end(), expected_prob.begin(), expected_prob.end());
     BOOST_CHECK_EQUAL(event.deferred_action(), false);
     BOOST_CHECK_EQUAL(event.pass_probability(), 0.33f);
-    if (i % 2 == 0) {
```
betamaxed, line 97: These module operators are no longer appropriate. A switch may be better suited. Also, think about moving away from this hardcoded 3 and referencing the size of the enum.

rlclientlib/live_model_impl.cc
```diff
@@ -91,8 +91,8 @@ namespace reinforcement_learning {
 
   int live_model_impl::request_decision(const char* context_json, unsigned int flags, decision_response& resp, api_status* status)
   {
-    if (_learning_mode == IMITATION) {
-      // Imitaion mode is not supported here at this moment
+    if (_learning_mode == APPRENTICE) {
```
jackgerrits, line 94: Is logging only supported here?

unit_test/fb_serializer_test.cc
zli12: sounds good. fixed.

rlclientlib/live_model_impl.cc
zli12: good catch. fixed as loggingonly is not supported either.

zli12: the idea for loggingonly mode is that we basically ignore the result from choose_rank like what we did for apprentice. So we return a response as actions in the same order as the input. Also adding a flag for other stuffs to be handled in joiner/trainer. Please let me know if you have any concerns and/or other questions.

zli12:
> I'm not quite sure I understand how LoggingOnly works. If I create a live model in LoggingOnly mode and call `choose_rank` what happens?

the idea for loggingonly mode is that we basically ignore the result from choose_rank like what we did for apprentice. So we return a response as actions in the same order as the input. Also adding a flag for other stuffs to be handled in joiner/trainer. Please let me know if you have any concerns and/or other questions.

unit_test/learning_mode_test.cc
```diff
@@ -26,7 +38,7 @@ BOOST_AUTO_TEST_CASE(learning_mode_online) {
 }
 
 BOOST_AUTO_TEST_CASE(learning_mode_other) {
-  const char* imitation = "other";
-  auto mode = reinforcement_learning::learning::to_learning_mode(imitation);
+  const char* apprentice = "other";
```
homezcx, line 41: I cannot remember why I put `imitation` before but I think this should be `other`

dwaijam:
LoggingOnly mode should write baseline order to EventHub.

rlclientlib/live_model_impl.cc
```diff
@@ -71,8 +71,19 @@ namespace reinforcement_learning {
     }
     response.set_event_id(event_id);
 
+    if (_learning_mode == LOGGINGONLY)
+    {
+      // Reset the ranked action order before logging
+      RETURN_IF_FAIL(reset_action_order(response));
```
dwaijam, line 77: Can you add a test around this logic?

rlclientlib/live_model_impl.cc
```diff
@@ -491,33 +502,23 @@ namespace reinforcement_learning {
     return error_code::success;
   }
 
-  int post_process_rank(ranking_response& response, learning_mode learning_mode) {
-    switch (learning_mode) {
-    case IMITATION:
-      {
+  int reset_action_order(ranking_response& response) {
```
ataymano, line 505: do we need to do it as std::sort? isn't it always going to be [0, 1, 2, 3, ..] which can be generated with linear complexity given only number of actions?

rlclientlib/live_model_impl.cc
zli12: yes we have tests in live_model_test.cc

unit_test/learning_mode_test.cc
zli12: renamed.

PR #168, opened 2020-08-18
Create ISender extension point in RL.Net

lokitoth:
RLClientLib is configured to use Azure EventHub or File-based logging out of the box. It is possible to extend this using the i_sender and factory APIs. This change projects the set of APIs needed to enable this extensibility in the .NET bindings.

* Implements factory_context as a mechanism to interact with RLClientLib factories in a principled manner and provides a pattern to register factories on the managed side.
* Implements "BINDING_SENDER"-typed ISender extensibility. Currently it is not possible to configure multiple types of ISender implementations, for simplicity.
* Implements background error propagation from Managed to Native (and back), as well as ApiStatusBuild API (similar-to native APIs, but only brings the string across the boundary once), and automatic conversion of Exceptions/RLExceptions => ApiStatus updates.
* Implements error code to error string lookup from managed side.

* Implements AsyncSender to simplify async/await-based implementations


bindings/cs/rl.net.native/rl.net.factory_context.cc
```diff
@@ -0,0 +1,74 @@
...
+    delete sender_factory;
+  }
+}
+
+API void DeleteFactoryContext(factory_context_t* context)
+{
+  cleanup_trace_logger_factory(context->trace_logger_factory);
+  cleanup_sender_factory(context->sender_factory);
+
+  // TODO: Once we project the others, we will need to add them to cleanup.
```
olgavrou, line 44: should this TODO be documented as a issue/ticket?

bindings/cs/rl.net.native/rl.net.buffer.h
```diff
@@ -0,0 +1,22 @@
...
+
+namespace rl_net_native
+{
+  using buffer = std::shared_ptr<reinforcement_learning::utility::data_buffer>;
+  using error_context = reinforcement_learning::error_callback_fn;
+}
+
+extern "C" 
+{
+  // API const buffer* CreateBufferWithSharedPointer();
```
olgavrou, line 15: remove commented out code

bindings/cs/rl.net.native/binding_sender.cc
```diff
@@ -0,0 +1,27 @@
...
+
+namespace rl_net_native {
+  int binding_sender::init(api_status* status)
+  {
+    return this->vtable.init(managed_handle, status);
+  }
+
+  int binding_sender::v_send(const buffer& data, api_status* status)
+  {
+    size_t length = data->buffer_filled_size();
```
olgavrou, line 14: `auto length`

lokitoth: I prefer to specify the types explicitly - it makes it much easier to read for people who are not using tools beyond notepad or who have not seen the code recently, and do not know the actual types. Using auto also tends to lead people to ignore auto-coersion between type-sizes, which we keep encountering in the form of precision-loss warnings.

While I will not comment about others' use of it, I view `auto` (and `var` in C#) as harmful in the majority of cases, particularly when the type name is short.

bindings/cs/rl.net/RLException.cs
```diff
@@ -40,6 +46,14 @@ public override void GetObjectData(SerializationInfo info, StreamingContext cont
...
             info.AddValue(ErrorCodeSerializationKey, this.ErrorCode);
             base.GetObjectData(info, context);
+        } 
+
+        internal int UpdateApiStatus(ApiStatus status)
+        {
+            ApiStatus.Update(status, this.ErrorCode, this.Message);
+            return this.ErrorCode;
+
+            // TODO: Project out InnerException?
```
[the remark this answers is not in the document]
lokitoth: I left some of these as discussion points for the review, but yes, I intend to remove them / turn them into issues.

bindings/cs/rl.net.native/rl.net.factory_context.cc
lokitoth: This is there to ensure that anyone working on projecting out the rest of the extensibility points will make sure they add them to the cleanup function, since there is no mechanism to actually change them away from the global factories right now.

The alternative is to add the code right now, so it does not lead to an issue in the future, but at the cost of superfluous code (in practice).

.scripts/restore.cmd
```diff
@@ -34,6 +34,9 @@ ECHO Restoring SDK-Style projects in "%rlRoot%\rl.sln"
 %dotnetPath% restore "%rlRoot%\rl.sln"
 ECHO.
 
+ECHO Restoring DotNet Tools (made simpler in .NET Core 3.0)
```
rajan-chari, line 37: I am curious when does this file run and what does the following statement do?

lokitoth: It restores the dotnet-t4 tool, which is a dotnet core (and thus, cross platform) version of the old TextTemplating tool. What this does it enable .NET Core SDK < 3.0 to restore the tool (it has to install it globally, the .json file trigger restoration for local).

This is then used in the new TextTemplating targets file to actually drive the template generation.

bindings/cs/rl.net/RLException.cs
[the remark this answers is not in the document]
lokitoth: I switched this to a debug-only expansion of the inner-exception. If this turns out to be useful in Release, we can remove the [Conditional] attribute.

lokitoth:
I tried using .lgtm.yml to get JavaScript to not false-positive here, but that seems like a project-level setting, rather than config file. We can ignore this check, because it is failing in the same place on Master.

bindings/cs/rl.net/FactoryContext.cs
```diff
@@ -0,0 +1,44 @@
...
+
+                    SenderAdapter adapter = new SenderAdapter(sender);
+
+                    // Unpinning this happens inside of SenderAdapter when Release is invoked
+                    GCHandle adapterHandle = GCHandle.Alloc(adapter);
+
+                    return GCHandle.ToIntPtr(adapterHandle);
+                };
+
+            SetFactoryContextBindingSenderFactory(this.DangerousGetHandle(), create_fn, SenderAdapter.VTable);
```
kumpera, line 41: create_fn becomes garbage after this call.
On the native side the callee stores the native funptr in `FactoryContext` native object.
We need to replicate that linkage on managed.

bindings/cs/rl.net/Native/SenderAdapter.cs
```diff
@@ -0,0 +1,145 @@
...
+        public static sender_vtable VTable
+        {
+            get;
+        } = CreateAdapterVTable();
+
+        private static sender_vtable CreateAdapterVTable()
+        {
+            return new sender_vtable
+            {
+                init = (managed_handle, status) => InvokeInit(managed_handle, status),
```
kumpera, line 47: We should not depend on the C# compiler caching those delegates in static variables and do it ourselves explicitly.

Those delegates are not explicitly tracked by managed code after a call uses `SenderAdapter.Vtable`.

bindings/cs/rl.net/Native/SenderAdapter.cs
```diff
@@ -0,0 +1,145 @@
...
+                    return apiStatus.ErrorCode;
+                },
+                status
+            );
+        }
+
+        private static int InvokeSend(IntPtr managed_handle, IntPtr buffer, IntPtr status)
+        {
+            return SenderAdapter.InvokeAndUnwrapExceptions(
+                (ApiStatus apiStatus) =>
```
kumpera, line 110: The `ApiStatus` argument is not really necessary since we're capturing `buffer` and `managed_handle` anyways

bindings/cs/rl.net.cli.test/CleanupContainer.cs
```diff
@@ -0,0 +1,52 @@
...
+
+namespace Rl.Net.Cli.Test
+{
+    public sealed class CleanupContainer : IDisposable
+    {
+        private Stack<IDisposable> cleanupStack = new Stack<IDisposable>();
+
+        public void Dispose()
+        {
+            Stack<IDisposable> localCleanupStack = Interlocked.Exchange(ref cleanupStack, new Stack<IDisposable>());
```
rajan-chari, line 13: Is the idea here that when Dispose is getting called, it's possible some other thread is adding to  cleanupStack?  Is that possible?  i.e. does calling Dispose() imply that there are no reachable instances?  If that is not the case, will the additional items get cleaned up properly since dispose has been already called?

bindings/cs/rl.net.cli.test/SenderExtensibilityTest.cs
```diff
@@ -0,0 +1,566 @@
...
+                    return new MockAsyncSender(callback)
+                    {
+                        Send = asyncSendFunc
+                    };
+                }
+            );
+        }
+
+        [TestMethod]
+        public void Test_CustomSender_FactoryCalled_WhenRequested()
```
rajan-chari, line 203: Great test names :)

bindings/cs/rl.net.native/CMakeLists.txt
```diff
@@ -15,3 +40,10 @@ endif()
 set_target_properties(rl.net.native PROPERTIES POSITION_INDEPENDENT_CODE ON)
 target_include_directories(rl.net.native PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
 target_link_libraries(rl.net.native PUBLIC rlclientlib)
+
+# Enable warnings as errors, crossplatform
+target_compile_options(rl.net.native PRIVATE
+     $<$<OR:$<CXX_COMPILER_ID:Clang>,$<CXX_COMPILER_ID:AppleClang>,$<CXX_COMPILER_ID:GNU>>:
+          -Werror>
```
jackgerrits, line 47: Do we want warnings as errors? While this seems like a good idea it can be a pain

bindings/cs/rl.net.native/binding_sender.h
```diff
@@ -0,0 +1,50 @@
...
+
+namespace rl_net_native {
+  namespace constants {
+    const char *const BINDING_SENDER = "BINDING_SENDER";
+  }
+
+  using buffer = std::shared_ptr<reinforcement_learning::utility::data_buffer>;
+  using error_context = reinforcement_learning::error_callback_fn;
+  
+  typedef void(*error_fn)(error_context* error_context, reinforcement_learning::api_status* api_status);
```
jackgerrits, line 17: use `using` instead of typedef

bindings/cs/rl.net.native/rl.net.native.cc
```diff
@@ -0,0 +1,23 @@
+#include "rl.net.native.h"
+
+using namespace reinforcement_learning::error_code;
+
+API const char* LookupMessageForErrorCode(int error_code)
```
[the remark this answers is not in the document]
lokitoth: Ah. Maybe it makes sense to uplevel this. I will address that separately. Good catch.

bindings/cs/rl.net.native/rl.net.api_status.cc
```diff
@@ -18,4 +18,14 @@ API const char* GetApiStatusErrorMessage(reinforcement_learning::api_status* sta
 API int GetApiStatusErrorCode(reinforcement_learning::api_status* status)
 {
     return status->get_error_code();
-}
\ No newline at end of file
+}
+
+API void UpdateApiStatusSafe(reinforcement_learning::api_status* status, int error_code, const char* message)
+{
+    reinforcement_learning::api_status::try_update(status, error_code, message);
```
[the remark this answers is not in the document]
lokitoth: I do not believe so: `api_status` takes a copy of `message` and holds on to it, as it is assumed that the caller could free the string coming into `try_update` whenever:

    static void try_update(api_status* status, int new_code, const char* new_msg);

The `api_status` object then owns freeing that copy - which it will do when it gets deleted. The incoming copy is the responsibility of the caller of `UpdateApiStatusSafe`.

bindings/cs/rl.net/ApiStatus.cs
```diff
@@ -52,5 +59,58 @@ public string ErrorMessage
...
         }
+
+        internal static void Update(ApiStatus status, int errorCode, string message)
+        {
+            unsafe 
+            { 
+                fixed (byte* messageBytes = NativeMethods.StringEncoding.GetBytes(message))
+                {
+                    IntPtr messagePtr = new IntPtr(messageBytes);
+                    UpdateApiStatusSafe(status.ToNativeHandleOrNullptrDangerous(), errorCode, messagePtr);
```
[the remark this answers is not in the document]
lokitoth: This should not be a leak because `messagePtr` is only valid for the lifetime of the `fixed` block. This code relies on the understanding that the `api_status` object copies the string on input. Broadly speaking, the RLClientLib bindings assume that the native-side will introduce a copy if necessary, but strings coming in only need to live until the NativeMethod returns.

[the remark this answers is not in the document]
kumpera: You're right, GetBytes allocates a `byte[]` and not malloc something.

bindings/cs/rl.net/AsyncSender.cs
```diff
@@ -0,0 +1,64 @@
...
+        {
+        }
+
+        public void Send(SharedBuffer buffer, ApiStatus status)
+        {
+            // Create a cloned handle so that we have an our own copy of the shared pointer to the buffer 
+            SharedBuffer ownedHandle = new SharedBuffer(buffer);
+            GC.KeepAlive(buffer);
+
+            Task backgroundTask = SendAsyncAndUnwrapExceptions(ownedHandle);
```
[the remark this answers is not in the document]
lokitoth: I tried here to capture the same semantics as in [eventhub_client.cc](https://github.com/VowpalWabbit/reinforcement_learning/blob/master/rlclientlib/logger/eventhub_client.cc#L126): It creates a task (though it manually manages the max number of tasks, rather than relying on the async/await system to manage this) and fire-and-forgets it.

The idea would be that it is possible to implement a sync sender, as well as a FNF one like the native EventHub one.

bindings/cs/rl.net/AsyncSender.cs
```diff
@@ -0,0 +1,64 @@
...
+        public AsyncSender(ErrorCallback callback)
+        {
+            this.errorCallback = callback;
+        }
+
+        public virtual void Init(ApiStatus status)
+        {
+        }
+
+        public void Send(SharedBuffer buffer, ApiStatus status)
```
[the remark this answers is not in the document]
lokitoth: AsyncSender is intended to be a FNF adapter over the synchronous ISender/i_sender API.

bindings/cs/rl.net/AsyncSender.cs
```diff
@@ -0,0 +1,64 @@
...
+using System.Threading.Tasks;
+using Rl.Net.Native;
+
+namespace Rl.Net
+{
+    public abstract class AsyncSender : ISender
+    {
+        private ErrorCallback errorCallback;
+
+        public AsyncSender(ErrorCallback callback)
```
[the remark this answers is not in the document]
lokitoth: I will add the event as an indirection mechanism, though it is odd to expose an event on a class, then only expect it to be used to raise events, rather than subscribe to them.

bindings/cs/rl.net/FactoryContext.cs
lokitoth: Nice catch!

bindings/cs/rl.net.cli.test/CleanupContainer.cs
lokitoth: This is primarily intended as a best-efforts cleanup mechanism for tests. The exchange is there to ensure that if tests are running in an unexpected multithread situations, we do not start throwing out of cleanup with iteration at the same time as collection changed.

Mostly this is a habit developed for dealing with IDisposables in multi-threaded situations: When it doubt, use Interlocked to null out / reset the shared object, and then clean up locally. The "right" solution to this is to have a separate background queue for calling .Dispose(), but it is overkill for unit tests.

bindings/cs/rl.net.native/CMakeLists.txt
lokitoth: Yes. I want to slowly bring this out through the project - after fixing the appropriate warnings on our shipping platforms. Right now it is contained to rl.net.native where I *really* do not want warnings.

bindings/cs/rl.net/Native/SenderAdapter.cs
lokitoth: Nice catch!

bindings/cs/rl.net.cli.test/CleanupContainer.cs
```diff
@@ -0,0 +1,52 @@
...
+            while (localCleanupStack.TryPop(out IDisposable cleanupDisposable))
+            {
+                if (cleanupDisposable != null)
+                {
+                    cleanupDisposable.Dispose();
```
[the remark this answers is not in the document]
lokitoth: Generally .Dispose() should avoid throwing unless there is a serious corruption - do we want to discover this during tests?

    AVOID throwing an exception from within Dispose(bool) except under critical situations where the containing process has been corrupted (leaks, inconsistent shared state, etc.).

(From the Framework Design Guidelines)

bindings/cs/rl.net.cli.test/TempFileDisposable.cs
```diff
@@ -0,0 +1,39 @@
...
+        public void Dispose()
+        {
+            try
+            {
+                if (File.Exists(this.Path))
+                {
+                    File.Delete(this.Path);
+                }
+
+                if (Directory.Exists(this.Path))
```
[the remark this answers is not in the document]
lokitoth: It should not be able to, since `.Path` is the result of Path.GetTempFileName(). The idea is to ensure it deletes directories created in this manner, since `File.Exists(path)` returns false if `path` points to a directory.

bindings/cs/rl.net.native/rl.net.live_model.cc
```diff
@@ -15,7 +16,7 @@ static void pipe_background_error_callback(const reinforcement_learning::api_sta
     }
 }
 
-API livemodel_context_t* CreateLiveModel(reinforcement_learning::utility::configuration* config)
+API livemodel_context_t* CreateLiveModel(reinforcement_learning::utility::configuration* config, factory_context_t* factory_context)
```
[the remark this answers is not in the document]
lokitoth: Yes, this is correct, but marking it `const` would not really do anything here, since this is an internal API intended to be used by bindings. Still, a fair point.

bindings/cs/rl.net/Native/SenderAdapter.cs
lokitoth: It seems a bit cleaner to have `InvokeAndUnwrapExceptions` be responsible for creating the `ApiStatus` managed wrapper and managing its lifetime - otherwise it would just be duplicated code across `InvokeInit` and `InvokeSend`

bindings/cs/rl.net.cli.test/CleanupContainer.cs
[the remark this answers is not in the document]
lokitoth: I ended up adding the try/catch. If we want to be testing for goodness during cleanup it should be explicit, rather than implicit in test cleanup.

bindings/cs/rl.net.native/binding_sender.h
lokitoth: Done. I keep forgetting that exists.

bindings/cs/rl.net.native/CMakeLists.txt
lokitoth: Funny bit is that I forgot to include warnings-as-errors in Windows. Fixed.

bindings/cs/rl.net.cli.test/CleanupContainer.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        public void Add(IDisposable disposable)
+        {
+            this.cleanupStack.Push(disposable);
+        }
+
+        public void Add(Action action)
+        {
+            this.cleanupStack.Push(new ActionDisposable(action));
```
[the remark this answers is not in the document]
lokitoth: Nice!

PR #169, opened 2020-08-19
Ataymano/protocol v2

ataymano:
Flaltbuffers schema update.

Important:
We have discussed changing CCB api from unique slot id (which is hard to support with new schema) to event id per slot index, and it is reflected here.
 ccb_logger_facade has 2 subsets of methods:
- with slot_id - supports only v1 protocol
- with event_id + slot index - supports only v2 protocol.


rlclientlib/logger/event_logger.h
```diff
@@ -169,4 +169,28 @@ class slates_logger : public event_logger<slates_decision_event> {
 
     int report_action_taken(const char* event_id, api_status* status);
   };
+
+  class generic_event_logger : public event_logger<generic_event> {
+  public:
+    generic_event_logger(i_message_sender* sender,
+      int send_high_watermark,
```
kumpera, line 176: Why the generic logged needs send_high_watermark,send_batch_interval_ms,etc while the old ones don't?

kumpera: My issues is that we're repeating initialization all over the place and this is error prone.

rlclientlib/logger/logger_facade.cc
```diff
@@ -48,14 +69,32 @@ namespace reinforcement_learning {
       }
     }
 
+    int ccb_logger_facade::log_decisions(const char* event_id, const char* context, unsigned int flags, const std::vector<std::vector<uint32_t>>& action_ids,
+      const std::vector<std::vector<float>>& pdfs, const std::string& model_version, api_status* status) {
+      switch (version) {
+        case 2: return v2->log(event_id, serializer.event(context, flags, action_ids, pdfs, model_version), serializer.type, status);
```
kumpera, line 75: Why no v1 here? Shouldn't we fail differently in that case?

rlclientlib/logger/logger_facade.cc
```diff
@@ -64,39 +103,67 @@ namespace reinforcement_learning {
...
 
     int observation_logger_facade::log(const char* event_id, const char* outcome, api_status* status) {
       switch (version) {
         case 1: return v1->log(event_id, outcome, status);
+        case 2: return v2->log(event_id, serializer.event(outcome), serializer.type, status);
+        default: return protocol_not_supported(status);
+      }
+    }
+
+    int observation_logger_facade::log(const char* event_id, int index, float outcome, api_status* status) {
```
kumpera, line 149: Same question, only v2 here, why? Where is this exposed?

rlclientlib/ranking_event.cc
```diff
@@ -125,4 +125,37 @@ namespace reinforcement_learning {
   const std::string& outcome_event::get_outcome() const { return _outcome; }
   float outcome_event::get_numeric_outcome() const { return _float_outcome; }
   bool outcome_event::get_action_taken() const { return _action_taken; }
+
+  generic_event::generic_event(const char* _id, const timestamp& ts, payload_type type, flatbuffers::DetachedBuffer&& payload, float pass_prob)
```
kumpera, line 129: nip: should this go to a separate file?

rlclientlib/ranking_event.cc
```diff
@@ -125,4 +125,37 @@ namespace reinforcement_learning {
...
+    : _id(_id)
+    , _client_time_gmt(ts)
+    , _payload_type(type)
+    , _payload(std::move(payload))
+    , _pass_prob(pass_prob) {}
+
+  bool generic_event::try_drop(float pass_prob, int drop_pass) {
+    _pass_prob *= pass_prob;
```
kumpera, line 137: This code is repeated all over the place, can we unify it?

rlclientlib/ranking_event.cc
```diff
@@ -125,4 +125,37 @@ namespace reinforcement_learning {
...
+  }
+
+  const char* generic_event::get_id() const { return _id.c_str(); }
+
+  float generic_event::get_pass_prob() const { return _pass_prob; }
+
+  timestamp generic_event::get_client_time_gmt() const { return _client_time_gmt; }
+
+  float generic_event::prg(int drop_pass) const {
+    const auto seed_str = _id + std::to_string(drop_pass);
```
kumpera, line 148: nip: code duplication.

rlclientlib/ranking_event.h
```diff
@@ -170,4 +172,47 @@ namespace reinforcement_learning {
     bool _action_taken = false;
     unsigned int _outcome_type = 0;
   };
+
+  enum payload_type {
+    CB = 1,
+    CCB = 2,
+    SLATES = 3,
+    OUTCOME_SINGLE = 4
```
kumpera, line 180: how is an activation event encoded?
will this handle the logging interface you added?

rlclientlib/ranking_event.h
```diff
@@ -170,4 +172,47 @@ namespace reinforcement_learning {
...
+  };
+
+  class generic_event {
```
kumpera, line 183: shouldn't we move this to a separate header since it's part of the v2 protocol?

rlclientlib/serialization/fb_serializer.h
```diff
@@ -242,6 +247,56 @@ namespace reinforcement_learning { namespace logger {
...
+  struct fb_event_serializer<generic_event> {
+    using fb_event_t = v2::Event;
+    using offset_vector_t = typename std::vector<flatbuffers::Offset<fb_event_t>>;
+    using batch_builder_t = v2::EventBatchBuilder;
+
+    static size_t size_estimate(const generic_event& evt) {
+      return evt.get_payload().size();
+    }
+
+    static int to_fb_payload_type(payload_type type, v2::PayloadType& result, api_status* status) {
```
kumpera, line 260: How is this not a numeric cast?

rlclientlib/serialization/payload_serializer.cc
```diff
@@ -0,0 +1,16 @@
...
+namespace reinforcement_learning {
+  using namespace messages::flatbuff;
+  namespace logger {
+    v2::LearningModeType GetLearningMode(learning_mode mode) {
+      v2::LearningModeType result;
+      switch (mode) {
+      case APPRENTICE: return v2::LearningModeType_Apprentice;
+      case ONLINE: return v2::LearningModeType_Online;
+      case LOGGINGONLY: return v2::LearningModeType_LoggingOnly;
+      default: return v2::LearningModeType_Online;
```
kumpera, line 12: This function smells bad. Why is it not a cast and why the default case not an error?

rlclientlib/serialization/payload_serializer.h
```diff
@@ -0,0 +1,134 @@
...
+        std::vector<float> probabilities;
+        for (auto const& r : response) {
+          action_ids.push_back(r.action_id + 1);
+          probabilities.push_back(r.probability);
+        }
+        std::vector<unsigned char> _context;
+        std::string context_str(context);
+        copy(context_str.begin(), context_str.end(), std::back_inserter(_context));
+
+        auto fb = v2::CreateCbEventDirect(fbb, flags | action_flags::DEFERRED, &action_ids, &_context, &probabilities, response.get_model_id(), GetLearningMode(learning_mode));
```
kumpera, line 46: setting action_flags::DEFERRED here looks wrong.

rlclientlib/serialization/payload_serializer.h
```diff
@@ -0,0 +1,134 @@
...
+
+      static generic_event::payload_buffer_t event(int index, float outcome) {
+        flatbuffers::FlatBufferBuilder fbb;
+        const auto evt = v2::CreateNumericEventIndexed(fbb, outcome, index).Union();
+        auto fb = v2::CreateOutcomeSingleEvent(fbb, v2::OutcomeSingleEventBody_NumericEventIndexed, evt);
+        fbb.Finish(fb);
+        return fbb.Release();
+      }
+
+      static generic_event::payload_buffer_t event(int index, const char* outcome) {
```
kumpera, line 117: Does C++ special case 0 as convertible to NULL and thus maybe leading to some weird overloading resolution conflict here?

rlclientlib/logger/event_logger.h
ataymano: They all need - if you look inside cb_logger / ccb_logger /... implementation you will find similar pieces of code which is not visible here. I believe it should be fixed by having non plain config:
"interactions_config":
 { 
    "send_batch_interval": 10, ... 
},
"observations_config":
{
    "send_batch_interval": ...
}
instead of 
{
  "interactions_send_batch_interval": 10, 
  ...
   "observations_send_batch_interval": ...
 ...
}
that we have now.
But it seems independent from what we try to acheve in this PR.

rlclientlib/logger/logger_facade.cc
ataymano: As we discussed, there is a breaking change in CCB api (from unique id per slot to event id + slot index).
Because of that,
CCB methods which are using event_id + slot index as key support only v2 protocol, and 
CCB methods which are using unique slot ids support only v1 protocol.
v2 CCB methods are not exposed to client yet.

rlclientlib/logger/logger_facade.cc
ataymano: Please see the answer above. Not exposed yet.

rlclientlib/ranking_event.cc
ataymano: Yes, we can do that.

rlclientlib/ranking_event.cc
ataymano: only 2 times - legacy base event class (which should be removed soon) + new generic_event class. We probably can unify it somehow, but I feel it is safer to have paralle v2 implementation on this level. It gives us:
- less risk to break existing v1 implementation
- more flexibility to do things properly in v2 implementation without having dependency on legacy artifacts.

rlclientlib/ranking_event.h
ataymano: Yes, we can do that.

rlclientlib/serialization/payload_serializer.h
ataymano: good catch, thank you

rlclientlib/serialization/fb_serializer.h
ataymano: even more, it is probably better just to get rid of intermediate enum. will do that

rlclientlib/serialization/payload_serializer.cc
ataymano: fixing default case handling. Regarding cast, not sure if it is a good idea - these are 2 independent enums defined in different places - casting them to each other can introduce hidden bugs if they become inconsistent at some point.

rlclientlib/logger/event_logger.h
kumpera: I see your point, with configuration sections it becomes easier to pass an uniform configuration container that can be used. 

Right now, that would mean passing a config prefix and depend on conventions. Actually, how bad would this be?

rlclientlib/logger/logger_facade.cc
kumpera: Makes sense.
We need to mark as obsolete all v1 methods in the API(s) then.

ataymano: agree (but when v2 are exposed on API level - not in this PR)

rlclientlib/logger/event_logger.h
ataymano: maybe it is not that bad, but let's not mix it (it seems independent improvement). I'll create separate issue for that.

ataymano: https://github.com/VowpalWabbit/reinforcement_learning/issues/172

rlclientlib/ranking_event.cc
kumpera: Fair point on risk, makes sense.
I don't buy this is a temporary thing. V1 will creeps around for quite a while after v2 goes live, unfortunately.

rlclientlib/serialization/payload_serializer.cc
kumpera: Good point, if we can't eliminate one of them, a numeric cast is fragile long term.

ataymano:
Open question:
After refactoring of id for CCB, CCB and Slates payloads became identical. Do we still want to have separate schemas for them?

rlclientlib/logger/logger_facade.cc
lalo: will it be under a flag initially?

ataymano: not 100% sure what do you mean - can you please elaborate?

rlclientlib/serialization/payload_serializer.h
```diff
@@ -0,0 +1,135 @@
...
+    };
+
+    struct ccb_serializer : payload_serializer<generic_event::payload_type_t::PayloadType_CCB> {
+      static generic_event::payload_buffer_t event(const char* context, unsigned int flags, const std::vector<std::vector<uint32_t>>& action_ids,
+        const std::vector<std::vector<float>>& pdfs, const std::string& model_version) {
+        flatbuffers::FlatBufferBuilder fbb;
+        std::vector<flatbuffers::Offset<v2::SlotEvent>> slots;
+        for (size_t i = 0; i < action_ids.size(); i++)
+        {
+          slots.push_back(v2::CreateSlotEventDirect(fbb, &action_ids[i], &pdfs[i]));
```
lalo, line 60: emplace_back instead of push_back?

ataymano: what's the benefit?

rlclientlib/ranking_event.h
```diff
@@ -6,6 +6,8 @@
 #include "decision_response.h"
 #include "slates_response.h"
 
+#include <flatbuffers/flatbuffers.h>
```
homezcx, line 9: is this line needed?

rlclientlib/serialization/payload_serializer.cc
```diff
@@ -0,0 +1,16 @@
+#include "payload_serializer.h"
+
+namespace reinforcement_learning {
+  using namespace messages::flatbuff;
+  namespace logger {
+    int get_learning_mode(learning_mode mode_in, v2::LearningModeType& mode_out, api_status* status) {
+      v2::LearningModeType result;
+      switch (mode_in) {
+      case APPRENTICE: mode_out = v2::LearningModeType_Apprentice; return error_code::success;
```
homezcx, line 9: [nit] space

rlclientlib/live_model_impl.cc
```diff
@@ -77,7 +77,7 @@ namespace reinforcement_learning {
       RETURN_IF_FAIL(reset_action_order(response));
     }
 
-    RETURN_IF_FAIL(_ranking_logger->log(event_id, context, flags, response, status, _learning_mode));
+    RETURN_IF_FAIL(_ranking_logger->log(context, flags, response, status, _learning_mode));
```
homezcx, line 80: Trying to read this PR with the proposal in the OneNote. Let me know if I missed this - 

How is the V2 protocol enabled from config/live_model? Here - not seeing a _ranking_logger change when it gets created.

rlclientlib/serialization/payload_serializer.h
lalo: just a nit, depending on the objects being constructed there might be performance difference but your call

rlclientlib/serialization/fb_serializer.h
```diff
@@ -242,6 +248,33 @@ namespace reinforcement_learning { namespace logger {
...
+  struct fb_event_serializer<generic_event> {
+    using fb_event_t = v2::Event;
+    using offset_vector_t = typename std::vector<flatbuffers::Offset<fb_event_t>>;
+    using batch_builder_t = v2::EventBatchBuilder;
+
+    static size_t size_estimate(const generic_event& evt) {
+      return evt.get_payload().size();
+    }
+
+    static int serialize(generic_event& evt, flatbuffers::FlatBufferBuilder& builder,
```
lalo, line 261: why not add a simple unit test for all the new serializers?

rlclientlib/live_model_impl.cc
kumpera: this will come in a future PR

kumpera:
> 
> 
> Open question:
> After refactoring of id for CCB, CCB and Slates payloads became identical. Do we still want to have separate schemas for them?

Payload should be an opaque byte array as it's what will be required by the throughput changes.

ataymano:
> > Open question:
> > After refactoring of id for CCB, CCB and Slates payloads became identical. Do we still want to have separate schemas for them?
> 
> Payload should be an opaque byte array as it's what will be required by the throughput changes.
Not sure if it is relevant - there is layer when it is opaque. But at some point it is deserialized into problem specific schema (CB/CCB/Slates/...). Although CCB and slates represent different problems, their payload schemas are identical right now (the only difference is how reward assigned).

rlclientlib/ranking_event.h
ataymano: no, will remove it.

rlclientlib/serialization/fb_serializer.h
ataymano: good idea, working on it.

rlclientlib/serialization/payload_serializer.cc
ataymano: thank you, fixed

rlclientlib/serialization/fb_serializer.h
ataymano: done

rlclientlib/live_model_impl.cc
ataymano: actually config parameter was introduced here:
https://github.com/VowpalWabbit/reinforcement_learning/pull/163

rlclientlib/logger/logger_facade.cc
```diff
@@ -11,31 +11,55 @@ namespace reinforcement_learning {
...
 
-    int cb_logger_facade::log(const char* event_id, const char* context, unsigned int flags, const ranking_response& response, api_status* status, learning_mode learning_mode) {
+    int cb_logger_facade::log(const char* context, unsigned int flags, const ranking_response& response, api_status* status, learning_mode learning_mode) {
       switch (version) {
-        case 1: return v1->log(event_id, context, flags, response, status, learning_mode);
+        case 1: return v1->log(response.get_event_id(), context, flags, response, status, learning_mode);
+        case 2:
+          v2::LearningModeType lmt;
+          RETURN_IF_FAIL(get_learning_mode(learning_mode, lmt, status));
+          return v2->log(response.get_event_id(), serializer.event(context, flags, lmt, response), serializer.type, status);
```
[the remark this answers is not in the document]
ataymano: yes. Let me share some performance tests results - I do not see any performance drop. Probably because serialization here is not differ that much from extra copying of the data (in v1 code path)

rlclientlib/logger/logger_facade.cc
```diff
@@ -11,31 +11,55 @@ namespace reinforcement_learning {
...
+      c.get_int(name::INTERACTION_SEND_BATCH_INTERVAL_MS, 1000),
+      c.get_int(name::INTERACTION_SEND_QUEUE_MAX_CAPACITY_KB, 16 * 1024) * 1024,
+      c.get(name::QUEUE_MODE, "DROP"),
+      watchdog,
+      time_provider,
+      perror_cb) : nullptr) {
     }
 
     int cb_logger_facade::init(api_status* status) {
       switch (version) {
```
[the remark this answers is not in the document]
ataymano: Also, if we will make it templated we need either:
1) propagate template parameter to live_model_impl and make it header-only (which seems undesirable)
2) make v1 and v2 loggers share some interface - it replaces current "if" with virtual call (which also has performance overhead, which seems minor compare to rest of inference in both cases)

kumpera (approved):
LGTM now

rlclientlib/logger/logger_facade.h
```diff
@@ -30,11 +31,14 @@ namespace reinforcement_learning
...
       int init(api_status* status);
 
-      int log(const char* event_id, const char* context, unsigned int flags, const ranking_response& response, api_status* status, learning_mode learning_mode = ONLINE);
+      int log(const char* context, unsigned int flags, const ranking_response& response, api_status* status, learning_mode learning_mode = ONLINE);
     
     private:
       const int version;
       const std::unique_ptr<interaction_logger> v1;
+      const std::unique_ptr<generic_event_logger> v2;
+
```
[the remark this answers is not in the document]
ataymano: fixed

rlclientlib/logger/logger_facade.h
```diff
@@ -30,11 +31,14 @@ namespace reinforcement_learning
...
+      const cb_serializer serializer;
```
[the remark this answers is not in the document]
ataymano: fixed

rlclientlib/logger/logger_facade.cc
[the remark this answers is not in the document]
ataymano: image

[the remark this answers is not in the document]
ataymano: each point - latency averaged by 10000 choose_rank calls.

rajan-chari:
> each point - latency averaged by 10000 choose_rank calls.
 

Very cool.  Thanks for the analysis.

rajan-chari:
> > each point - latency averaged by 10000 choose_rank calls.
 

Very cool.  Thanks for the analysis.   

[Not for this PR] It would be good to avoid this copy in inference thread. Perhaps with example builder, the lifetime of context can be managed by rlclientlib.

rajan-chari (approved):

homezcx (approved):

PR #170, opened 2020-08-19
Pull down latest for vowpal_wabbit

SchuylerGoodman:
* Update vowpal_wabbit submodule to 2c1597778a1e25492ad7ada8abd9cbbf713bee3e


kumpera:
We pulled latest VW on more recent PRs.
Sorry for missing this PR for so long.
Next time you need to bump VW, it would be great if you could mention what do you need from VW.

PR #176, opened 2020-09-02
Implement call stack recording for error propagation

jackgerrits:
This is a first draft at call stack recording for error propagation. Open to any and all comments/feedback/criticism.

This does not apply it to the codebase, there will need to be a migration following this to make use of it.


jackgerrits:
There are OS apis we could use instead of this hand created approach, closing as this was a PR for feedback

PR #178, opened 2020-09-03
[IDEA] Use a transparent type to achieve nodiscard semantics

jackgerrits:
This is just an idea as a code sample.

[[nodiscard]] is very helpful, especially for error code values. However, since the interface just uses ints right now we can't easily make use of it. There are several ways we could:

1. Mark everything [[nodiscard]] manually
    - Cumbersome
2. Use a macro such as `API` -> `[[nodiscard]] int`
    - Still just a raw int
    - It's a macro...
3. Use an enum that is marked [[nodiscard]]
   - I am a big fan of using enums for error codes
4. Use a transparent type that is marked [[nodiscard]]
   - This has implicit operators that means it behaves similar to an int but can't be discarded as easily
   - It does use implicits though which isn't great

### DIFF
--- a/include/api_status.h
+++ b/include/api_status.h
@@ -10,6 +10,15 @@
 #include <string>
 #include <sstream>
 #include "err_constants.h"
+#include "future_compat.h"
+
+struct RL_ATTR(nodiscard) error_code_no_discard
+{
+  error_code_no_discard(int value) : _value(value) {}
+  operator int() const { return _value; }
+  int value() const { return _value; }
+  int _value;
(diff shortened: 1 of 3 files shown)


jackgerrits:
The plan here is to use a macro instead

kumpera:
The issue I see with using structs is that it's an ABI breaking change.

jackgerrits:
Yeah the result of the discussion was to use a macro to enforce it. I agree struct seems weird along with the ABI change...

PR #180, opened 2020-09-10
Add compression config for dedup + zstd

cheng-tan:


rlclientlib/logger/event_logger.h
```diff
@@ -192,5 +192,8 @@ class slates_logger : public event_logger<slates_decision_event> {
     {}
 
     int log(const char* event_id, generic_event::payload_buffer_t&& payload, generic_event::payload_type_t type, api_status* status);
-  };
+
+    private:
+      const char* compression_mode;
```
jackgerrits, line 197: Can this be an enum?

jackgerrits (changes requested):

rlclientlib/logger/event_logger.h
```diff
@@ -170,30 +170,40 @@ class slates_logger : public event_logger<slates_decision_event> {
     int report_action_taken(const char* event_id, api_status* status);
   };
 
-  class generic_event_logger : public event_logger<generic_event> {
+  enum compression_mode_enum
```
jackgerrits, line 173: This should be an enum class so that is namespaced and type checked properly
```suggestion
  enum class compression_mode_enum
```

kumpera (commented):
Shouldn't we use names that convey this is dedup+zstd?
Both in the option name and its value

cheng-tan:
> Shouldn't we use names that convey this is dedup+zstd?
> Both in the option name and its value

Will it be better if changing the name to: `compression_dedup_mode`
value to:  `no_compression_dedup` and `zstd+dedup`

kumpera:
That naming suggests that this is a boolean flag. We can just use the same as the protocol, "identity"  which actually tells what's doing instead of what's not doing.

Think of the name if we add a 3rd one later.

kumpera (approved):
LGTM

rlclientlib/logger/event_logger.cc
```diff
@@ -3,6 +3,17 @@
...
 namespace reinforcement_learning { namespace logger {
+  content_encoding_enum to_content_encoding_enum(const char *content_encoding)
+  {
+    if (std::strcmp(content_encoding, value::CONTENT_ENCODING_IDENTITY) == 0)
+    {
+      return content_encoding_enum::IDENTITY;
+    }
+    else
+    {
+      return content_encoding_enum::ZSTD_AND_DEDUP;
```
jackgerrits, line 14: If `content_encoding` is not a known value you will get `content_encoding_enum::ZSTD_AND_DEDUP`. Does it make sense to fail instead? If someone misconfigures this, I think it's not intuitive how it is

rlclientlib/logger/event_logger.cc
```diff
@@ -3,6 +3,17 @@
 #include "err_constants.h"
 #include "time_helper.h"
 namespace reinforcement_learning { namespace logger {
+  content_encoding_enum to_content_encoding_enum(const char *content_encoding)
```
jackgerrits, line 6: nit:
```suggestion
  content_encoding_enum to_content_encoding_enum(const char* content_encoding)
```

rlclientlib/logger/event_logger.cc
kumpera: Good catch, we should either fail or pick identity.

cheng-tan:
@jackgerrits & @kumpera can you take another look of this PR, all comments have been addressed.

kumpera (approved):
Looks good.
A few things for followup work:
- We should emit a warning and/or error if the config content encoding is not recognized.
- Logger constructors are monsters now and we should clean that up
- enums are consistently named with a _t suffix and not _enum.

PR #181, opened 2020-09-11
Update event schema to include content_encoding field

cheng-tan:
Adding in this field to determine using current mode or action dedup + zstd.

### DIFF
--- a/rlclientlib/serialization/fb_serializer.h
+++ b/rlclientlib/serialization/fb_serializer.h
@@ -15,7 +15,6 @@
 
 #include "logger/message_type.h"
 #include "err_constants.h"
-#include <iostream>
 
 using namespace reinforcement_learning::messages::flatbuff;
 namespace reinforcement_learning { namespace logger {
@@ -234,8 +233,7 @@ namespace reinforcement_learning { namespace logger {
       return;
     }
 
-    void add_header(typename serializer_t::batch_builder_t &batch_builder)
-    {
+    void add_header(typename serializer_t::batch_builder_t &batch_builder) {
       return;
     }
 
@@ -259,7 +257,7 @@ namespace reinforcement_learning { namespace logger {
     flatbuffers::FlatBufferBuilder _builder;
     buffer_t& _buffer;
     content_encoding_enum _content_encoding;
-    flatbuffers::Offset<reinforcement_learning::messages::flatbuff::v2::BatchMetadata> _batch_metadata_offset;
+    flatbuffers::Offset<v2::BatchMetadata> _batch_metadata_offset;
   };
 
   template <>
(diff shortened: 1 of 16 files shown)


kumpera (changes requested):
Making this an enum is problematic because we might decide down the line to pick just dedup or compression. It should be a string.

Additionally, we didn't change rlclientlib to send the default value of identity.

cheng-tan:
> Making this an enum is problematic because we might decide down the line to pick just dedup or compression. It should be a string.
> 
> Additionally, we didn't change rlclientlib to send the default value of identity.

Will set the default value to identity.

Regards to enum vs string, if we only want to pick dedup or compression, can we just add another option in the enum? won't string be more error prone if we got some random string not supported?

kumpera (changes requested):
Content encoding should be a string.
This PR needs tests that we're encoding the right value.

rlclientlib/serialization/fb_serializer.h
```diff
@@ -273,6 +275,59 @@ namespace reinforcement_learning { namespace logger {
...
+      _event_offsets.push_back(offset);
+      return error_code::success;
+    }
+
+    uint64_t size() const { return _builder.GetSize(); }
+
+    const char * getTextForEnum(content_encoding_enum enumVal)
+    {
+      if (enumVal == content_encoding_enum::ZSTD_AND_DEDUP) {
+        return "ZSTD_AND_DEDUP";
```
kumpera, line 301: We should use the config constants for those two.

rlclientlib/serialization/fb_serializer.h
```diff
@@ -273,6 +275,59 @@ namespace reinforcement_learning { namespace logger {
...
+    {
+      flatbuffers::Offset<typename serializer_t::fb_event_t> offset;
+      RETURN_IF_FAIL(serializer_t::serialize(evt, _builder, offset, status));
+      _event_offsets.push_back(offset);
+      return error_code::success;
+    }
+
+    uint64_t size() const { return _builder.GetSize(); }
+
+    const char * getTextForEnum(content_encoding_enum enumVal)
```
kumpera, line 298: This should be another function in context_encoding.h

unit_test/fb_serializer_test.cc
```diff
@@ -1,7 +1,7 @@
 #define BOOST_TEST_DYN_LINK
 #ifdef STAND_ALONE
 #   define BOOST_TEST_MODULE Main
-#endif
+#endif
```
kumpera, line 4: nit: line damage

rlclientlib/serialization/fb_serializer.h
```diff
@@ -273,6 +275,59 @@ namespace reinforcement_learning { namespace logger {
...
+      } else {
+        return "IDENTITY";
+      }
+    }
+
+    void finalize()
+    {
+      auto event_offsets = _builder.CreateVector(_event_offsets);
+
+            const auto batch_metadata_offset = v2::CreateBatchMetadataDirect(_builder, getTextForEnum(_content_encoding));
```
kumpera, line 311: nit: bad identation. @jackgerrits don't we have a bot for this?

kumpera (commented):
The PR lacks test for content encoding in serialized data.
Another concern is the code duplication in fb_collection_serializer that can be avoided by introducing another templatized type to deal with adding the header.

We need to make sure, and test for, that we can't do any encoding but IDENTITY with schema v1.

rlclientlib/serialization/fb_serializer.h
cheng-tan: This will be entirely removed. I'm changing content_encoding to string instead of enum.

kumpera (approved):

PR #183, opened 2020-09-15
Schema v2 improvements.

kumpera:
Modify schema v2 with the following changes:

Unify CCB and Slates under MultiSlotDecision.
Simplify OutcomeSingle to make resulting code a lot nicer to use.
Action_taken and index become top level, optional fields instead of separate tables
part of the union.

This simplifies _a lot_ how those fields are extracted by the joiner by not
requiring lots of union type testing.


ataymano:
1) I like the idea of re-using schema between CCB and Slates. But it is still 2 different statements of the problem and 2 different APIs. Do we maybe want to secure ourselves with extra abstraction layer (in case if they diverge in future)? I.E.
table MultiSlotEvent { ... your code... };
table CcbEvent {event:MultiSloEvent};
table SlatesEvent {event:MultiSlotEvent};

2) Regarding outcome - it looks cool, but maybe we want to consider multistep changes now. Please see draft here: https://github.com/ataymano/reinforcement_learning/blob/ataymano/multistep/rlclientlib/schema/v2/OutcomeSingle.fbs
So it is going to have string as id. Given that, do we want to:
- rename OutcomeSingle to Outcome
- replace int index; with union {int, string} index;
?

kumpera:
On (1). The APIs are only different due to the event id formulation. In the v2 world, the APIs will be identical and we can in fact use the slates API for CCB albeit the funky name.

I buy your argument of having separate evolution. I'll assume we're doing backwards compat evolution where we don't introduce a new message type. There's an alternative design that is future proof that doesn't require us to compromise on the design today:

```
table CcbSpecificData { }
table SlatesSpecificData { }
union ProblemSpecificData { CcbSpecificData, SlatesSpecificData }

table MultiSlotEvent {
    context:[ubyte];               // context
    slots:[SlotEvent];        // actions and probabilities
    model_id:string;               // model ID
    deferred_action:bool = false;  // delayed activation flag
    specific: ProblemSpecificData;
}
```

This allow us later to add problem specific data if/when needed. We already have a couple features missing from ccb/slates compared to cb (I.E. logging and imitation mode) and my argument against your proposal is that it will act as a forcing function for us implement them only to one problem and not both.

It would great to get feedback from @JohnLangford or @pmineiro into this.

On (2), makes sense.

ataymano:
(1) Sounds good to me.

kumpera:
Updated the PR with support for multi-step reward plus fixed/extended test suite.

rlclientlib/serialization/payload_serializer.h
```diff
@@ -64,69 +63,68 @@ namespace reinforcement_learning {
...
-        const auto evt = v2::CreateStringEventIndexedDirect(fbb, outcome, index).Union();
-        auto fb = v2::CreateOutcomeSingleEvent(fbb, v2::OutcomeSingleEventBody_StringEventIndexed, evt);
+        const auto evt = fbb.CreateString(outcome).Union();
+        const auto idx = v2::CreateNumericIndex(fbb, index).Union();
+        auto fb = v2::CreateOutcomeEvent(fbb, v2::OutcomeValue_literal, evt, v2::IndexValue_numeric, idx);
+        fbb.Finish(fb);
+        return fbb.Release();
+      }
+
+      static generic_event::payload_buffer_t string_event(const char* index, const char* outcome) {
```
ataymano, line 116: potential ambiguity between int / char* (as you pointed out before)

rlclientlib/schema/v2/Metadata.fbs
```diff
@@ -1,6 +1,6 @@
 namespace reinforcement_learning.messages.flatbuff.v2;
 
-enum PayloadType : ubyte { CB, CCB, Slates, OutcomeSingle }
+enum PayloadType : ubyte { CB, MultiSlot, OutcomeSingle }
```
ataymano, line 3: OutcomeSingle -> Outcome

rlclientlib/serialization/payload_serializer.h
kumpera: I looked at this closely and this doesn't happen. Int won't be implicitly cast to const char*

ataymano (approved):

PR #185, opened 2020-09-22
Replace CPPRestSDK JSON parsing with RapidJSON

SchuylerGoodman:
**Summary**

This is the first of several PRs that will enable building this project without the cpprest dependency.

**Changes**

* Replaced JSON parsing implementation in `config_utility.cc` and `context_helper.cc` with rapidjson.
* Added additional unit tests to `configuration_test.cc`
  * `configuration_parse_invalid_json_object_element_expect_error` - Ensure that an object element at the root level will throw a json_parse_error
  * `configuration_parse_invalid_json_array_element_expect_error` - Ensure that an object element at the root level will throw a json_parse_error
  * These are necessary because rapidjson does not make it easy to convert the parsed json elements back into strings (which is how `configuration.cc` works), and objects and array elements are not supported for this function, anyways

**Test Coverage**

Changes are 100% covered by tests in both `configuration_test.cc` and `json_context_parse_test`


rlclientlib/utility/context_helper.cc
```diff
@@ -71,27 +61,110 @@ namespace reinforcement_learning { namespace utility {
...
+   * and that _multi comes before _slots
+   * 
+   * \param context : String with context json
+   * \param trace   : Pointer to the trace logger
+   * \param status  : Pointer to api_status object that contains an error code and error description in
+   *                  case of failure
+   * \return  error_code::success if there are no errors.  If there are errors then the error code is
+   *          returned.
+   */
   int validate_multi_before_slots(const char *context, i_trace* trace, api_status* status)
```
SchuylerGoodman, line 168: There's a comment in `live_model_impl.cc` that this was necessary due to limitations with the parser. Since this PR changes the parser, this function might not be necessary anymore. What exactly was the limitation?

kumpera: This is a limitation in VW itself and not the one on rlclientlib.

rlclientlib/utility/context_helper.cc
```diff
@@ -1,65 +1,55 @@
...
-      if ( json_obj.has_array_field(slots) ) {
-        auto const arr = json_obj.at(slots).as_array();
-        for (int i = 0; i < arr.size(); i++) {
-          auto current = arr.at(i);
-          if(current.has_string_field(event_id))
-          {
-            auto event_id_string = current.at(event_id).as_string();
+      const Value::ConstMemberIterator itr = obj.FindMember(slots);
+      if (itr != obj.MemberEnd() && itr->value.IsArray()) {
+        const auto arr = itr->value.GetArray();
```
kumpera, line 42: shouldn't this be a ref?

kumpera (approved):
LGTM.

@jackgerrits @lokitoth both of you worked with rapidjson, so it would be nice for you to take a look at this PR.

rlclientlib/utility/context_helper.cc
SchuylerGoodman: Probably. Are you referring to `itr` or `arr`? Or possibly both?

rlclientlib/utility/context_helper.cc
```diff
@@ -1,65 +1,55 @@
...
+   * \param event_ids : Reference to the mapping from slot index to event ID string, results will be
+   *                    put in here
+   * \param trace     : Pointer to the trace logger
+   * \param status    : Pointer to api_status object that contains an error code and error description in
+   *                    case of failure
    * \return  error_code::success if there are no errors.  If there are errors then the error code is
    *          returned.
    */
-  int get_action_count(size_t& count, const char *context, i_trace* trace, api_status* status) {
+  int get_event_ids(const char* context, std::map<size_t, std::string>& event_ids, i_trace* trace, api_status* status) {
```
rajban2017, line 31: Do you need the keys to be in sorted order ? If not, I would recommend to use std::unordered_map() which give O(constant) for insertion and check.

rlclientlib/utility/config_utility.cc
```diff
@@ -110,23 +109,35 @@ namespace reinforcement_learning { namespace utility { namespace config {
...
+
+      const char *string_value;
+      if (prop_value.IsString()) {
+        string_value = prop_value.GetString();
+      }
+      else if (prop_value.IsBool()) {
+        string_value = prop_value.GetBool() ? "true" : "false";
+      }
+      else {
+        RETURN_ERROR_LS(trace, status, json_parse_error) << "Invalid json type found: " << prop_value.GetType();
```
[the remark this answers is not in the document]
SchuylerGoodman: I was able to find a flag that makes rapidjson parse numbers as strings, so those are handled by the `IsString()` logic. I was not able to get it to parse `bool`, `object`, or `array` as strings, so I had to handle those separately.

rlclientlib/utility/context_helper.cc
SchuylerGoodman: Hmm, I think that would be fine, but it was using std::map before, as well. @lokitoth would this be a valid optimization?

lokitoth: At a guess, based on where this code is being used: The expectation is that the number of entries to this map will be small (as it is capped by the number of slots in a given CCB request), and immediately after getting out of this call, it gets iterated, start-to-finish.

(A better solution here might be to turn it into a "foreach_event_id" template, to avoid constructing this map in the first place, but that is out of the scope of this change, I think.)

cc @jackgerrits who may have more context about the history here

rajban2017: Yeah, I am fine with keeping it std::map, just wanted to bring it up because of potential runtime complexity improvement here.

rajban2017 (approved):

kiprop12 (approved):

rlclientlib/utility/config_utility.cc
```diff
@@ -1,18 +1,17 @@
...
 #include "constants.h"
 #include "err_constants.h"
 #include "api_status.h"
 #include "str_util.h"
 #include "trace_logger.h"
-#include "../model_mgmt/restapi_data_transport.h"
 
-using namespace web;
-using namespace utility::conversions; // string conversions utilities
+using namespace rapidjson;
```
jackgerrits, line 14: I'd prefer to either shorted this to `rj` or just use the full namespace.

jackgerrits (approved):

rlclientlib/utility/config_utility.cc
SchuylerGoodman: Done

PR #187, opened 2020-09-23
Add live_model entry points for indexed reward.

kumpera:
This adds indexed rewards to LiveModel.
It doesn't include C# or python bindings.

### DIFF
--- a/include/live_model.h
+++ b/include/live_model.h
@@ -220,6 +220,54 @@ namespace reinforcement_learning {
      */
     int report_outcome(const char* event_id, float outcome, api_status* status= nullptr);
 
+    /**
+     * @brief Report partial outcome of a decision based on an index.
+     *
+     * @param event_id  The unique event_id used when choosing an action should be presented here.  This is so that
+     *                  the action taken can be matched with feedback received.
+     * @param outcome Outcome as float.
+     * @param index Index of the partial outcome.
+     * @param status  Optional field with detailed string description if there is an error
+     * @return int Return error code.  This will also be returned in the api_status object
+     */
+    int report_outcome(const char* event_id, float outcome, int index, api_status* status= nullptr);
+
+    /**
+     * @brief Report partial outcome of a decision based on an index.
+     *
+     * @param event_id  The unique event_id used when choosing an action should be presented here.  This is so that
+     *                  the action taken can be matched with feedback received.
+     * @param outcome Outcome as float.
+     * @param index Index of the partial outcome.
+     * @param status  Optional field with detailed string description if there is an error
+     * @return int Return error code.  This will also be returned in the api_status object
+     */
+    int report_outcome(const char* event_id, float outcome, const char *index, api_status* status= nullptr);
+
+    /**
+     * @brief Report partial outcome of a decision based on an index.
+     *
+     * @param event_id  The unique event_id used when choosing an action should be presented here.  This is so that
+     *                  the action taken can be matched with feedback received.
+     * @param outcome Outcome as float.
+     * @param index Index of the partial outcome.
+     * @param status  Optional field with detailed string description if there is an error
+     * @return int Return error code.  This will also be returned in the api_status object
+     */
+    int report_outcome(const char* event_id, const char* outcome, int index, api_status* status= nullptr);
+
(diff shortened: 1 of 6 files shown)


include/live_model.h
```diff
@@ -220,6 +220,54 @@ namespace reinforcement_learning {
...
+     * @brief Report partial outcome of a decision based on an index.
+     *
+     * @param event_id  The unique event_id used when choosing an action should be presented here.  This is so that
+     *                  the action taken can be matched with feedback received.
+     * @param outcome Outcome as float.
+     * @param index Index of the partial outcome.
+     * @param status  Optional field with detailed string description if there is an error
+     * @return int Return error code.  This will also be returned in the api_status object
+     */
+    int report_outcome(const char* event_id, float outcome, int index, api_status* status= nullptr);
```
ataymano, line 233: [nit] do you think it is more intuitive to reorder to  event_id, index, outcome (event_id and index logically more like "Key" while outcome - "Value")

include/live_model.h
```diff
@@ -220,6 +220,54 @@ namespace reinforcement_learning {
...
+     * @brief Report partial outcome of a decision based on an index.
+     *
+     * @param event_id  The unique event_id used when choosing an action should be presented here.  This is so that
+     *                  the action taken can be matched with feedback received.
+     * @param outcome Outcome as float.
+     * @param index Index of the partial outcome.
+     * @param status  Optional field with detailed string description if there is an error
+     * @return int Return error code.  This will also be returned in the api_status object
+     */
+    int report_outcome(const char* event_id, float outcome, const char *index, api_status* status= nullptr);
```
ataymano, line 245: We maybe need to think about naming here. For multistep case it make sense to rename event_id (from envelope) to episode_id or session_id, and index to event_id (but in this case it brings extra confusion with non-multistep methods)

include/live_model.h
kumpera: Yeah, makes sense.

include/live_model.h
kumpera: Hmmm, yeah.
Actually I'm not that super excited with this API sprawl, what about introducing some sort of "struct outcome_id" and have a single method instead?

rlclientlib/logger/logger_facade.cc
```diff
@@ -161,6 +161,20 @@ namespace reinforcement_learning {
       }
     }
 
+    int observation_logger_facade::log(const char* event_id, const char* index, float outcome, api_status* status) {
```
olgavrou, line 164: nit: following this down to here the arguments have been passed as: `outcome, index`, can we keep that here too instead of switching to `index, outcome`?

include/live_model.h
olgavrou: Or couldn't we template this `report_outcome` method same as we template the `report_outcome_internal`?

olgavrou:
Some unit test cases for the new `report_outcome` method would be good

include/live_model.h
ataymano: This is top level API call, so I am not sure if it is a good idea to template it.

jackgerrits (approved):

include/live_model.h
ataymano: Regarding outcome_id structure - really not sure:
Although episode_id (join_id) + index are somehow logically represents outcome id, if we move to example builder in the future, these arguments should be structured in a way that is more aligned with schema: {join_id, {index, value}}.
So I would leave arguments structure plain for now.

kumpera: I prototyped an `outcome_id` struct and it felt like over-engineering.

Let's agree to revisit this once we're either adding one more parameter to report_outcome or another type to the existing pair. As long as the strings don't have any internal structure, I'm happy.

rlclientlib/logger/logger_facade.cc
kumpera: good catch. fixed the order and naming to be consistent everywhere.

include/live_model.h
ataymano: sure sounds good to me

ataymano (approved):

PR #188, opened 2020-09-23
Introduce i_model::model_type() and rename slates to multi_slot.

kumpera:


include/model_mgmt.h
```diff
@@ -61,13 +61,21 @@ namespace reinforcement_learning { namespace model_management {
...
     class i_model {
     public:
       virtual int update(const model_data& data, bool& model_ready, api_status* status = nullptr) = 0;
       virtual int choose_rank(uint64_t rnd_seed, const char* features, std::vector<int>& action_ids, std::vector<float>& action_pdf, std::string& model_version, api_status* status = nullptr) = 0;
       virtual int request_decision(const std::vector<const char*>& event_ids, const char* features, std::vector<std::vector<uint32_t>>& actions_ids, std::vector<std::vector<float>>& action_pdfs, std::string& model_version, api_status* status = nullptr) = 0;
-      virtual int request_slates_decision(const char* event_id, uint32_t slot_count, const char* features, std::vector<std::vector<uint32_t>>& actions_ids, std::vector<std::vector<float>>& action_pdfs, std::string& model_version, api_status* status = nullptr) = 0;
+      virtual int request_multi_slot_decision(const char* event_id, uint32_t slot_count, const char* features, std::vector<std::vector<uint32_t>>& actions_ids, std::vector<std::vector<float>>& action_pdfs, std::string& model_version, api_status* status = nullptr) = 0;
+      virtual model_type_t model_type() = 0;
```
ataymano, line 78: () const = 0;

include/model_mgmt.h
```diff
@@ -61,13 +61,21 @@ namespace reinforcement_learning { namespace model_management {
...
     class i_model {
     public:
       virtual int update(const model_data& data, bool& model_ready, api_status* status = nullptr) = 0;
       virtual int choose_rank(uint64_t rnd_seed, const char* features, std::vector<int>& action_ids, std::vector<float>& action_pdf, std::string& model_version, api_status* status = nullptr) = 0;
       virtual int request_decision(const std::vector<const char*>& event_ids, const char* features, std::vector<std::vector<uint32_t>>& actions_ids, std::vector<std::vector<float>>& action_pdfs, std::string& model_version, api_status* status = nullptr) = 0;
-      virtual int request_slates_decision(const char* event_id, uint32_t slot_count, const char* features, std::vector<std::vector<uint32_t>>& actions_ids, std::vector<std::vector<float>>& action_pdfs, std::string& model_version, api_status* status = nullptr) = 0;
+      virtual int request_multi_slot_decision(const char* event_id, uint32_t slot_count, const char* features, std::vector<std::vector<uint32_t>>& actions_ids, std::vector<std::vector<float>>& action_pdfs, std::string& model_version, api_status* status = nullptr) = 0;
```
ataymano, line 77: Actually, maybe just request_decision is better?
- signature is explicit about slots
- it has different signature from legacy ccb one (vector<char*> vs const char* for id)

kumpera: 👍

olgavrou: Not sure I agree with this, since the method is not going away, and since we have many `request_decision` methods already, it might be easier to navigate the code having the method name explicitly state that it is doing multi slot than having to look at the arguments passed to understand if we are dealing with ccb or slates

kumpera: I'd agree with Olga here, not a big fan of overloading. All `request_decision` methods are deprecated/on the way out, so let's avoid that name until the current usage is gone.

PR #190, opened 2020-09-25
Continuous actions

olgavrou:
Introduces continuous actions to rlclientlib

next step: c# bindings


examples/rl_sim_cpp/CMakeLists.txt
```diff
@@ -1,8 +1,8 @@
 add_executable(rl_sim_cpp.out
   main.cc
   person.cc
+  robot_joint.cc
   rl_sim.cc
-  simulation_stats.cc
```
olgavrou, line 5: templatized `simulation_stats` to accept continuous actions

rlclientlib/live_model_impl.cc
```diff
@@ -100,6 +100,48 @@ namespace reinforcement_learning {
       status);
   }
 
+  int live_model_impl::request_continuous_action(const char* event_id, const char* context, unsigned int flags, continuous_action_response& response, api_status* status)
+  {
+    if (!_configuration.get_bool(name::CONTINUOUS_ACTIONS_ENABLED, value::DEFAULT_CONTINUOUS_ACTIONS_ENABLED))
```
ataymano, line 105: Why do we need explicit flag? We already have model_type() which seems to be answering this question.

olgavrou: It's a safeguard to make sure continuous actions is not being used until changes are implemented in the joiner/trainer also

olgavrou: But also allows us to turn it on for testing etc, where as if we just use the model type we can't enable/disable continuous actions usage.

rlclientlib/live_model_impl.h
```diff
@@ -87,6 +90,7 @@ namespace reinforcement_learning
     std::unique_ptr<model_management::i_data_transport> _transport{nullptr};
     std::unique_ptr<model_management::i_model> _model{nullptr};
     std::unique_ptr<logger::cb_logger_facade> _ranking_logger{nullptr};
+    std::unique_ptr<logger::ca_logger_facade> _continuous_action_logger{nullptr};
```
ataymano, line 93: Not sure if want to have it in this PR, but:
Since we are working on new loggers which are not having legacy implementation like ca and multistep (and v2 is unified), we probably can refactor this code like this:
class interaction_logger_facade{
...
const std::unique_ptr<interaction_logger> _v1_cb;
const std::unique_ptr<ccb_logger> _v1_ccb;
...
const std::unique_ptr<generic_logger> _v2;
}
It will at reduce copy-paste around v2 + localize legacy code inside logger_facade v1 branches.

rlclientlib/schema/v2/Metadata.fbs
```diff
@@ -1,6 +1,6 @@
 namespace reinforcement_learning.messages.flatbuff.v2;
 
-enum PayloadType : ubyte { CB, CCB, Slates, Outcome }
+enum PayloadType : ubyte { CB, CA, CCB, Slates, Outcome }
```
ataymano, line 3: In practice we haven't released anything yet, so it does not really matter, but we probably should append new enum valuees to the end in that kind of cases.

include/continuous_action_response.h
```diff
@@ -0,0 +1,42 @@
...
+
+#include <cstddef>
+#include <iterator>
+#include <vector>
+#include <string>
+
+namespace reinforcement_learning {
+  class api_status;
+
+  class continuous_action_response {
```
kumpera, line 11: Types and methods in the public API must be documented.

include/continuous_action_response.h
```diff
@@ -0,0 +1,42 @@
...
+  public:
+    continuous_action_response() = default;
+    ~continuous_action_response() = default;
+
+    float get_chosen_action() const;
+    void set_chosen_action(float action);
+
+    float get_chosen_action_pdf_value() const;
+    void set_chosen_action_pdf_value(float pdf_value);
```
kumpera, line 20: Don't continuous actions have mass and not density?

include/continuous_action_response.h
```diff
@@ -0,0 +1,42 @@
...
+
+    void set_event_id(const char* event_id);
```
kumpera, line 22: Event it should be a constructor argument to be consistent with the rest of rlclientlib.

include/model_mgmt.h
```diff
@@ -64,6 +64,7 @@ namespace reinforcement_learning { namespace model_management {
     enum class model_type_t {
       UNKNOWN,
       CB,
+      CA,
```
kumpera, line 67: Enum entries must always be append otherwise it's a breaking change.

rlclientlib/CMakeLists.txt
```diff
@@ -1,157 +1,160 @@
-set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/Modules/")
```
kumpera, line 1: Line damage in this file, please avoid it as no one can tell whether there was a change or not.

rlclientlib/live_model_impl.cc
kumpera: We don't need such safeguard, APS will work fine if we send CA events to it.
This extra required knob will lead to fragility down the line.

rlclientlib/live_model_impl.h
kumpera: Supporting CA under v1 will be harder than needed.

kumpera (changes requested):
This PR is humongous and has multiple issues.

This style of development of dropping a huge change of the fence is undesirable for many reasons:

- any changes in core assumptions will lead to a lot more waster work than needed.
- incremental work usually leads to less rigid structures
- it's easier to review
- by merging changes in smaller pieces it reduces the chance of merge conflicts down the line.

include/continuous_action_response.h
olgavrou: Nope, continuous actions have density

olgavrou:
> Supporting CA under v1 will be harder than needed.

@kumpera not sure what you mean here, I haven't added support for v1 for continuous actions

rlclientlib/CMakeLists.txt
olgavrou: thanks, fixed

include/model_mgmt.h
olgavrou: Changed to be appended

include/continuous_action_response.h
olgavrou: Done

include/continuous_action_response.h
olgavrou: I can add it but as an fyi `ranking_response.h` is the only response api that has documentation

rlclientlib/live_model_impl.cc
olgavrou: Good to know, removed

rlclientlib/schema/v2/Metadata.fbs
olgavrou: Appended to end

include/continuous_action_response.h
olgavrou: Done - documented public api methods in `continuous_action_response.h` (also cleaned up a bit the move/copy constructor/operators for that object).

Also added documentation for all of the new `request_continuous_action` methods in `live_model.h`. 

I can document the ones in `slates_response.h` and `decision_response.h` in a separate PR for consistency.

rlclientlib/live_model_impl.h
olgavrou: I can look at this in a separate PR once the loggers are a bit more stabilized (regarding https://github.com/VowpalWabbit/reinforcement_learning/pull/192#discussion_r496224392)

ataymano: sure, sounds good.

ataymano (approved):

include/continuous_action_response.h
kumpera: I'm documenting slates as part of another PR, thanks for the offer though :)

kumpera (approved):
LGTM.
The situation around loggers should be looked into as I believe with schema v2 we don't need to pile up so many of them.

PR #192, opened 2020-09-28
Rework slates API to support CCB under schema v2. Add example_gen utility.

kumpera:
This PR introduce the last batch of changes to have

### DIFF
--- a/rlclientlib/live_model_impl.cc
+++ b/rlclientlib/live_model_impl.cc
@@ -286,7 +286,8 @@ namespace reinforcement_learning {
     _t_factory{ t_factory },
     _m_factory{ m_factory },
     _sender_factory{ sender_factory },
-    _time_provider_factory{ time_provider_factory }
+    _time_provider_factory{ time_provider_factory },
+    _protocol_version(_configuration.get_int(name::PROTOCOL_VERSION, value::DEFAULT_PROTOCOL_VERSION))
   {
     // If there is no user supplied error callback, supply a default one that does nothing but report unhandled background errors.
     if (fn == nullptr) {
@@ -298,7 +299,6 @@ namespace reinforcement_learning {
     }
 
     _learning_mode = learning::to_learning_mode(_configuration.get(name::LEARNING_MODE, value::LEARNING_MODE_ONLINE));
-    _protocol_version = _configuration.get_int(name::PROTOCOL_VERSION, value::DEFAULT_PROTOCOL_VERSION);
   }
 
   int live_model_impl::init_trace(api_status* status) {
--- a/rlclientlib/live_model_impl.cc
+++ b/rlclientlib/live_model_impl.cc
@@ -170,6 +170,10 @@ namespace reinforcement_learning {
       return error_code::not_supported;
     }
 
+    if(_protocol_version == 1 && _model->model_type() != model_management::model_type_t::SLATES) {
+      RETURN_ERROR_ARG(_trace_logger.get(), status, protocol_not_supported, "Only Slates models are supported under legacy v1 protocol");
+    }
+
     resp.clear();
     //clear previous errors if any
     api_status::try_clear(status);
@@ -294,6 +298,7 @@ namespace reinforcement_learning {
     }
 
     _learning_mode = learning::to_learning_mode(_configuration.get(name::LEARNING_MODE, value::LEARNING_MODE_ONLINE));
+    _protocol_version = _configuration.get_int(name::PROTOCOL_VERSION, value::DEFAULT_PROTOCOL_VERSION);
   }
 
   int live_model_impl::init_trace(api_status* status) {
--- a/rlclientlib/live_model_impl.cc
+++ b/rlclientlib/live_model_impl.cc
@@ -192,7 +192,7 @@ namespace reinforcement_learning {
     RETURN_IF_FAIL(_model->request_multi_slot_decision(event_id, num_decisions, context_json, actions_ids, actions_pdfs, model_version, status));
 
     RETURN_IF_FAIL(populate_slates_response(actions_ids, actions_pdfs, std::string(event_id), std::string(model_version), resp, _trace_logger.get(), status));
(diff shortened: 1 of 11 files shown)


rlclientlib/live_model_impl.cc
```diff
@@ -170,6 +170,10 @@ namespace reinforcement_learning {
       return error_code::not_supported;
     }
 
+    if(_protocol_version == 1 && _model->model_type() != model_management::model_type_t::SLATES) {
```
olgavrou, line 173: is this a check that can be done by the logger facade on initialization so that we don't have to leak the protocol version to the live model impl?

rlclientlib/logger/logger_facade.cc
```diff
@@ -92,19 +77,39 @@ namespace reinforcement_learning {
...
+      }
+      return error_code::success;
+    }
+
+    int multi_slot_logger_facade::log_decision(model_type model_type, const std::string& event_id, const char* context, unsigned int flags, const std::vector<std::vector<uint32_t>>& action_ids,
       const std::vector<std::vector<float>>& pdfs, const std::string& model_version, api_status* status) {
       switch (_version) {
-        case 1: return _v1->log_decision(event_id, context, flags, action_ids, pdfs, model_version, status);
-        case 2: return _v2->log(event_id.c_str(), _serializer.event(context, flags, action_ids, pdfs, model_version), _serializer.type, status);
+        case 1: {
```
olgavrou, line 103: same here, can we pass the model type during the logger object `init()` and make this check there once? from what I understand it's not like we can continue execution if we are using the wrong protocol version with the wrong logger, right?

ataymano (commented):
Am I understand correctly about API changes?:
1) CB (no changes) - choose_rank
2) CCB v1 - request_decisions (TODO: deprecate)
3) Slates v1/v2 + CCB v2 - request_slates_decision (TODO: rename)

rlclientlib/logger/logger_facade.cc
ataymano: agree, it seems better to make model_type as part of logger_facade state

rlclientlib/live_model_impl.cc
kumpera: It's done there as well, but the error message won't be as good as the one we can get here.

rlclientlib/logger/logger_facade.cc
kumpera: We can't check during initialization because it's not conditional. multi_slot_logger_facade is always initialized, regardless of the underlying model.

We can move the model type probing to construction, but that doesn't really buy us much.

rlclientlib/live_model_impl.h
```diff
@@ -83,6 +83,7 @@ namespace reinforcement_learning
     model_management::data_callback_fn _data_cb;
     utility::watchdog _watchdog;
     learning_mode _learning_mode;
+    int _protocol_version;
```
ataymano, line 86: [nit] const

rlclientlib/live_model_impl.cc
kumpera: We get an invalid_argument with empty message.

kumpera:
> Do I understand correctly about API changes?:
> 
>     1. CB (v1/v2) - choose_rank (no changes)
> 
>     2. CCB v1 - request_decisions (no changes) (TODO: deprecate)
> 
>     3. Slates v1/v2 + CCB v2 - request_slates_decision (TODO: rename)

Yes, that's the plan. The ccb APIs are already marked as deprecated, so the big item is remove them once there are no resources left that could possibly use it.

For 3. I have a branch that depends on this one waiting to push PR'd.

rlclientlib/live_model_impl.h
kumpera: done

rlclientlib/logger/logger_facade.cc
ataymano: That's actually not great. Every logger initialization contributes to amount of background threads. Now we have model_type explosed, so we can fix it: combination of model_type + version clearly identify the only one valid rank call, so we can instantiate only one required underlying logger during init call (and throw error for incompatible combination).
It seems like other step of refactoring though, so we can postpone it to next PR (happy to do it).

ataymano (approved):

test_tools/example_gen/example_gen.cc
```diff
@@ -0,0 +1,181 @@
...
+          std::cout << status.get_error_msg() << std::endl;
+      break;
+    };
+    case F_REWARD: // "float"
+      if( rl.report_outcome("event_id", 1.5, &status) != err::success ) 
+          std::cout << status.get_error_msg() << std::endl;
+      break;
+
+    case F_I_REWARD: // "float-int",
+      if( rl.report_outcome("event_id", 1.5, 10, &status) != err::success )
```
olgavrou, line 122: Is this supposed to go to the `report_outcome(const char* primary_id, int secondary_id, float outcome, api_status* status)` method? Are the arguments supposed to be reversed here, i.e. `"event_id", 10, 1.5`?

test_tools/example_gen/example_gen.cc
```diff
@@ -0,0 +1,181 @@
...
+    case F_REWARD: // "float"
+      if( rl.report_outcome("event_id", 1.5, &status) != err::success ) 
+          std::cout << status.get_error_msg() << std::endl;
+      break;
+
+    case F_I_REWARD: // "float-int",
+      if( rl.report_outcome("event_id", 1.5, 10, &status) != err::success ) 
+          std::cout << status.get_error_msg() << std::endl;
+      break;
+    case F_S_REWARD: // "float-string"
```
olgavrou, line 125: same here? I can't find a `report_outcome` with signature `(const char*, float, const char*)`

kumpera: I have a follow up PR fixing the 2 issues you found. they were submerged in thousands of linker warnings.

PR #193, opened 2020-09-28
Add CMake option to include azure factories components at compile time.

SchuylerGoodman:
In order to compile reinforcement_learning without the cpprestsdk dependency, this PR adds a precompiler definition, `USE_AZURE_FACTORIES`. This is used in three places:
1. [include/constants.h](https://github.com/VowpalWabbit/reinforcement_learning/blob/master/include/constants.h) to change default behavior when azure factories are not linked from
  * OBSERVATION_EH_SENDER -> OBSERVATION_FILE_SENDER
  * INTERACTION_EH_SENDER -> INTERACTION_FILE_SENDER
  * AZURE_STORAGE_BLOB -> NO_MODEL_DATA
  * NULL_TIME_PROVIDER -> CLOCK_TIME_PROVIDER
2. [rlclientlib/factory_resolver.cc](https://github.com/VowpalWabbit/reinforcement_learning/blob/master/rlclientlib/factory_resolver.cc) to #ifdef out the call to register_azure_factories()
3. [unit_test/model_mgmt_test.cc](https://github.com/VowpalWabbit/reinforcement_learning/blob/master/unit_test/model_mgmt_test.cc) to exclude some tests that explicitly use cpprest-dependent components (like restapi_data_transport).

This was done for both Windows and Linux as follows:

**Windows**
[.scripts/init.cmd](https://github.com/VowpalWabbit/reinforcement_learning/blob/master/.scripts/init.cmd) now contains an additional environment variable definition for SkipAzureFactories. This is set to false by default (meaning azure factories are not skipped), but can be set to true to compile without cpprestsdk.
[rlclientlib/rlclientlib.vcxproj](https://github.com/VowpalWabbit/reinforcement_learning/blob/master/rlclientlib/rlclientlib.vcxproj) - Now defines `USE_AZURE_FACTORIES` when `SkipAzureFactories!='true'`
[unit_test/unit_test.vcxproj](https://github.com/VowpalWabbit/reinforcement_learning/blob/master/unit_test/unit_test.vcxproj) - Same as above

**Linux**
[CMakeLists.txt](https://github.com/VowpalWabbit/reinforcement_learning/blob/master/CMakeLists.txt) - the option `vw_USE_AZURE_FACTORIES` was added.

This is used in both [rlclientlib/CMakeLists.txt](https://github.com/VowpalWabbit/reinforcement_learning/blob/master/rlclientlib/CMakeLists.txt) and [unit_test/CMakeLists.txt](https://github.com/VowpalWabbit/reinforcement_learning/blob/master/unit_test/CMakeLists.txt) to optionally include cpprestsdk-dependent factory classes (like azure_factories, eventhub_client, restapi_data_transport, http_client, etc.). It also adds the `USE_AZURE_FACTORIES` precompiler definition.


kumpera (approved):
LGTM

kumpera:
The windows compilation is now failing, it's missing the new define.

jackgerrits (approved):

SchuylerGoodman:
Still trying to verify that I didn't break stdin2rllib

test_tools/stdin2rllib/main.cc
```diff
@@ -215,11 +223,11 @@ void parse_and_send_cb_event(json::value obj, r::live_model& rl, r::api_status&
...
 
 	//send outcome if cost exists and is non-zero
-	auto cost = U("_label_cost");
-	bool has_cost = obj.has_field(cost);
+	auto cost = "_label_cost";
+	bool has_cost = obj.HasMember(cost);
 	if (!has_cost) return;
 
-	float reward = -(obj[cost].as_double());
+	float reward = -(obj[cost].GetFloat());
```
olgavrou, line 230: GetDouble?

olgavrou (approved):

test_tools/stdin2rllib/main.cc
SchuylerGoodman: Good catch, I must have tab-completed incorrectly

test_tools/stdin2rllib/main.cc
```diff
@@ -185,26 +190,29 @@ void parse_and_send_ca_event(json::value obj, r::live_model& rl, r::api_status&
...
-	auto evt_id = U("EventId");
-	bool has_event_id = obj.has_field(evt_id);
+	auto evt_id = "EventId";
+	bool has_event_id = obj.HasMember(evt_id);
 	if (!has_event_id) {
 		std::cout << "missing 'EventId' field" << std::endl;
 		return;
 	}
-	std::string event_id = to_utf8string(obj.at(evt_id).as_string());
+	std::string event_id = obj[evt_id].GetString();
```
zwd-ms, line 201: GetString() returns a pointer. Do we need GetStringLength() for Unicode?
https://rapidjson.org/md_doc_tutorial.html#QueryString

Can we avoid the string copy, and pass the pointer to choose_rank() below?

test_tools/stdin2rllib/main.cc
```diff
@@ -185,26 +190,29 @@ void parse_and_send_ca_event(json::value obj, r::live_model& rl, r::api_status&
...
 		return;
 	}
 	//extract context string
-	json::value v = obj.at(context);
-	std::string c = to_utf8string(v.serialize());
+	auto& context_obj = obj[context];
+	rj::StringBuffer sb;
+	rj::Writer<rj::StringBuffer> writer(sb);
+	context_obj.Accept(writer);
+	std::string c = sb.GetString();
```
zwd-ms, line 215: Similarly, sb.GetString() returns a pointer, need sb.GetSize()? Perhaps just pass the pointer to avoid copying.

zwd-ms (commented):
thanks for making the changes.

test_tools/stdin2rllib/main.cc
SchuylerGoodman: I think I fixed this, could you take a look? Thanks!

ZhangWangda: LGTM, thanks

include/constants.h
```diff
@@ -74,5 +74,16 @@ namespace reinforcement_learning {  namespace value {
       const bool DEFAULT_MODEL_BACKGROUND_REFRESH = true;
       const int DEFAULT_VW_POOL_INIT_SIZE = 4;
       const int DEFAULT_PROTOCOL_VERSION = 1;
+#ifdef USE_AZURE_FACTORIES
```
kumpera, line 77: Having a define in a public header that doesn't depend on the compiler/system is a bad practice and we should avoid if possible.

They should be functions and the ifdef should exist within their implementation.

I don't understand the change to the default clock.

kumpera (changes requested):

rajban2017 (approved):

include/constants.h
SchuylerGoodman: Thanks for your feedback, @kumpera! That being the case, should we define static functions and implement them in a new constants.cc?

As far as the change to the default clock, it was my understanding that the file based logger requires a time provider implementation. Since `NULL_TIME_PROVIDER` was the current default, that would cause an issue. @lokitoth is that right?

kumpera (approved):
LGTM, let's get this merged

Issue #142, opened 2020-04-22
[improvement] Remove left overs from CCB


kumpera:
As CCB designed evolved, some bits turned out be to redundant and we slipped from removing them.
They should be removed from the code base.

In no particular order:
- We don't need a separate decision sender
https://github.com/VowpalWabbit/reinforcement_learning/blob/master/rlclientlib/live_model_impl.cc#L281

- We don't need to register a new sender type (it's identical to existing sender types):
https://github.com/VowpalWabbit/reinforcement_learning/blob/master/rlclientlib/azure_factories.cc#L22

prateekiiest:
Hey @kumpera. This seems a good first issue. Is anyone working on this at the moment?

PR #198, opened 2020-10-01
Single interaction logger facade

ataymano:


kumpera (approved):
LGTM.

One thing we could do is pass model_type_t to init instead of the ctor. It would allow us to skip the conversion call in the logger path.

rlclientlib/live_model_impl.h
```diff
@@ -96,11 +96,10 @@ namespace reinforcement_learning
 
     std::unique_ptr<model_management::i_data_transport> _transport{nullptr};
     std::unique_ptr<model_management::i_model> _model{nullptr};
-    std::unique_ptr<logger::cb_logger_facade> _ranking_logger{nullptr};
-    std::unique_ptr<logger::ca_logger_facade> _continuous_action_logger{nullptr};
+
+    std::unique_ptr<logger::interaction_logger_facade> _ranking_logger{nullptr};
```
olgavrou, line 100: I am wondering if this should be named more broadly `_interaction_logger` since its type is of `interaction_logger_facade` and ranking doesn't really make sense in the continuous actions context

olgavrou (approved):

olgavrou (commented):
Really like this change

rlclientlib/live_model_impl.h
ataymano: agree, let me fix it

ataymano:
@kumpera let me think about moving model_type_t to init in next PR - maybe it will make ctor uglier. Want to think it through, but do not want to block merge.

ataymano:
> Really like this change

thank you

ataymano:
#142

PR #200, opened 2020-10-02
Continuous action csharp simulation

olgavrou:
- `StatisticsCalculator` to use generics so that it can take continuous actions too
- added robot joint simulation example in `RLSimulator`
- `enum LoopKind` added in `RLDriver`
- added `ca` command line option and refactor `CommandBase` to take `--slates`, `--ca` and return the loop kind

@kumpera I re-renamed the `ChosenActionDensity` to `ChosenActionPdfValue` since (afaik) both the terms `pdf` and `density` can both be used in a continuous actions context to refer to the probability density function. We want it to be clear that we are returning the value of the pdf (density) at the chosen action location


bindings/cs/rl.net.native/rl.net.continuous_action_response.cc
```diff
@@ -25,7 +25,7 @@ API float GetContinuousActionChosenAction(reinforcement_learning::continuous_act
     return response->get_chosen_action();
 }
 
-API float GetContinuousActionChosenActionDensity(reinforcement_learning::continuous_action_response* response)
+API float GetContinuousActionChosenActionPdfValue(reinforcement_learning::continuous_action_response* response)
```
kumpera, line 28: This looks like reverting the rename you had in the previous PR, right?

kumpera (approved):

bindings/cs/rl.net.native/rl.net.continuous_action_response.cc
olgavrou: yes (see PR description):

I re-renamed the ChosenActionDensity to ChosenActionPdfValue since (afaik) both the terms pdf and density can both be used in a continuous actions context to refer to the probability density function. We want it to be clear that we are returning the value of the pdf (density) at the chosen action location

PR #202, opened 2020-10-06
Introduce get_context_info that collects a bunch of information from the json payload.

kumpera:
The new function uses the SAX parser which is significantly faster and avoids doing multiple passes
over the payload.

One thing we must discuss as part of this PR is that I'd like to replace all usage of `const char *context` in the logging stack
with an instance of this object as it will be needed for dedup to compute the contents of each action.

### DIFF
--- a/rlclientlib/live_model_impl.cc
+++ b/rlclientlib/live_model_impl.cc
@@ -208,11 +208,15 @@ namespace reinforcement_learning {
     RETURN_IF_FAIL(check_null_or_empty(event_id, _trace_logger.get(), status));
     RETURN_IF_FAIL(check_null_or_empty(context_json, _trace_logger.get(), status));
 
-    // Ensure multi comes before slots, this is a current limitation of the parser.
-    RETURN_IF_FAIL(utility::validate_multi_before_slots(context_json, _trace_logger.get(), status));
+    utility::ContextInfo context_info;
+    RETURN_IF_FAIL(utility::get_context_info(context_json, context_info, _trace_logger.get(), status));
 
-    size_t num_decisions;
-    RETURN_IF_FAIL(utility::get_slot_count(num_decisions, context_json, _trace_logger.get(), status));
+    // Ensure multi comes before slots, this is a current limitation of the parser.
+    if(context_info.slots.size() < 1 || context_info.actions.size() < 1 || context_info.slots[0].first < context_info.actions[0].first) {
+      RETURN_ERROR_LS(_trace_logger.get(), status, json_parse_error) << "There must be both a _multi field and _slots, and _multi must come first.";
+    }
+
+    size_t num_decisions = context_info.slots.size();
 
     std::vector<std::vector<uint32_t>> actions_ids;
     std::vector<std::vector<float>> actions_pdfs;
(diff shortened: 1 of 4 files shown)


rlclientlib/live_model_impl.cc
```diff
@@ -208,11 +208,15 @@ namespace reinforcement_learning {
     RETURN_IF_FAIL(check_null_or_empty(event_id, _trace_logger.get(), status));
     RETURN_IF_FAIL(check_null_or_empty(context_json, _trace_logger.get(), status));
 
+    utility::ContextInfo context_info;
+    RETURN_IF_FAIL(utility::get_context_info(context_json, context_info, _trace_logger.get(), status));
+
     // Ensure multi comes before slots, this is a current limitation of the parser.
-    RETURN_IF_FAIL(utility::validate_multi_before_slots(context_json, _trace_logger.get(), status));
```
ataymano, line 212: are we still using these functions? do we want to clean their implementation up as well?

rlclientlib/live_model_impl.cc
```diff
@@ -208,11 +208,15 @@ namespace reinforcement_learning {
...
+    if(context_info.slots.size() < 1 || context_info.actions.size() < 1 || context_info.slots[0].first < context_info.actions[0].first) {
+      RETURN_ERROR_LS(_trace_logger.get(), status, json_parse_error) << "There must be both a _multi field and _slots, and _multi must come first.";
+    }
 
-    size_t num_decisions;
-    RETURN_IF_FAIL(utility::get_slot_count(num_decisions, context_json, _trace_logger.get(), status));
```
ataymano, line 215: same question as about validate_multi_before_slots

kumpera: All those functions would go away as all we need from them is provided by ContextInfo.
I'd probably enrich ContextInfo to include relevant helpers such as `is_multi_before_slots`.
One thing I need to add is id extraction for slots/actions for it to completely subsumes all the helpers we have.

kumpera:
Do we want ContextInfo to sit where it is or move it to a different place in the codebase?

rlclientlib/utility/context_helper.cc
```diff
@@ -61,120 +62,83 @@ namespace reinforcement_learning { namespace utility {
...
-   *                  case of failure
-   * \return  error_code::success if there are no errors.  If there are errors then the error code is
-   *          returned.
-   */
-  int validate_multi_before_slots(const char *context, i_trace* trace, api_status* status)
+  int get_context_info(const char *context, ContextInfo &info, i_trace* trace, api_status* status)
   {
-    auto slots_pos = strstr(context, "_slots");
-    auto multi_pos = strstr(context, "_multi");
+    std::string copy(context);
```
ataymano, line 130: is it required to make copy here?

kumpera: Yes, I'm using kParseInsituFlag which modified the string.
I did some testing and it's measurably faster even with the copy.

ataymano (approved):

ataymano:
> Do we want ContextInfo to sit where it is or move it to a different place in the codebase?

Current place is fine for me. We can always move it later.

PR #207, opened 2020-10-13
Add dedup_dict class that will later be used by the dedup work.

kumpera:
This type will be used during serialization at two spots.

Before an event is serialized into a generic_event we process the json payload and include the
resulting action_list into the generated generic_event.

During batch creation we use get_action and remove_action to compute the per batch dictionary
and unregister actions as events sink into the batch.

Next is an adaptative per-batch dictionary builder and schema change.

### DIFF
--- a/rlclientlib/CMakeLists.txt
+++ b/rlclientlib/CMakeLists.txt
@@ -25,6 +25,7 @@ set(PROJECT_SOURCES
   console_tracer.cc
   continuous_action_response.cc
   decision_response.cc
+  dedup.cc
   error_callback_fn.cc
   factory_resolver.cc
   live_model_impl.cc
@@ -98,6 +99,7 @@ set(PROJECT_PUBLIC_HEADERS
 set(PROJECT_PRIVATE_HEADERS
   azure_factories.h
   console_tracer.h
+  dedup.h
   live_model_impl.h
   logger/async_batcher.h
   logger/event_logger.h
--- a/rlclientlib/CMakeLists.txt
+++ b/rlclientlib/CMakeLists.txt
@@ -113,6 +113,7 @@ set(PROJECT_PRIVATE_HEADERS
   moving_queue.h
   generic_event.h
   ranking_event.h
+  rl_string_view.h
   sampling.h
   serialization/fb_serializer.h
   serialization/json_serializer.h
--- a/rlclientlib/dedup.cc
+++ b/rlclientlib/dedup.cc
@@ -19,13 +19,13 @@ namespace reinforcement_learning
     return uniform_hash(start, size, 0);
   }
 
-  dedup_dict::action_id_t dedup_dict::add_action(const char *start, size_t size)
+  dedup_dict::action_id_t dedup_dict::add_action(const char *start, size_t length)
   {
-    action_id_t hash = hash_content(start, size);
+    action_id_t hash = hash_content(start, length);
     auto it = _entries.find(hash);
     if (it == _entries.end())
     {
-      _entries.insert({ hash, dict_entry(start, size) });
+      _entries.insert({ hash, dict_entry(start, length) });
     }
     else
     {
@@ -34,9 +34,9 @@ namespace reinforcement_learning
     return hash;
   }
 
-  bool dedup_dict::remove_action(action_id_t hash)
+  bool dedup_dict::remove_action(action_id_t aid)
   {
-    auto it = _entries.find(hash);
+    auto it = _entries.find(aid);
     if (it == _entries.end())
       return false;
 
@@ -45,6 +45,14 @@ namespace reinforcement_learning
       _entries.erase(it);
 
     return true;
+  }
+
+  string_view dedup_dict::get_action(action_id_t aid)
+  {
+    auto it = _entries.find(aid);
+    if (it == _entries.end())
+      return string_view();
+    return string_view(it->second._content.data(), it->second._length);
   }
 
   int dedup_dict::transform_payload(const char* payload, std::string& edited_payload, action_list_t& action_ids, api_status* status)
--- a/rlclientlib/dedup.cc
+++ b/rlclientlib/dedup.cc
@@ -47,7 +47,7 @@ namespace reinforcement_learning
     return true;
   }
 
-  string_view dedup_dict::get_action(action_id_t aid)
+  string_view dedup_dict::get_action(action_id_t aid) const
   {
     auto it = _entries.find(aid);
     if (it == _entries.end())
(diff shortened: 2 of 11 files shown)


kumpera:
TODO
update the dedup docs exampling the inner workings of this.

rlclientlib/dedup.h
```diff
@@ -0,0 +1,42 @@
...
+    dedup_dict(dedup_dict&&) = default;
+    dedup_dict& operator=(dedup_dict&&) = default;
+    ~dedup_dict() = default;
+
+    //! Returns true if the action was found. This doesn't tell the ref count status of that action
+    bool remove_action(action_id_t aid);
+    //! Returns the action id of the action described by [start, start+size[
+    action_id_t add_action(const char* start, size_t length);
+    //! Return a string_view of the action content, or an empty view if not found
+    string_view get_action(action_id_t aid);
```
ataymano, line 28: const;

rlclientlib/dedup.h
```diff
@@ -0,0 +1,42 @@
...
+
+    int transform_payload(const char* payload, std::string& edited_payload, action_list_t& action_ids, api_status* status);
```
ataymano, line 30: const?

kumpera: It's not const as it extract the actions from the payload and include on the dictionary. I guess the name is misleading.

kumpera:
I'm going to update the docs once the schema change and other bits land.

rlclientlib/CMakeLists.txt
```diff
@@ -25,6 +25,7 @@ set(PROJECT_SOURCES
   console_tracer.cc
   continuous_action_response.cc
   decision_response.cc
+  dedup.cc
```
olgavrou, line 28: shouldn't these be added to the `vcxproj` file too?

rlclientlib/rl_string_view.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
```
olgavrou, line 1: This is also available in the vw submodule: `vowpalwabbit/vw_string_view.h` would it make sense to re-use it from there

olgavrou: Would it make sense to place `vw_string_view.h` into the `common` dir/lib @jackgerrits ? (see common lib [PR](https://github.com/VowpalWabbit/vowpal_wabbit/pull/2573)

rlclientlib/dedup.h
```diff
@@ -0,0 +1,42 @@
...
+#include "rl_string_view.h"
+
+#include <vector>
+#include <unordered_map>
+
+
+namespace reinforcement_learning {
+class dedup_dict {
+public:
+    typedef uint64_t action_id_t;
```
olgavrou, line 12: `using` instead of `typedef?`

rlclientlib/dedup.h
```diff
@@ -0,0 +1,42 @@
...
+    dedup_dict(const dedup_dict&) = delete;
+    dedup_dict& operator=(const dedup_dict&) = delete;
+
+    dedup_dict(dedup_dict&&) = default;
+    dedup_dict& operator=(dedup_dict&&) = default;
+    ~dedup_dict() = default;
+
+    //! Returns true if the action was found. This doesn't tell the ref count status of that action
+    bool remove_action(action_id_t aid);
+    //! Returns the action id of the action described by [start, start+size[
```
olgavrou, line 25: nit: 
```suggestion
    //! Returns the action id of the action described by [start, start+size]
```

rlclientlib/dedup.cc
```diff
@@ -0,0 +1,83 @@
...
+  {
+    auto it = _entries.find(aid);
+    if (it == _entries.end())
+      return string_view();
+    return string_view(it->second._content.data(), it->second._length);
+  }
+
+  int dedup_dict::transform_payload_and_add_actions(const char* payload, std::string& edited_payload, action_list_t& action_ids, api_status* status)
+  {
+    utility::ContextInfo context_info;
```
olgavrou, line 60: nit:
```suggestion
    u::ContextInfo context_info;
```

rlclientlib/dedup.cc
```diff
@@ -0,0 +1,83 @@
...
+    RETURN_IF_FAIL(utility::get_context_info(payload, context_info, nullptr, status));
```
olgavrou, line 61: nit:
```suggestion
    RETURN_IF_FAIL(u::get_context_info(payload, context_info, nullptr, status));
```

olgavrou:
This is more of a question so that I can understand the dedup work: 

we are hashing the actions in their entirety (i.e. feature name _with_ the feature value), do we know if actions generally share feature names and feature values, or if it will create too many unique action ids?

rlclientlib/rl_string_view.h
kumpera: We must not include non-public headers out of a submodule. If we're to reuse, it should be from a place like `dir/lib`.

rlclientlib/CMakeLists.txt
kumpera: good point

rlclientlib/dedup.h
kumpera: yeah, probably

rlclientlib/dedup.h
kumpera: done

kumpera:
> This is more of a question so that I can understand the dedup work:
> 
> we are hashing the actions in their entirety (i.e. feature name _with_ the feature value), do we know if actions generally share feature names and feature values, or if it will create too many unique action ids?

The core assumption of the dedup work is some core workloads have a small set of actions and that leads to significant redundancy elimination.

The worst case, where all actions are unique we get a small size expansion due to the extra book keeping. If that turns out to be an issue, we'll address it by decoupling dedup from compression and apply them selectively based on their effectiveness on previous batches.

olgavrou (approved):

rlclientlib/rl_string_view.h
jackgerrits: > We must not include non-public headers out of a submodule.

Whats the reason behind this?

kumpera: Because it's a tight dependencies between the two repositories in an invisible way.

Since we don't CI or review for external usage of internal headers, the chance of us introducing breakage when we update the vw submodule increases. Introducing friction in the submodule update process is a massive hurdle to our agility.

Once we have something like that _commons_ set of headers in VW that @jackgerrits proposed, we can remove this file.

Issue #172, opened 2020-08-24
Get rid of copy paste in loggers initialization


ataymano:
Right now we have several sets of parameters required for logger initialization in a single plain list:
"interaction.eventhub.host",
"interaction.eventhub.name",
"interaction.eventhub.keyname",
...
"observation.eventhub.host",
"observation.eventhub.name",
"observation.eventhub.keyname",
...
It introduces some amount of copy-paste for all logger implementations.
It would be helpful either to move to more structured config:
"interaction_config": {"eventhub" { "host", "name"}},...
or at least simulate this behavior by accessing to certain keys of config by prefix

kumpera:
This issue was fixed by #210.

PR #210, opened 2020-10-20
Introduce i_async_batcher and i_payload_transformer interfaces.

kumpera:
The plan is to use i_async_batcher to abstract over the collection serializer used.
We'll use that to have a serializer that applies dedup.

i_payload_transformer will be used to apply compression to each event.

### DIFF
--- a/rlclientlib/logger/logger_facade.cc
+++ b/rlclientlib/logger/logger_facade.cc
@@ -11,24 +11,35 @@ namespace reinforcement_learning {
       RETURN_ERROR_ARG(nullptr, status, protocol_not_supported, "Current protocol version is not supported");
     }
 
+
+    template<typename T>
+    i_async_batcher<T>* create_async_batcher(const utility::configuration& c, i_message_sender* sender, utility::watchdog& watchdog, error_callback_fn* perror_cb, const char *section) {
+      auto config = utility::get_batcher_config(c, section);
+      return new async_batcher<T, fb_collection_serializer>(
+        sender,
+        watchdog,
+        perror_cb,
+        config
+      );
+    }
+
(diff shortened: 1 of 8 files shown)


rlclientlib/logger/logger_facade.cc
```diff
@@ -11,28 +11,37 @@ namespace reinforcement_learning {
       RETURN_ERROR_ARG(nullptr, status, protocol_not_supported, "Current protocol version is not supported");
     }
 
+    template<typename T>
+    i_async_batcher<T>* create_async_batcher(const utility::configuration& c, i_message_sender* sender, utility::watchdog& watchdog, error_callback_fn* perror_cb) {
+      return new async_batcher<T, fb_collection_serializer>(
+        sender,
+        watchdog,
+        perror_cb,
+        c.get_int(name::INTERACTION_SEND_HIGH_WATER_MARK, 198 * 1024),
```
ataymano, line 20: this factory method does not seem INTERACTION specific.
Do we want to fix https://github.com/VowpalWabbit/reinforcement_learning/issues/172 before?

kumpera: uggg, yeah, this won't work as is. I'll take a run at that issue.

kumpera:
@ataymano this PR cleans up the config bits properly: https://github.com/VowpalWabbit/reinforcement_learning/pull/211
I'm going to rebase this PR on top of #211

rlclientlib/logger/logger_facade.cc
kumpera: PR rebased on top of master with the batcher config cleanup.

kumpera:
Hmm, I'm probably not going to use this as there might be better abstractions.

PR #211, opened 2020-10-23
Introduce config_helper.h to sanitized async_batch configuration.

kumpera:
This commit introduce the following changes:

Introduce `get_batcher_config` that can load all async_batcher related configuration
at once into an instance of async_batcher_config.

All related properties now have a global and specific configuration and if you don't set
the specific one, the global one is used instead.

Introduce interaction.queue.mode, observation.queue.mode and observation.content.encoding to
make them all symetric.


rlclientlib/logger/async_batcher.h
```diff
@@ -164,19 +152,15 @@ namespace reinforcement_learning { namespace logger {
     i_message_sender* sender,
     utility::watchdog& watchdog,
     error_callback_fn* perror_cb,
-    const size_t send_high_water_mark,
-    const size_t batch_timeout_ms,
-    const size_t queue_max_capacity,
-    queue_mode_enum queue_mode,
-    content_encoding_enum content_encoding)
+    utility::async_batcher_config& config)
```
ataymano, line 155: const?

ataymano (approved):

rlclientlib/logger/async_batcher.h
kumpera: good catch :)

PR #217, opened 2020-10-27
Schema for Action Dictionaries and bunch of other support code.

kumpera:
In no particular order:

Add action_ids to generic_event.
Add shared_state for collections serializers.
Introduce i_async_batcher interface to abstract over the collection serializer being used.

### DIFF
--- a/rlclientlib/dedup.h
+++ b/rlclientlib/dedup.h
@@ -1,5 +1,6 @@
 #pragma once
 #include "api_status.h"
+#include "generic_event.h"
 #include "rl_string_view.h"
 
 #include <vector>
@@ -9,8 +10,6 @@
 namespace reinforcement_learning {
 class dedup_dict {
 public:
-    using action_id_t = uint64_t;
-    using action_list_t = std::vector<action_id_t>;
     dedup_dict() = default;
 
     dedup_dict(const dedup_dict&) = delete;
@@ -21,13 +20,13 @@ public:
     ~dedup_dict() = default;
 
     //! Returns true if the action was found. This doesn't tell the ref count status of that action
-    bool remove_action(action_id_t aid);
+    bool remove_action(generic_event::action_id_t aid);
     //! Returns the action id of the action described by [start, start+length[
-    action_id_t add_action(const char* start, size_t length);
+    generic_event::action_id_t add_action(const char* start, size_t length);
     //! Return a string_view of the action content, or an empty view if not found
-    string_view get_action(action_id_t aid) const;
+    string_view get_action(generic_event::action_id_t aid) const;
 
-    int transform_payload_and_add_actions(const char* payload, std::string& edited_payload, action_list_t& action_ids, api_status* status);
+    int transform_payload_and_add_actions(const char* payload, std::string& edited_payload, generic_event::action_list_t& action_ids, api_status* status);
 private:
     struct dict_entry {
         size_t _count;
@@ -36,7 +35,7 @@ private:
 
         dict_entry(const char* data, size_t length);
     };
-    std::unordered_map<action_id_t, dict_entry> _entries;
+    std::unordered_map<generic_event::action_id_t, dict_entry> _entries;
 };
 
 }
--- a/rlclientlib/dedup.h
+++ b/rlclientlib/dedup.h
@@ -20,11 +20,11 @@ public:
     ~dedup_dict() = default;
 
     //! Returns true if the object was found. This doesn't tell the ref count status of that object
-    bool remove_object(generic_event::object_id_t aid);
+    bool remove_object(generic_event::object_id_t oid);
     //! Returns the object id of the object described by [start, start+length[
     generic_event::object_id_t add_object(const char* start, size_t length);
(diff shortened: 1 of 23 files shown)


rlclientlib/generic_event.h
```diff
@@ -11,7 +11,11 @@ namespace reinforcement_learning {
     using payload_type_t = messages::flatbuff::v2::PayloadType;
 
   public:
+    using action_id_t = uint64_t;
```
ataymano, line 14: [nit] The word "action" make generic event less generic (i.e. outcome_event has no actions). At the same time we actually do not really care if they are actions - just some re-usable sub-objects (it also easily can be extended to slots and maybe to some parts of shared context). Given that do we want to use more generic word here instead of action (like "object" or something). I understand that we maybe far from implementing this now, so it is purely naming question, to have some room to be able to move this direction in future.

rlclientlib/logger/async_batcher.h
```diff
@@ -52,6 +65,7 @@ namespace reinforcement_learning { namespace logger {
     event_queue<TEvent> _queue;       // A queue to accumulate batch of events.
     size_t _send_high_water_mark;
     error_callback_fn* _perror_cb;
+    typename TSerializer<TEvent>::shared_state_t& _shared_state;
```
ataymano, line 68: who is going to own this guy? Looks like with current dummy version we end up having reference to non-existing stack variable - we are not using it, but it seems to be UB. If async_batcher is going to be owner, do we need it as input argument to ctr? Or am I missing something?

rlclientlib/schema/v2/ActionDictionary.fbs
```diff
@@ -0,0 +1,8 @@
+namespace reinforcement_learning.messages.flatbuff.v2;
+
+table ActionDictionary {
```
ataymano, line 3: similar to comment about "action". Do we want to have some less specific structure here - i.e. DedupInfo or something like this (which may have only action_dictionary for now)? It will help to have generic_event interface look more generic + give us more space for further development.

rlclientlib/generic_event.h
kumpera: Good point, I'll pick a more neutral name.

rlclientlib/logger/async_batcher.h
kumpera: In the full PR this is owned by live_model_impl and passed down the pipe.
I've chosen to make it unconditional because the template magic required to make this optional turned out to be way too complicated to be worth it. This UB goes away later when I introduce the class that holds the state.

rlclientlib/schema/v2/ActionDictionary.fbs
kumpera: I see your point about naming it in a more generic way.
One thight though is that I rather keep this struct dedup specific for now.

rlclientlib/generic_event.h
kumpera: it's object_id now.

rlclientlib/schema/v2/ActionDictionary.fbs
kumpera: Renamed.

rlclientlib/dedup.h
```diff
@@ -20,14 +19,14 @@ class dedup_dict {
...
     ~dedup_dict() = default;
 
-    //! Returns true if the action was found. This doesn't tell the ref count status of that action
-    bool remove_action(action_id_t aid);
-    //! Returns the action id of the action described by [start, start+length[
-    action_id_t add_action(const char* start, size_t length);
-    //! Return a string_view of the action content, or an empty view if not found
-    string_view get_action(action_id_t aid) const;
+    //! Returns true if the object was found. This doesn't tell the ref count status of that object
+    bool remove_object(generic_event::object_id_t aid);
```
olgavrou, line 23: nit: `aid` to be renamed to `oid`

rlclientlib/generic_event.cc
```diff
@@ -5,6 +5,14 @@
 
 using namespace std;
 namespace reinforcement_learning {
+  generic_event::generic_event(const char* _id, const timestamp& ts, payload_type_t type, flatbuffers::DetachedBuffer&& payload, object_list_t &&objects, float pass_prob)
+    : _id(_id)
```
olgavrou, line 9: nit
```suggestion
    : _id(id)
```

rlclientlib/logger/async_batcher.h
```diff
@@ -19,18 +19,32 @@ namespace reinforcement_learning {
 };
 
 namespace reinforcement_learning { namespace logger {
+  template<typename TEvent>
+  class i_async_batcher {
+  public:
+    virtual ~i_async_batcher() {}
```
olgavrou, line 25:
```suggestion
    virtual ~i_async_batcher() = default;
```

ataymano (approved):

kumpera:
Addressed Olga's comments.

olgavrou (approved):

PR #218, opened 2020-10-28
Resolve valgrind reported errors

olgavrou:
Fixes valgrind reported errors from new valgrind pipeline: https://github.com/VowpalWabbit/reinforcement_learning/pull/219


olgavrou:
making a draft until I actually create the pipeline

unit_test/watchdog_test.cc
```diff
@@ -83,6 +91,12 @@ BOOST_AUTO_TEST_CASE(watchdog_report_with_error_handler) {
 }
 
 BOOST_AUTO_TEST_CASE(watchdog_multiple_threads) {
+  if (is_invoked_with("valgrind"))
```
kumpera, line 94: nit, what do you think of having a single macro for this such that it's easier to grep for them?
Something like RL_TEST_VALGRIND_DISABLE, or something.

kumpera (approved):

unit_test/watchdog_test.cc
olgavrou: don't know how that would work, this is a runtime argument

PR #220, opened 2020-10-28
Fixes for warnings C4018 and C4456

SchuylerGoodman:
C4018: `live_model_impl.cc` and `safe_vw` have some implicit casts from `size_t`. Explicit casting bypasses warning C4018
C4456: The variable `itr` in `context_helper.cc` is declared twice in a nested loop

### DIFF
--- a/rlclientlib/vw_model/safe_vw.cc
+++ b/rlclientlib/vw_model/safe_vw.cc
@@ -220,7 +220,7 @@ namespace reinforcement_learning {
 
     VW::read_line_json<false>(*_vw, examples, &line_vec[0], get_or_create_example_f, this);
     // In order to control the seed for the sampling of each slot the event id + app id is passed in as the seed using the example tag.
-    for(int i = 0; i < (int)slot_count; i++)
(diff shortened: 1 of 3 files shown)


kiprop12 (approved):
Looks good

rlclientlib/vw_model/safe_vw.cc
```diff
@@ -220,7 +220,7 @@ namespace reinforcement_learning {
 
     VW::read_line_json<false>(*_vw, examples, &line_vec[0], get_or_create_example_f, this);
     // In order to control the seed for the sampling of each slot the event id + app id is passed in as the seed using the example tag.
-    for(int i = 0; i < slot_count; i++)
+    for(int i = 0; i < (int)slot_count; i++)
```
olgavrou, line 223: should `i` be made `uint32_t` instead?

SchuylerGoodman: yes, good idea

olgavrou (approved):

PR #222, opened 2020-10-30
Introduce dedup+zstd suppport.

kumpera:
This is done by introducing i_logger_extensions which hides all dedup behavior and state from the rest of the codebase.
The main touchpoint is in logger_facade.cc, that uses it to produce transformed events.
The use of the is_enabled() call is to shortcut to a fast path when available.

Finally, this commit includes test_tools/log_parser/parser.py which can parse rlclientlib output when using the file
backend.

I can further slice this change by taking the i_logger_extensions out if desireable.


rlclientlib/dedup-internals.h
```diff
@@ -0,0 +1,108 @@
...
+    ewma(float initial = 1, float weigth = 0.5): _current(initial), _weigth(weigth) {}
+
+    void update(float new_value) {
+      _current = (1 - _weigth) * _current + (_weigth * new_value);
+    }
+
+    float value() const { return _current; }
+  private:
+    float _current;
+    const float _weigth;
```
ataymano, line 54: [nit] typo

rlclientlib/dedup-internals.h
```diff
@@ -0,0 +1,108 @@
...
+      std::vector<char> _content;
+
+      dict_entry(const char* data, size_t length);
+    };
+    std::unordered_map<generic_event::object_id_t, dict_entry> _entries;
+  };
+
+  class ewma {
+  public:
+    ewma(float initial = 1, float weigth = 0.5): _current(initial), _weigth(weigth) {}
```
ataymano, line 45: explicit

ataymano: oh nevermind

rlclientlib/dedup.cc
```diff
@@ -1,83 +1,315 @@
...
+size_t action_dict_builder::size() const
+{
+  return _size_estimate * _state.get_ewma_value();
+}
+
+int action_dict_builder::finalize(generic_event& evt, api_status* status)
+{
+  i_time_provider* _time_provider = nullptr;
+  l::dedup_info_serializer ser;
+  const auto now = _time_provider != nullptr ? _time_provider->gmt_now() : timestamp();
```
olgavrou, line 220: how can `_time_provider` not be `nullptr` if it is just set to `nullptr` two lines above?

olgavrou (commented):
Some comments:

- there is a mismatch on passing arguments between the PR and the existing code: `type &arg` or `type *arg` vs existing: `type& arg` or `type* arg`
- `dedup-internals.h` rename to `dedup_internals.h` (underscore instead of hyphen)
- there are some comments in the code starting with `XXX` and some of them seem to be resolved or TODOs, can they be removed or made TODOs?
- I like the logger extension logic :)

rlclientlib/dedup-internals.h
```diff
@@ -0,0 +1,108 @@
...
+
+    //FIXME pass a value-oblivious interator to the keys
+    int get_all_values(const std::unordered_map<generic_event::object_id_t, size_t>& src, generic_event::object_list_t& action_ids, std::vector<string_view>& action_values, api_status* status);
+    int remove_all_values(const std::unordered_map<generic_event::object_id_t, size_t>&input, api_status* status);
+
+    void update_ewma(float value);
+    int compress(generic_event::payload_buffer_t& input, api_status* status) const;
+    int transform_payload_and_add_objects(const char* payload, std::string& edited_payload, generic_event::object_list_t& object_ids, api_status* status);
+
+    //test helpers, don't use them directly
```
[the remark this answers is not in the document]
kumpera: Hmmm, good point. I wonder what the rest of the codebase does, as it this probably isn't the first instance of this.

rlclientlib/dedup-internals.h
```diff
@@ -0,0 +1,108 @@
...
+    inline dedup_dict& get_dict() { return _dict; }
+    inline ewma& get_ewma() { return _ewma; }
+  private:
+    ewma _ewma;
+    dedup_dict _dict;
+    zstd_compressor _compressor;
+    std::mutex _mutex;
+  };
+
+  static const char* DEDUP_DICT_EVENT_ID = "3defd95a-0122-4aac-9068-0b9ac30b66d8";
```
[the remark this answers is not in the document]
kumpera: So it's visible to the test suite.

unit_test/event_logger_test.cc
```diff
@@ -0,0 +1,37 @@
...
+#include "constants.h"
+#include "serialization/payload_serializer.h"
+
+namespace r = reinforcement_learning;
+namespace err = reinforcement_learning::error_code;
+namespace v = reinforcement_learning::value;
+namespace l = reinforcement_learning::logger;
+namespace fb = flatbuffers;
+
+struct my_async_batcher : public l::i_async_batcher<r::generic_event>\
```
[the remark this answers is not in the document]
kumpera: Ooops, old file! This was used to test an old design.

rlclientlib/dedup-internals.h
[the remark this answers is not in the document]
olgavrou: ah ok, can it be named `DEDUP_DICT_EVENT_ID_TEST` then?

rlclientlib/dedup.cc
kumpera: This is a TODO that slipped over, good catch.

rlclientlib/dedup-internals.h
```diff
@@ -0,0 +1,108 @@
...
+  };
+
+  class dedup_state {
+  public:
+    explicit dedup_state(const utility::configuration &c);
+
+    string_view get_object(generic_event::object_id_t aid);
+    float get_ewma_value() const;
+
+    //FIXME pass a value-oblivious interator to the keys
```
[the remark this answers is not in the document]
kumpera: I'm unhappy with passing the map. Changed to templated methods and iterators.

kumpera:
> Some comments:
> 
>     * there is a mismatch on passing arguments between the PR and the existing code: `type &arg` or `type *arg` vs existing: `type& arg` or `type* arg`

Fixed.
> 
>     * `dedup-internals.h` rename to `dedup_internals.h` (underscore instead of hyphen)

Fixed.

> 
>     * there are some comments in the code starting with `XXX` and some of them seem to be resolved or TODOs, can they be removed or made TODOs?
> 
Fixed

rlclientlib/dedup.cc
```diff
@@ -1,83 +1,288 @@
...
+                                                                      _content(data, data + length)
+{
+}
+
+static generic_event::object_id_t hash_content(const char*start, size_t size)
+{
+  return uniform_hash(start, size, 0);
+}
+
+generic_event::object_id_t dedup_dict::add_object(const char*start, size_t length)
```
ataymano, line 27: looks like it can be called concurrently? do we need to introduce guard here?

rlclientlib/dedup.cc
```diff
@@ -1,83 +1,288 @@
...
+    RETURN_IF_FAIL(_builder.add(evt.get_object_list(), status));
+    return _ser.add(evt, status);
+  }
+
+  uint64_t size() const {
+    return _ser.size() + _builder.size();
   }
 
-  int dedup_dict::transform_payload_and_add_objects(const char* payload, std::string& edited_payload, generic_event::object_list_t& object_ids, api_status* status)
+  int finalize(api_status* status)
```
olgavrou, line 238: `status` to default to `nullptr`, same as all other method signatures with `status`?

rlclientlib/dedup_internals.h
```diff
@@ -0,0 +1,140 @@
...
+    float _current;
+    const float _weight;
+  };
+
+  class zstd_compressor {
+  public:
+    const static int ZSTD_DEFAULT_COMPRESSION_LEVEL = 1;
+
+    explicit zstd_compressor(int level);
+    int compress(generic_event::payload_buffer_t& input, api_status* status) const;
```
olgavrou, line 62: same here? default `status` to `nullptr`?

rlclientlib/dedup_internals.h
```diff
@@ -0,0 +1,140 @@
...
+    int get_all_values(I start, I end, generic_event::object_list_t& action_ids, std::vector<string_view>& action_values, api_status* status);
+
+    template<typename I>
+    int remove_all_values(I start, I end, api_status* status);
+
+    void update_ewma(float value);
+    int compress(generic_event::payload_buffer_t& input, api_status* status) const;
+    int transform_payload_and_add_objects(const char* payload, std::string& edited_payload, generic_event::object_list_t& object_ids, api_status* status);
+
+    i_time_provider* get_time_provider() { return _time_provider.get(); }
```
olgavrou, line 85: as far as I can see, we don't check anywhere that time_provider is `nullptr` or not when it is passed into the `_time_provider` unique_ptr so it could be `nullptr` here resulting in a segmentation fault

rlclientlib/dedup_internals.h
```diff
@@ -0,0 +1,140 @@
...
+  public:
+
+    explicit action_dict_builder(dedup_state& state);
+
+    int add(const generic_event::object_list_t& object_ids, api_status* status);
+    size_t size() const;
+    int finalize(generic_event& evt, api_status* status);
+  private:
+    dedup_state& _state;
+    int _size_estimate;
```
olgavrou, line 109:
```suggestion
    size_t _size_estimate;
```

is there a reason for it to be an int?

rlclientlib/dedup.cc
kumpera: I designed dedup_dict as explicitly thread unsafe.
All synchronization is done on `dedup_state`.

rlclientlib/dedup.cc
kumpera: This usage of default values on internal APIs is wrong and we must stop doing it.
There is never a reason why we'd internally not pass a `status` objects. It only enables us to make mistakes.

rlclientlib/dedup_internals.h
kumpera: same as above

rlclientlib/dedup_internals.h
kumpera: `_time_provider` is a value of type `unique_ptr`. Accessing it will never result in a segfault.
Calling `get()` on it will either return a valid pointer or `nullptr`.

PR #225, opened 2020-11-11
Allow cb model loading with ccb arguments

slahabar:


rlclientlib/utility/config_utility.cc
```diff
@@ -103,7 +103,10 @@ namespace reinforcement_learning { namespace utility { namespace config {
       { "InitialExplorationEpsilon" , name::INITIAL_EPSILON },
       { "ModelRefreshIntervalMs"    , name::MODEL_REFRESH_INTERVAL_MS },
       { "QueueMode"                 , name::QUEUE_MODE }, // expect either DROP or BLOCK, default is DROP
-      { "LearningMode"              , name::LEARNING_MODE}
+      { "LearningMode"              , name::LEARNING_MODE},
+      { "VWInitialCommandLine"      , name::MODEL_VW_INITIAL_COMMAND_LINE},
+      { "VWCommandLine"             , name::VW_CMDLINE },
```
slahabar, line 108: upgrade using vw command line of the model instead of passing this value

rlclientlib/vw_model/safe_vw.cc
```diff
@@ -350,9 +371,22 @@ safe_vw_factory::safe_vw_factory(const model_management::model_data&& master_dat
...
   {}
 
-  safe_vw* safe_vw_factory::operator()()
-  {
-    if (_master_data.data())
+safe_vw_factory::safe_vw_factory(const model_management::model_data& master_data, const std::string& command_line)
+  : _master_data(master_data), _command_line(command_line)
+  {}
+
+safe_vw_factory::safe_vw_factory(const model_management::model_data&& master_data, const std::string& command_line)
```
olgavrou, line 378: When this constructor is called I can see we do `safe_vw_factory(std::move(data))` in vw_model.cc. To take advantage of actually moving and not copying the data the constructor should remove `const` (i.e. become `safe_vw_factory(model_management::model_data&& master_data, ...)` and also `master_data` it should be moved into `_master_data` (i.e. line 379 `_master_data(std::move(master_data))` )

rlclientlib/vw_model/safe_vw.cc
```diff
@@ -326,8 +334,8 @@ mm::model_type_t safe_vw::get_model_type(const VW::config::options_i* args)
 }
 
 bool safe_vw::is_compatible(const std::string& args) const {
-  const auto local_model_type = get_model_type(_vw->options);
-  const auto inbound_model_type = get_model_type(args);
+  const auto local_model_type = get_model_type(args);
```
homezcx, line 337: why did we swap the parameters we passed in in lines 337, 338?

rlclientlib/vw_model/safe_vw.cc
```diff
@@ -338,6 +346,19 @@ bool safe_vw::is_compatible(const std::string& args) const {
   return local_model_type == inbound_model_type;
 }
 
+bool safe_vw::is_CB_to_CCB_model_upgrade(const std::string& args) const
+{
+    const auto local_model_type = get_model_type(args);
+    const auto inbound_model_type = get_model_type(_vw->options);
+
+    if (local_model_type == mm::model_type_t::CCB && inbound_model_type == mm::model_type_t::CB)
```
homezcx, line 354: nit: line 354- line 359 could be simplified to one line

rlclientlib/vw_model/safe_vw.cc
```diff
@@ -338,6 +346,19 @@ bool safe_vw::is_compatible(const std::string& args) const {
   return local_model_type == inbound_model_type;
 }
 
+bool safe_vw::is_CB_to_CCB_model_upgrade(const std::string& args) const
+{
+    const auto local_model_type = get_model_type(args);
```
homezcx, line 351: just to double check with comment at line 337.

rlclientlib/vw_model/safe_vw.cc
```diff
@@ -25,8 +25,16 @@ namespace reinforcement_learning {
     _vw = VW::initialize("--quiet --json", &buf, false, nullptr, nullptr);
   }
 
-  safe_vw::safe_vw(std::string vw_commandline)
+  safe_vw::safe_vw(const char* model_data, size_t len, std::string vw_commandline)
```
ataymano, line 28: Actually looks like we had it before and impact is probably not that high, but it is weird to have string instead of const string& as argument here. Can you please fix it?

rlclientlib/vw_model/safe_vw.cc
```diff
@@ -25,8 +25,16 @@ namespace reinforcement_learning {
...
   {
+	io_buf buf;
+	buf.add_file(VW::io::create_buffer_view(model_data, len));
+
+	_vw = VW::initialize(vw_commandline, &buf, false, nullptr, nullptr);
+  }
+
+  safe_vw::safe_vw(std::string vw_commandline)
```
ataymano, line 36: const string&

rlclientlib/vw_model/safe_vw.cc
slahabar: The initial values were incorrect. It didn't affect functionality because there was only equality check.
local model is created from initial command line arguments. 
inbound model is created from vw model.

rlclientlib/vw_model/safe_vw.cc
slahabar: The initial values were incorrect. More details in the other comment.

rlclientlib/vw_model/safe_vw.cc
slahabar: Thanks Olga. That is great feedback. Here, I have copied the pattern from other constructor. I do not have much experience with c++ to be able to confidently make that change though.

olgavrou: Ah we do the same, probably by mistake. No worries I can patch it right after and tag you so that you see the change?

jackgerrits (approved):

olgavrou (approved):

ataymano (approved):

PR #229, opened 2020-11-20
Add id for slot event to v2 MultiSlotEvent.fbs

olgavrou:


rlclientlib/schema/v2/MultiSlotEvent.fbs
```diff
@@ -2,6 +2,7 @@
 namespace reinforcement_learning.messages.flatbuff.v2;
 
 table SlotEvent {
+    decision_slot_id:string;
```
ataymano, line 5: [nit] maybe just "id" is better here? In MultiSlotEvent->Slots->DecisionSlotId construction "Slot" seems too repetitive + not 100% about the role of Decision prefix

ataymano (approved):

rlclientlib/schema/v2/MultiSlotEvent.fbs
olgavrou: Good point about removing `slot`. I am also not sure about `decision_id` vs `id` as to not confuse it with the slates `slot_id` and make it clear that this is coming from the decision service

kumpera: New fields must *always* got at the end of a table. See https://google.github.io/flatbuffers/flatbuffers_guide_writing_schema.html

Additionally, please document with a comment those fields that are new.

kumpera (changes requested):

rlclientlib/schema/v2/MultiSlotEvent.fbs
olgavrou: Done

PR #234, opened 2020-11-23
Context helper to parse slot ids when provided

olgavrou:


rlclientlib/utility/context_helper.cc
```diff
@@ -144,4 +145,35 @@ namespace reinforcement_learning { namespace utility {
     }
     return error_code::success;
   }
+
+  int get_slot_ids_or_add_string_index(const char* context, const ContextInfo::index_vector_t& slots, std::vector<std::string>& slot_ids, i_trace* trace, api_status* status)
```
ormichae, line 149: [nit] I think get_slot_ids is enough for the name. If you want to elaborate I would change the second part to convert/generate_from_index

olgavrou: I tend to prefer elaborate names so opted for changing to: `or_generate_from_index`

rlclientlib/utility/context_helper.cc
```diff
@@ -144,4 +145,35 @@ namespace reinforcement_learning { namespace utility {
...
+        if (obj.HasParseError()) {
+          RETURN_ERROR_LS(trace, status, json_parse_error) << "JSON parse error: " << rj::GetParseError_En(obj.GetParseError()) << " (" << obj.GetErrorOffset() << ")";
+        }
+
+        const rj::Value::ConstMemberIterator& itr = obj.FindMember(slot_id);
+        if (itr != obj.MemberEnd() && itr->value.IsString()) {
+          slot_ids.emplace_back(std::string(itr->value.GetString()));
+        }
+        else {
+          slot_ids.emplace_back(std::to_string(slot_ids.size()));
```
zwd-ms, line 167: check for duplicates with provided slot ids?

unit_test/json_context_parse_test.cc
```diff
@@ -191,4 +191,85 @@ BOOST_AUTO_TEST_CASE(get_context_info_test) {
...
+BOOST_AUTO_TEST_CASE(get_slot_ids_some_slots_missing)
+{
+  auto const context = R"({
+    "UserAge":15,
+    "_multi":[
+      {"_text":"elections maine", "Source":"TV"},
+      {"Source":"www", "topic":4, "_label":"2:3:.3"}
+    ],
+    "_slots": [
+      {"a":4, "id": "provided_id_0"},
```
zwd-ms, line 258: what if the provided id is "1"? it will be the same as the auto-gen id for the slot 1.

olgavrou: That's a good point. Changed this method to just extract the provided id's and we can deal with them in the live-model side

rlclientlib/utility/context_helper.cc
olgavrou: My understanding is that we don't check for this at this point, and assume that user provided slot ids won't be identical here

rlclientlib/utility/context_helper.cc
```diff
@@ -17,6 +17,7 @@ namespace reinforcement_learning { namespace utility {
   const auto multi = "_multi";
   const auto slots = "_slots";
   const auto event_id = "_id";
+  const auto slot_id = "id";
```
ataymano, line 20: I think it "_id" is the right choice here given that we are going to generate uuids which are meaningless from learning point of view

olgavrou: That is true. So then provided slot ids will not be able to be used as features

olgavrou: Oh no actually, this is fine, we can use "id" here since it is in the context and can be used as feature. The uuids are not added in the context

ataymano: we can double check with product team, but I feel it is better to be consistent by default: if we have context when id is used only for joining and does not make sense for learning (when it is auto-generated), it is safer to have this assumption always.

olgavrou: switched to `_id`

ataymano (approved):

PR #236, opened 2020-11-24
multi slot to use string slot id or autogenerate one using uuid

olgavrou:
Depends on #234 
I'll add C# bindings for multi slot response (slot id) in next PR

### DIFF
--- a/include/multi_slot_response.h
+++ b/include/multi_slot_response.h
@@ -10,20 +10,17 @@ namespace reinforcement_learning {
   class api_status;
 
   /**
-   * @brief Holds triples (slot_id, action_id, probability) that tells which action was choosen for the given slot.
+   * @brief Holds (action_id, probability) that tells which action was choosen for the given slot.
    */
   struct slot_entry {
   public:
     ~slot_entry() = default;
 
-    slot_entry(uint32_t _slot_id, uint32_t _action_id, float _probability);
+    slot_entry(uint32_t _action_id, float _probability);
 
-    uint32_t get_slot_id() const;
     uint32_t get_action_id() const;
     float get_probability() const;
   private:
-    //! slot_id
-    uint32_t slot_id;
     //! action id
     uint32_t action_id;
     //! probability associated with the action id
--- a/include/multi_slot_response.h
+++ b/include/multi_slot_response.h
@@ -10,17 +10,20 @@ namespace reinforcement_learning {
   class api_status;
 
   /**
-   * @brief Holds (action_id, probability) that tells which action was choosen for the given slot.
+   * @brief Holds (id, action_id, probability) that tells which action was choosen for the given slot.
    */
   struct slot_entry {
   public:
     ~slot_entry() = default;
 
-    slot_entry(uint32_t _action_id, float _probability);
+    slot_entry(std::string id, uint32_t _action_id, float _probability);
 
+    const char* get_id() const;
     uint32_t get_action_id() const;
     float get_probability() const;
   private:
+    //! slot entry id
+    std::string id;
     //! action id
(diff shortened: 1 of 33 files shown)


include/multi_slot_response.h
```diff
@@ -10,17 +10,20 @@ namespace reinforcement_learning {
...
   /**
-   * @brief Holds (action_id, probability) that tells which action was choosen for the given slot.
+   * @brief Holds (id, action_id, probability) that tells which action was choosen for the given slot.
    */
   struct slot_entry {
   public:
     ~slot_entry() = default;
 
-    slot_entry(uint32_t _action_id, float _probability);
+    slot_entry(std::string id, uint32_t _action_id, float _probability);
```
ataymano, line 19: const string&?

include/multi_slot_response.h
```diff
@@ -46,7 +49,7 @@ namespace reinforcement_learning {
     ~multi_slot_response() = default;
 
     // push_back calls must be done in slot order
-    void push_back(uint32_t action_id, float prob);
+    void push_back(std::string id, uint32_t action_id, float prob);
```
ataymano, line 52: const string&?

ataymano (approved):

rlclientlib/live_model_impl.cc
```diff
@@ -226,17 +219,22 @@ namespace reinforcement_learning {
       RETURN_ERROR_LS(_trace_logger.get(), status, json_parse_error) << "There must be both a _multi field and _slots, and _multi must come first.";
     }
 
-    size_t num_decisions = context_info.slots.size();
+    std::vector<std::string> slot_ids(context_info.slots.size(), "");
+    std::map<size_t, std::string> found_ids;
+    RETURN_IF_FAIL(utility::get_slot_ids(context_json, context_info.slots, found_ids, _trace_logger.get(), status));
+    
+    autogenerate_missing_uuids(found_ids, slot_ids, _seed_shift);
```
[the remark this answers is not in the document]
olgavrou: You are right, generating the uuids would be slower than re-using the event_id. The final slot id will then be `<event_id>_<event_id>_<slot_index>`.

The concern I have here is that we are semantically overloading the slot index, which could potentially lead to errors down the road. One could see the slot index at the end of a full id and assume that that is the slot it is referring to, which in the case that some user-ids are provided would not be true.

rlclientlib/live_model_impl.cc
```diff
@@ -560,4 +558,19 @@ namespace reinforcement_learning {
 
     return error_code::success;
   }
+
+  void autogenerate_missing_uuids(const std::map<size_t, std::string>& found_ids, std::vector<std::string>& complete_ids, uint64_t seed_shift) {
+    for (auto ids : found_ids)
```
zwd-ms, line 563: const auto& id

unit_test/live_model_test.cc
```diff
@@ -822,11 +823,15 @@ BOOST_AUTO_TEST_CASE(slates_explore_only_mode) {
   auto it = response.begin();
   size_t action_id = 0;
   auto& slot_response = *it;
+  std::string first_slot_id = slot_response.get_id();
+  BOOST_CHECK_NE(slot_response.get_id(), ""); //uuid
```
zwd-ms, line 827: is there a way to test uuid?

zwd-ms (approved):

unit_test/live_model_test.cc
olgavrou: I don't know of any, so the best I could think of is to check that the first autogenerated id is not the same as the second autogenerated one

ataymano (approved):

PR #248, opened 2021-01-12
Ormichae/add baseline actions to multi slot decision

ormichae:
PR with necessary changes for adding apprentice mode to CCB. 

Idea: add baseline_actions argument to Rl client ccb endpoint. baseline_actions will be needed when joining, so it needs to be written to EH (already merged the FB schema change).

Live_model is becoming a mess with lots of overloads for request_multi_slot_decision


rlclientlib/live_model.cc
```diff
@@ -127,36 +136,73 @@ namespace reinforcement_learning
...
-	  INIT_CHECK();
-	  return _pimpl->request_multi_slot_decision(event_id, context_json, flags, resp, status);
+    INIT_CHECK();
+    return _pimpl->request_multi_slot_decision(event_id, context_json, flags, resp, live_model::default_baseline_vector, status);
   }
 
   int live_model::request_multi_slot_decision(const char * event_id, const char * context_json, multi_slot_response_detailed& resp, api_status* status)
   {
-	  return request_multi_slot_decision(event_id, context_json, action_flags::DEFAULT, resp, status);
+    return request_multi_slot_decision(event_id, context_json, action_flags::DEFAULT, resp, status);
```
olgavrou, line 174: whitespace

rlclientlib/live_model_impl.cc
```diff
@@ -114,23 +114,23 @@ namespace reinforcement_learning {
...
+
     if (_watchdog.has_background_error_been_reported())
     {
       RETURN_ERROR_LS(_trace_logger.get(), status, unhandled_background_error_occurred);
     }
 
     return error_code::success;
   }
-    
+
```
olgavrou, line 133: whitespace

rlclientlib/live_model_impl.cc
```diff
@@ -605,6 +624,6 @@ namespace reinforcement_learning {
       {
         complete_ids[i] = boost::uuids::to_string(boost::uuids::random_generator()()) + std::to_string(seed_shift);
       }
-    }    
+    }
```
olgavrou, line 627: whitespace

rlclientlib/live_model.cc
ormichae: It was a tab so I changed it into a 2 spaces to match formatting in file

rlclientlib/live_model_impl.cc
ormichae: added back in. I will remove trailing white spaces on different PR

rlclientlib/live_model_impl.cc
ormichae: added back in. I will remove trailing white spaces on different PR

olgavrou (commented):
Can we add some assertions for baseline in the `payload_serializer_test.cc`?

Also `unit_test/interaction.txt` seemed to have gotten checked in by mistake?

rlclientlib/serialization/payload_serializer.h
```diff
@@ -69,7 +69,7 @@ namespace reinforcement_learning {
 
     struct multi_slot_serializer : payload_serializer<generic_event::payload_type_t::PayloadType_Slates> {
       static generic_event::payload_buffer_t event(const char* context, unsigned int flags, const std::vector<std::vector<uint32_t>>& action_ids,
-        const std::vector<std::vector<float>>& pdfs, const std::string& model_version, const std::vector<std::string>& slot_ids) {
+        const std::vector<std::vector<float>>& pdfs, const std::string& model_version, const std::vector<std::string>& slot_ids, const std::vector<int>& baseline_actions = std::vector<int>()) {
```
olgavrou, line 72: it looks like you are passing in a vector in all the method calls, does it need a default?

rlclientlib/live_model.cc
```diff
@@ -49,6 +49,15 @@ namespace reinforcement_learning
     return err_code;
   }
 
+  std::vector<int> live_model::_c_array_to_vector(const int* c_array, size_t array_size)
```
olgavrou, line 52: nit: I haven't seen another method been named with an underscore, so maybe just `c_array_to_vector`?

olgavrou (commented):
Discussed a bit with Oren already offline. An alternative to creating a copy of the array into a vector and passing that around, would be to pass the array pointer and size all the way to the payload serializer and then do something like: ` auto fb_vector_offset=  fbb.CreateVector<type>(array_pointer, array_size);`

The downside to this is that it differs from the way we initialize the flatbuffer objects in that file (we won't be able to do `CreateEventXDirect` but will have to serialize all the other fb objects for that event first and then call `CreateEventX` instead). Not sure it is worth the hassle

rlclientlib/live_model.cc
ormichae: fixed

rlclientlib/live_model.cc
```diff
@@ -127,36 +136,73 @@ namespace reinforcement_learning
...
+    std::vector<int> baseline_vector = _c_array_to_vector(baseline_actions, baseline_actions_size);
+    if (event_id == nullptr && flags == NULL)
+    {
+      return _pimpl->request_multi_slot_decision(context_json, action_flags::DEFAULT, resp, baseline_vector, status);
+    }
+    if (event_id == nullptr)
+    {
+      return _pimpl->request_multi_slot_decision(context_json, flags, resp, baseline_vector, status);
+    }
+    if (flags == NULL)
```
olgavrou, line 159: We call into this via the `LiveModelRequestMultiSlotDecisionWithBaselineAndFlags` as far as I can see, could `flags` be passed in here explicitly as `NULL` ?

rlclientlib/live_model.cc
```diff
@@ -148,15 +176,33 @@ namespace reinforcement_learning
...
+    return _pimpl->request_multi_slot_decision(context_json, flags, resp, live_model::default_baseline_vector, status);
   }
 
   int live_model::request_multi_slot_decision(const char * context_json, multi_slot_response_detailed& resp, api_status* status)
   {
-	  return request_multi_slot_decision(context_json, action_flags::DEFAULT, resp, status);
+    return request_multi_slot_decision(context_json, action_flags::DEFAULT, resp, status);
   }
 
+  int live_model::request_multi_slot_decision(const char * event_id, const char * context_json, unsigned int flags, multi_slot_response_detailed& resp, const int* baseline_actions, size_t baseline_actions_size, api_status* status)
```
olgavrou, line 188: do the unit tests exercise all these combinations?

rlclientlib/live_model.cc
ormichae: Good point. As discussed, from the c# layer flags is an enum which is not nullable. For anyone calling directly into the c++ layer NULL == 0 == ActionFlags::DEFAULT so this is indeed an unnecessary check. Will remove

rlclientlib/live_model.cc
ormichae: added more tests

rlclientlib/live_model_impl.cc
```diff
@@ -220,16 +221,25 @@ namespace reinforcement_learning {
...
   {
     resp.clear();
 
+    if (_learning_mode == APPRENTICE)
+    {
+      if (baseline_actions.size() == 0)
+      {
+        return error_code::baseline_actions_not_defined;
+      }
+      //TODO: validity check for baseline actions
```
ataymano, line 240: [nit] maybe this check can be moved between populate_multi_slot_response and log_decisions, where we have more informations about slots, but it can be done as followup

rlclientlib/live_model_impl.cc
```diff
@@ -220,16 +221,25 @@ namespace reinforcement_learning {
...
   }
 
-  int live_model_impl::request_multi_slot_decision(const char * event_id, const char * context_json, unsigned int flags, multi_slot_response& resp, api_status* status)
+  int live_model_impl::request_multi_slot_decision(const char * event_id, const char * context_json, unsigned int flags, multi_slot_response& resp, const std::vector<int>& baseline_actions, api_status* status)
   {
     resp.clear();
 
+    if (_learning_mode == APPRENTICE)
+    {
+      if (baseline_actions.size() == 0)
```
ataymano, line 236: empty()

rlclientlib/live_model.cc
```diff
@@ -49,6 +49,15 @@ namespace reinforcement_learning
     return err_code;
   }
 
+  std::vector<int> live_model::c_array_to_vector(const int* c_array, size_t array_size)
+  {
+    if (c_array == nullptr)
+    {
+      return live_model::default_baseline_vector;
```
ataymano, line 56: it is weird to have method with generic name ("c_array_to_vector"), which actually returns default_baseline_vector (which is about baselines => more specific than arbitrary arrays). So I would replace reference to default_baseline_vector with just explicit vector() here.

include/live_model.h
```diff
@@ -407,6 +413,8 @@ namespace reinforcement_learning {
   private:
     std::unique_ptr<live_model_impl> _pimpl;  //! The actual implementation details are forwarded to this object (PIMPL pattern)
     bool _initialized = false;                //! Guard to ensure that live_model is properly initialized. i.e. init() was called and successfully initialized.
+    std::vector<int> default_baseline_vector = std::vector<int>();
```
ataymano, line 416: const

olgavrou (approved):
LGTM :)

rlclientlib/live_model_impl.cc
ormichae: I removed the comment for now. From what I remember we decided not to have a validity check for baselineactions on the RL client side (we have one on the APS side), but we can revisit this if needed.

rlclientlib/live_model.cc
ormichae: agreed

ataymano (approved):

rlclientlib/live_model_impl.cc
```diff
@@ -247,16 +253,25 @@ namespace reinforcement_learning {
...
+      {
+        return error_code::baseline_actions_not_defined;
+      }
+      //TODO: validity check for baseline actions
```
lokitoth, line 272: Is there a task for this todo?

bindings/cs/rl.net/LiveModel.cs
```diff
@@ -953,6 +1045,39 @@ public MultiSlotResponseDetailed RequestMultiSlotDecisionDetailed(string eventId
...
+            response = new MultiSlotResponseDetailed();
+
+            GC.KeepAlive(this);
+            return this.TryRequestMultiSlotDecisionDetailed(eventId, contextJson, flags, response, baselineActions, apiStatus);
+        }
+
+        public bool TryRequestMultiSlotDecisionDetailed(string eventId, string contextJson, ActionFlags flags, MultiSlotResponseDetailed response, int[] baselineActions, ApiStatus apiStatus)
+        {
+            int result = LiveModelRequestMultiSlotDecisionDetailedWithBaselineAndFlags(this.DangerousGetHandle(), eventId, contextJson, (uint)flags, response.DangerousGetHandle(), baselineActions, apiStatus.ToNativeHandleOrNullptrDangerous());
+            return result == NativeMethods.SuccessStatus;
```
lokitoth, line 1059: You should add `GC.KeepAlive(this)` (and maybe `GC.KeepAlive(response)` before the return, though that is probably not necessary, since it should be on the stack in this call), since otherwise the GC might decide to deallocate the object while the request is in flight.

bindings/cs/rl.net/LiveModel.cs
```diff
@@ -899,6 +964,33 @@ public MultiSlotResponse RequestMultiSlotDecision(string eventId, string context
...
+        {
+            response = new MultiSlotResponse();
+
+            GC.KeepAlive(this);
+            return this.TryRequestMultiSlotDecision(eventId, contextJson, flags, response, baselineActions, apiStatus);
+        }
+
+        public bool TryRequestMultiSlotDecision(string eventId, string contextJson, ActionFlags flags, MultiSlotResponse response, int[] baselineActions, ApiStatus apiStatus)
+        {
+            int result = LiveModelRequestMultiSlotDecisionWithBaselineAndFlags(this.DangerousGetHandle(), eventId, contextJson, (uint)flags, response.DangerousGetHandle(), baselineActions, apiStatus.ToNativeHandleOrNullptrDangerous());
```
lokitoth, line 977: You should add `GC.KeepAlive(this)` (and maybe `GC.KeepAlive(response)` before the return, though that is probably not necessary, since it should be on the stack in this call), since otherwise the GC might decide to deallocate the object while the request is in flight.

lokitoth (changes requested):

rlclientlib/live_model_impl.cc
ormichae: We decided not to have a basline action validity check on the rl client side. I forgot to remove the comment.

lokitoth (approved):

PR #249, opened 2021-01-15
Ormichae/set chosen action bug

ormichae:
This is the fix for https://msazure.visualstudio.com/One/_sprints/taskboard/DecisionService/One/AIR/DecisionService/%5BFe%5D/01.05-01.18?workitem=9115872. 

The bug was in set_chosen_action in slot_ranking. The associated test in live_model_test was not catching the issue so I changed it as well.


olgavrou (approved):

rlclientlib/slot_ranking.cc
```diff
@@ -24,7 +24,7 @@ namespace reinforcement_learning {
   }
 
   int slot_ranking::set_chosen_action_id(size_t action_id, api_status* status) {
-    if (action_id >= _ranking.size()) {
+    if (action_id > _ranking.size()) {
```
ataymano, line 27: action_ids presented in _ranking can be any set on nonnegative integers depending on combination of _inc field and actions chosen for previous slots (but not [0, _ranking.size]). 
So invariant that could be potentially verified here is something like _ranking.keys.contains(action_id).

Although it seems easier to remove action_id from set_chosen_action_id arguments (since we always populate with 0) and add 
validation if !action_ids.empty() in populate_multi_slot_response.

ataymano (changes requested):

rlclientlib/slot_ranking.cc
ormichae: resolved

ataymano (approved):

PR #255, opened 2021-02-03
Reset Actions for ccb logging only and apprentice mode

ormichae:
1. Enabled logging only mode for CCB 
2. added logic to reset actions for logging only and apprentice mode in ccb 

Workflow: 

Logging only: 
1. inference to populate response  
2. reset actions to original order 
3. log to EH 

Apprentice mode: 
1. inference to populate response 
2. log to EH 
3. reset actions to original order  

**Addressing a question that may come up when reviewing:**

CCB  logging only/apprentice reset actions step: we reset recommended action to "baseline" for each slot. There is no guarantee that baseline action will be part of the PDF associated with the slot (if it is the vw selected action for any of the above slots).  

**Decision based on discussion with Tyler Jeff Ed Sheetal and Dwai** 
Some clients use the probabilities to decide on actions to populate in a "carousel". This is where not having a probability associated with the recommended action may be problematic.  

In apprentice mode we promise that the user can move to online with a "flip of a switch" so we cannot have different functionality for apprentice and online mode.  

This issue is not applicable to anyone using APS because we do not return the list of action-probability pairs for CCB calls. We only return the recommended action.  The only case we will return the action-probability pairs is on a cb-ccb conversion, but in that case we will only have a single slot, so the baseline action is guaranteed to be part of the action probability list. 

 For 1-p customers calling in rl-client directly (only through the request_multi_slot_decision_detailed endpoint) there is no guarantee that when in apprentice mode the recommended action (baseline_action) will have a corresponding probability for each slot. 

1-p customers using ccb should call the request_multi_slot_decision endpoint (which does not have this issue because it does not return the PDF, only the recommended action). The request_multi_slot_decision_detailed endpoint was added for APS cb-ccb conversion reasons.

### DIFF
--- a/rlclientlib/live_model_impl.cc
+++ b/rlclientlib/live_model_impl.cc
@@ -632,7 +632,7 @@ namespace reinforcement_learning {
     int index = 0;
     for (auto &slot : response)
     {
-      if (baseline_actions.size() != 0 && baseline_actions.size() >= index)
+      if (!baseline_actions.empty() && baseline_actions.size() >= index)
       {
         slot.set_action_id(baseline_actions[index]);
       }
@@ -652,7 +652,7 @@ namespace reinforcement_learning {
     int index = 0;
     for (auto &slot : response)
     {
-      if (baseline_actions.size() != 0 && baseline_actions.size() >= index)
+      if (!baseline_actions.empty() && baseline_actions.size() >= index)
       {
         slot.set_chosen_action_id(baseline_actions[index]);
       }
--- a/rlclientlib/live_model_impl.cc
+++ b/rlclientlib/live_model_impl.cc
@@ -38,6 +38,8 @@ namespace reinforcement_learning {
   int check_null_or_empty(const char* arg1, i_trace* trace, api_status* status);
   int reset_action_order(ranking_response& response);
   void autogenerate_missing_uuids(const std::map<size_t, std::string>& found_ids, std::vector<std::string>& complete_ids, uint64_t seed_shift);
+  int reset_action_order_multi_slot(multi_slot_response& response, const std::vector<int>& baseline_actions = std::vector<int>());
+  int reset_action_order_multi_slot(multi_slot_response_detailed& response, const std::vector<int>& baseline_actions = std::vector<int>());
 
   void default_error_callback(const api_status& status, void* watchdog_context) {
     auto watchdog = static_cast<utility::watchdog*>(watchdog_context);
@@ -192,11 +194,6 @@ namespace reinforcement_learning {
 
   int live_model_impl::request_multi_slot_decision_impl(const char *event_id, const char * context_json, std::vector<std::string>& slot_ids, std::vector<std::vector<uint32_t>>& action_ids, std::vector<std::vector<float>>& action_pdfs, std::string& model_version, api_status* status)
   {
-    if (_learning_mode == LOGGINGONLY) {
-      //LoggingOnly mode are not supported here at this moment
-      return error_code::not_supported;
-    }
-
     //clear previous errors if any
     api_status::try_clear(status);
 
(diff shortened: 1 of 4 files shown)


rlclientlib/live_model_impl.cc
```diff
@@ -604,6 +627,45 @@ namespace reinforcement_learning {
     return error_code::success;
   }
 
+  int reset_action_order_multi_slot(multi_slot_response& response, const std::vector<int>& baseline_actions)
+  {
+    int index = 0;
+    for (auto &slot : response)
+    {
+      if (baseline_actions.size() != 0 && baseline_actions.size() >= index)
```
olgavrou, line 635: nit: `!baseline_actions.empty()`

rlclientlib/live_model_impl.cc
```diff
@@ -604,6 +627,45 @@ namespace reinforcement_learning {
...
+      {
+        slot.set_action_id(baseline_actions[index]);
+      }
+      else
+      {
+        //implicit baseline is the action corresponding to the slot index
+        slot.set_action_id(index);
+      }
+      slot.set_probability(1);
```
olgavrou, line 644: why is the probability set to 1 here?

ormichae: When we are in apprentice/logging only mode, the probability of returning the baseline action (either implicit or explicit) is 1.

rlclientlib/live_model_impl.cc
```diff
@@ -604,6 +627,45 @@ namespace reinforcement_learning {
     return error_code::success;
   }
 
+  int reset_action_order_multi_slot(multi_slot_response& response, const std::vector<int>& baseline_actions)
+  {
+    int index = 0;
```
olgavrou, line 632:
```suggestion
    size_t index = 0;
```

olgavrou (approved):

rlclientlib/live_model_impl.cc
ataymano: But do we want to have the same change for cb then?

ormichae: I explained in the description why we decided against it for the cb case. 

Some clients use the probabilities to decide on actions to populate in a "carousel". This is where not having a probability associated with the recommended action may be problematic. In apprentice mode we promise that the user can move to online with a "flip of a switch" so we cannot have different functionality for apprentice and online mode.

So basically in the cb case, some customers in apprentice mode put baseline action in the first "slot" of the carousel and then populate the rest of the carousel based on the probabilities in descending order. If we change the probabilities so that baseline action has a probability of 1 and the rest 0, this will break our promise that they can move to online mode with a "flip of a switch".

In the multi_slot_response case the customers do not have the PDF anymore (they only have recommended action and its associated probability) so this case scenario does not apply.

rlclientlib/live_model_impl.cc
```diff
@@ -604,6 +627,45 @@ namespace reinforcement_learning {
...
+      if (!baseline_actions.empty() && baseline_actions.size() >= index)
+      {
+        slot.set_action_id(baseline_actions[index]);
+      }
+      else
+      {
+        //implicit baseline is the action corresponding to the slot index
+        slot.set_action_id(index);
+      }
+      slot.set_probability(1);
```
jackgerrits, line 644: Mega nit:
```suggestion
      slot.set_probability(1.f);
```

jackgerrits (approved):

rlclientlib/live_model_impl.cc
```diff
@@ -244,8 +241,21 @@ namespace reinforcement_learning {
     RETURN_IF_FAIL(live_model_impl::request_multi_slot_decision_impl(event_id, context_json, slot_ids, action_ids, action_pdfs, model_version, status));
 
     RETURN_IF_FAIL(populate_multi_slot_response(action_ids, action_pdfs, std::string(event_id), std::string(model_version), slot_ids, resp, _trace_logger.get(), status));
+
+    if (_learning_mode == LOGGINGONLY)
```
ataymano, line 245: I am a bit confused here. I s difference between the LOGGINGONLY and APPRENTICE about patching response before or after logging it? That seems to be true for CB, but in this case resp object is not the one that logged, so either we are not doing what we want, or it is cleaner to concatenate these 2 ifs in one place.

ormichae: As far as rl client is concerned this is the only difference. LOGGINGONLY also doesn't train, but that logic is in the trainer. 

Regarding the two if statements, I agree 100% and I will merge them. In CCB it does not make sense to reset the action order because the list of actions available for each slot is not deterministic, so we just have to make this clear to whoever uses logging only.

rlclientlib/live_model_impl.cc
```diff
@@ -38,6 +38,8 @@ namespace reinforcement_learning {
   int check_null_or_empty(const char* arg1, i_trace* trace, api_status* status);
   int reset_action_order(ranking_response& response);
   void autogenerate_missing_uuids(const std::map<size_t, std::string>& found_ids, std::vector<std::string>& complete_ids, uint64_t seed_shift);
+  int reset_chosen_action_multi_slot(multi_slot_response& response, const std::vector<int>& baseline_actions = std::vector<int>());
```
ataymano, line 41: it seems like UB (next line as well)

olgavrou: Why is this UB?

ataymano: sorry it is not

ataymano (approved):

PR #256, opened 2021-02-05
Schema v2 rework schema to make events payloads completely free-standing.

kumpera:


kumpera:
I'll update my binary_joiner branch on top of this next.

rlclientlib/generic_event.cc
```diff
@@ -45,4 +47,14 @@ namespace reinforcement_learning {
   const generic_event::payload_buffer_t& generic_event::get_payload() const {
     return _payload;
   }
+
+  generic_event::encoding_type_t generic_event::get_encoding() const {
+    switch (_content_type) {
+      case event_content_type::ZSTD:
+        return encoding_type_t::EventEncoding_Zstd;
+      case event_content_type::IDENTITY:
+      default:
```
ataymano, line 56: seems leftover

rlclientlib/logger/logger_facade.h
```diff
@@ -27,10 +27,12 @@ namespace reinforcement_learning
 
       virtual ~i_logger_extensions();
 
-      virtual bool is_enabled() = 0;
+      virtual bool is_object_extraction_enabled() = 0;
```
ataymano, line 30: const?

rlclientlib/logger/logger_facade.h
```diff
@@ -27,10 +27,12 @@ namespace reinforcement_learning
...
+      virtual bool is_serialization_transform_enabled() = 0;
```
ataymano, line 31: const?

rlclientlib/schema/v2/Event.fbs
```diff
@@ -9,12 +9,15 @@ table Event {
 }
 
 table BatchMetadata {
-    content_encoding: string;
+    content_encoding: string; //valid values: IDENTITY and DEDUP
+}
+
+table SerializedEvent {
```
cheng-tan, line 15: q:Where does the meta field go for SerializedEvent?

rlclientlib/generic_event.cc
kumpera: That's not a leftover, we need to handle default is some way otherwise it's UB.

rlclientlib/logger/logger_facade.h
kumpera: good point.

rlclientlib/schema/v2/Event.fbs
kumpera: There isn't one, all metadata lives in Event itself.

rlclientlib/serialization/fb_serializer.h
```diff
@@ -269,32 +269,45 @@ namespace reinforcement_learning { namespace logger {
...
 
       const auto& ts = evt.get_client_time_gmt();
       v2::TimeStamp client_ts(ts.year, ts.month, ts.day, ts.hour,
         ts.minute, ts.second, ts.sub_second);
-      const auto meta_offset = v2::CreateMetadataDirect(builder, evt.get_id(), &client_ts, nullptr, evt.get_payload_type(), evt.get_pass_prob());
+      const auto meta_offset = v2::CreateMetadataDirect(
+        builder, 
+        evt.get_id(), 
+        &client_ts, 
+        nullptr, //FIXME app-id
```
olgavrou, line 298: Does this need to be fixed before the PR is merged?

rlclientlib/schema/v2/Event.fbs
```diff
@@ -9,12 +9,15 @@ table Event {
...
+    payload:[ubyte]; //serialized Event objects
 }
 
-// Collection of Ranking events
 table EventBatch {
-    events:[Event];
+    events:[SerializedEvent];
     metadata: BatchMetadata;
 }
```
olgavrou, line 22: Looking at the schema we would have to deserialize the payload in `SerializedEvent` to get the event-id, right? Would it be worth adding the event id as a field to `SerializedEvent`? Could that save us from having to deserialize the event to retrieve the event id?

kumpera:
Here's the branch with the new python joiner. The code is a  lot simpler.
https://github.com/kumpera/reinforcement_learning/tree/binary_joiner2

rlclientlib/dedup.cc
```diff
@@ -90,6 +90,11 @@ int dedup_dict::transform_payload_and_add_objects(const char* payload, std::stri
   return error_code::success;
 }
 
+int dedup_dict::size() const
```
olgavrou, line 93: `size_t`?

rlclientlib/logger/logger_extensions.cc
```diff
@@ -38,10 +40,9 @@ i_logger_extensions::~i_logger_extensions() { }
 
 
 i_logger_extensions* i_logger_extensions::get_extensions(const utility::configuration& config, i_time_provider* time_provider) {
-	const bool enable_dedup = config.get_int(name::PROTOCOL_VERSION, 1) == 2 &&
-		to_content_encoding_enum(config.get(name::INTERACTION_CONTENT_ENCODING, value::CONTENT_ENCODING_IDENTITY)) == content_encoding_enum::ZSTD_AND_DEDUP;
-
-	return enable_dedup ? create_dedup_logger_extension(config, time_provider) : new default_extensions(config, time_provider);
+	const char *section = "interaction"; //fixme lift this to live_model_impl;
```
olgavrou, line 43: Is this fixme still valid?

rlclientlib/schema/v2/Event.fbs
```diff
@@ -9,12 +9,15 @@ table Event {
...
 
 table BatchMetadata {
-    content_encoding: string;
+    content_encoding: string; //valid values: IDENTITY and DEDUP
```
olgavrou, line 12: should this maybe be an enum?

test_tools/log_parser/parser.py
```diff
@@ -110,19 +114,98 @@ def dump_event(evt, idx, is_dedup):
...
+        dump_event(batch.Events(i), i)
     print("----\n")
 
+
+def dump_preamble_file(file_name, buf):
+    preamble = parse_preamble(buf)
+    print(f'parsing preamble file {file_name}\n\tpreamble:{preamble}')
+    dump_event_batch(buf[PREAMBLE_LENGTH : PREAMBLE_LENGTH + preamble["msg_size"]])
+
+MSG_TYPE_HEADER = 0x55555555
```
olgavrou, line 134: Did this get checked in by mistake? The joined fb schema is missing

rlclientlib/dedup.cc
```diff
@@ -253,36 +277,59 @@ struct dedup_collection_serializer
...
 																									error_callback_fn* perror_cb, const char* section) override {
 		auto config = utility::get_batcher_config(_config, section);
 		int _dummy = 0;
-		return new logger::async_batcher<generic_event, dedup_collection_serializer>(
-				sender,
-				watchdog,
-				_dedup_state,
-				perror_cb,
-				config);
+    if(_use_dedup) {
```
olgavrou, line 287: `if(_use_dedup || _use_compression)`?

rlclientlib/serialization/fb_serializer.h
kumpera: after, this was already there before this PR.

rlclientlib/dedup.cc
```diff
@@ -149,13 +156,23 @@ void dedup_state::update_ewma(float value)
...
+int dedup_state::compress(generic_event::payload_buffer_t& input, event_content_type& content_type, api_status* status) const {
+  if(_use_compression) {
+    content_type = event_content_type::ZSTD;
+    return _compressor.compress(input, status);
+  }
+  content_type = event_content_type::IDENTITY;
+  return error_code::success;
 }
 
 int dedup_state::transform_payload_and_add_objects(const char* payload, std::string& edited_payload, generic_event::object_list_t& object_ids, api_status* status){
```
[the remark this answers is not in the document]
kumpera: Dedup_state could be renamed to something else.
Using a decorator stack, or splitting the logger_extensions interface, would allow for a more orthogonal design.

rlclientlib/schema/v2/Event.fbs
kumpera: A string allows for open ended values so that unknown values can be properly inspected by intermediaries.

test_tools/log_parser/parser.py
kumpera: yes, I checked in parser.py with partial support for the new file format :(

rlclientlib/dedup.cc
kumpera: good catch.

rlclientlib/dedup.cc
kumpera: This controls whether dedup_collection_serializer is used, which is a decorator on top of fb_collection_serializer that injects the dedup dictionary into the batch.

rlclientlib/logger/logger_extensions.cc
kumpera: Yes, it is. We should lift that to live_model_impl and make it so that extensions can be applied to observations as well.
I left as a FIXME because neither of the current extensions matter for observations since the payloads are too small to compress and are not dedup'able.

kumpera:
To summarize the issues in this PR:

the dedup_state type needs to be renamed or broken up apart.
we need to fix appid
logger_extensions currently has an implicit assumption of being only used for interactions, we should lift that restriction and remove the fixme.

olgavrou (approved):

PR #257, opened 2021-02-09
This is the scafold for the binary file format.

kumpera:
Update schema for file format.
Add sample joiner and parser.
Update example_gen to generate multiple examples per file.

### DIFF
--- a/rlclientlib/schema/v2/FileFormat.fbs
+++ b/rlclientlib/schema/v2/FileFormat.fbs
@@ -1,0 +1,25 @@
+include "Event.fbs";
+
+namespace reinforcement_learning.messages.flatbuff.v2;
+
+table JoinedEvent {
+    event: [ubyte];
+    timestamp: TimeStamp;
+}
+
+table JoinedPayload {
+    events: [JoinedEvent];
+}
+
+table KeyValue {
+    key: string;
+    value: string;
+}
+
+table FileHeader {
+    join_time: TimeStamp;
+    properties: [KeyValue];
+}
+
+root_type JoinedPayload;
+root_type FileHeader;
--- a/test_tools/log_parser/joiner.py
+++ b/test_tools/log_parser/joiner.py
@@ -1,0 +1,194 @@
+#! /usr/bin/env python3 -W ignore::DeprecationWarning
+
+from reinforcement_learning.messages.flatbuff.v2.EventBatch import *
+from reinforcement_learning.messages.flatbuff.v2.BatchMetadata import *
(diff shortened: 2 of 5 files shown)


ataymano (approved):

test_tools/log_parser/joiner.py
```diff
@@ -0,0 +1,194 @@
...
+            obs_ids += 1
+        observations[evt_id].append(ser_evt.PayloadAsNumpy())
+        obs_count +=1
+
+print(f'found {obs_count} observations with {obs_ids} ids')
+
+bin_f = BinLogWriter(result_file)
+bin_f.write_header({ 'eud': '-1', 'joiner': 'joiner.py', 'reward': 'latest'})
+
+for msg in interactions_file.messages():
```
olgavrou, line 178: So a joined batch could have events that do not have observations in them, right?

rlclientlib/schema/v2/FileFormat.fbs
```diff
@@ -0,0 +1,25 @@
+include "Event.fbs";
+
+namespace reinforcement_learning.messages.flatbuff.v2;
+
```
olgavrou, line 4: I am wondering whether this should be called plain `Event` or `ProcessedEvent` instead of `JoinedEvent` since `Joined` implies that the event has found its partner(s) and they are together now in the byte array. Where as from what I can see these will just be the events plus a timestamp

test_tools/log_parser/joiner.py
kumpera: Yes, interactions without observations should be included in the output, it's a left outer join.

rlclientlib/schema/v2/FileFormat.fbs
kumpera: Makes sense to change the name.

rlclientlib/schema/v2/FileFormat.fbs
```diff
@@ -0,0 +1,25 @@
...
+    value: string;
+}
+
+table FileHeader {
+    join_time: TimeStamp;
+    properties: [KeyValue];
+}
+
+root_type JoinedPayload;
+root_type FileHeader;
```
olgavrou, line 25: just fyi it seems that in c++ only one file is generated for this schema named: `FileFormat_generated.h` from that file only `GetFileHeader` is available, so if these stay in the same file we will have to do `flatbuffers::GetRoot<reinforcement_learning::messages::flatbuff::v2::JoinedPayload>(buffer)` to get JoinedPayload root

PR #258, opened 2021-02-22
Binary parser scaffolding

olgavrou:
Basic file parsing and event deserialising
Still a bit clunky but want to get early feedback
TODOs indicate where the interaction with VW will happen
Might be holding more information around than needed but that can be simplified later
Only processing CB for this PR


rlclientlib/deserialization/joined_log_parser.cc
```diff
@@ -0,0 +1,92 @@
...
+  fs.read(buffer.data(), buffer.size());
+  if (static_cast<int>(buffer[0]) != 1) {
+    return err::file_read_error;
+  }
+
+  // payload type, check for header
+  unsigned int payload_type;
+  fs.read((char *)(&payload_type), sizeof(payload_type));
+  if (payload_type != MSG_TYPE_HEADER) {
+    return err::file_read_error;
```
kumpera, line 50: We'll have to harden this later using ApiStatus and include better error messages.

rlclientlib/deserialization/joined_log_parser.cc
```diff
@@ -0,0 +1,92 @@
...
+  // read header size
+  fs.read(buffer.data(), buffer.size());
+  uint32_t payload_size = *reinterpret_cast<const uint32_t *>(buffer.data());
+  // read the payload
+  std::vector<char> payload(payload_size, 0);
+  fs.read(payload.data(), payload.size());
+
+  read_header(payload);
+
+  fs.read((char *)(&payload_type), sizeof(payload_type));
```
kumpera, line 62: We need to error check calls to `read`.  Since we don't configure the stream to throw, we need to check after every call.

kumpera (changes requested):
Is there a reason for merged.log to be in this PR?

Looks good otherwise.

olgavrou:
> Is there a reason for merged.log to be in this PR?
> 
> Looks good otherwise.

No, just forgotten, will remove

rlclientlib/deserialization/joined_log_parser.cc
olgavrou: Added `fs.fail()` checks, and there is a TODO to produce better error messages

olgavrou: will make the various TODOs into tasks

PR #259, opened 2021-02-22
ExampleJoiner to create vw examples and call learn (for new binary parser)

olgavrou:
Adding vw example creation and vw learning in the `ExampleJoiner` for the new binary parser

We need to do something similar with what `safe_vw` does, i.e. use an example pool. I have refactored that bit out of `safe_vw` so that it can be re-used, let me know if you think of a better alternative

Other options are: 
- duplicate the example pool code/logic
- extend `safe_vw` so that we can re-use it in `ExampleJoiner`


olgavrou:
closing in favor of #265

PR #263, opened 2021-03-01
Small updates to example_gen and parser.py

kumpera:
parser.py now recognize baseline actions in multi-slot decisions and
print them.

example_gen now has ccb-baseline to include baseline actions.

Changed the generated reward slot so it can be used together with ccb
examples to produce joined examples.

### DIFF
--- a/test_tools/example_gen/example_gen.cc
+++ b/test_tools/example_gen/example_gen.cc
@@ -22,6 +22,7 @@ bool enable_dedup = false;
 static const char *options[] = {
   "cb",
   "ccb",
+  "ccb-baseline",
   "slates",
   "ca",
   "f-reward",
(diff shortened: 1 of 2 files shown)


test_tools/example_gen/example_gen.cc
```diff
@@ -152,6 +155,14 @@ int take_action(r::live_model& rl, const char *event_id, int action) {
         if( rl.report_action_taken(event_id, &status) != err::success )
           std::cout << status.get_error_msg() << std::endl;
       break;
+    case CCB_BASELINE_ACTION: {// "ccb",
+      r::multi_slot_response response;
+      std::vector<int> baselines { 1, 0 };
+      if(rl.request_multi_slot_decision(event_id, JSON_CCB_CONTEXT, 0, response, &baselines[0], 2, &status) != err::success)
```
ormichae, line 161: Why are you only passing in baseline[0]?  the length of the baseline_actions passed in should equal the number of slots.

kumpera:
the OSX failure is spurious.

PR #264, opened 2021-03-02
reward functions

cheng-tan:
- parsed out reward function info
- added simple reward functions


kumpera (commented):
I thought about the flatbuffer schema/naming parts of this PR, but realized they should be addressed later once we figure out that all of this works well end-2-end.

It looks great but we still have to figure out how we'll make the reward code usable by other parts of our ecosystem.

olgavrou (approved):
LGTM

Outside the scope of this PR: on the whole making rewards re-usable front: could we potentially add everything reward related into a header file and have it imported where needed?

i.e. the reward function doesn't necessarily need everything in `joined_event`. The reward header file could define the function type, structs/objects it needs (e.g. `output_event`), and the different functions

PR #267, opened 2021-03-09
Add default reward to new external parser

cheng-tan:


external_parser/parse_example_binary.cc
```diff
@@ -148,8 +148,13 @@ bool binary_parser::parse_examples(vw *all, v_array<example *> &examples) {
 
     auto reward_function_info =
         flatbuffers::GetRoot<v2::RewardFunctionInfo>(payload);
```
olgavrou, line 150: if `_default_reward` is a member of `example_joiner` instead of `example_binary_parser` then it could be set here i.e. `example_joiner.set_default_reward(reward_function_info->default_reward())` and then it could just be used in `process_joined` instead of being passed in

external_parser/example_joiner.cc
```diff
@@ -12,10 +12,6 @@
 
 namespace RewardFunctions {
 float average(const joined_event &event) {
-  if (event.outcome_events.size() == 0) {
-    return 0.f;
-  }
-
   float sum = 0.f;
   for (const auto &o : event.outcome_events) {
```
cheng-tan, line 16: Assuming all the outcome_events won't have a NaN value.

external_parser/example_joiner.cc
```diff
@@ -251,16 +247,18 @@ int example_joiner::process_joined(v_array<example *> &examples) {
...
-  float reward = 0.f;
+  float reward = _default_reward;
   auto &je = _batch_grouped_examples[id];
-  if (je.interaction_metadata.payload_type == v2::PayloadType_CB &&
+  if (je.outcome_events.size() > 0) {
+    if (je.interaction_metadata.payload_type == v2::PayloadType_CB &&
       je.interaction_metadata.learning_mode ==
           v2::LearningModeType_Apprentice) {
-    if (je.interaction_data.actions[0] == 1) {
+      if (je.interaction_data.actions[0] == 1) {
```
cheng-tan, line 256: Will revisit the apprentice mode, here 1 is the baseline action.

olgavrou (approved):
LGTM

test_tools/log_parser/joiner.py
```diff
@@ -142,12 +142,14 @@ def write_join_msg(self, events):
         builder.Finish(joined_payload_off)
         self.write_message(MSG_TYPE_REGULAR, builder.Output())
```
olgavrou, line 143: does `parser.py` know how to parse this?

PR #271, opened 2021-03-11
Update live_model_impl.cc

jakub-szymanski:
vc 16.9.0 enforces that as clang do

### DIFF
--- a/rlclientlib/live_model_impl.cc
+++ b/rlclientlib/live_model_impl.cc
@@ -598,7 +598,6 @@ namespace reinforcement_learning {
   }
 
   int reset_action_order(ranking_response& response) {
-#ifdef __clang__
     std::vector<action_prob> tmp;
     std::copy(response.begin(), response.end(), std::back_inserter(tmp));
     std::sort(tmp.begin(), tmp.end(), [](const action_prob& a, const action_prob& b) {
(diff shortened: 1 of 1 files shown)


rlclientlib/live_model_impl.cc
```diff
@@ -598,20 +598,13 @@ namespace reinforcement_learning {
   }
 
   int reset_action_order(ranking_response& response) {
-#ifdef __clang__
     std::vector<action_prob> tmp;
```
rajan-chari, line 601: We try to avoid allocations when possible.  So this is undesirable.  I believe this was introduced to fix a Mac issue but the desirable approach is to not create a vector here.   It's better to do the sort on the existing collection.

rajan-chari (changes requested):

rlclientlib/live_model_impl.cc
kumpera: That's not possible because the `response` collection is not sortable. The standard requires a random access iterator and `response::begin`  returns a `container_iterator` which is tagged as a `std::forward_iterator_tag`.

Some compilers enforce the tag while others just take anything that implements all required methods. C++ is beautiful.

An alternative solution would be to change the tag in `container_iterator` to random access, but that would be a breaking change.

kumpera: Would that breaking change work for you, Jakub?

kumpera:
Here's a PR that changes container_iterator: https://github.com/VowpalWabbit/reinforcement_learning/pull/273

jakub-szymanski:
not needed

PR #275, opened 2021-03-12
Add build/test/valgrind runs for the binary parser in CI

olgavrou:


external_parser/example_joiner.cc
```diff
@@ -70,7 +70,8 @@ float median(const joined_event &event) {
 float earliest(const joined_event &event) { return 0.0; }
 } // namespace RewardFunctions
 
-example_joiner::example_joiner(vw *vw) : _vw(vw) {}
+example_joiner::example_joiner(vw *vw)
+    : _vw(vw), _reward_calculation(&RewardFunctions::average) {}
```
cheng-tan, line 74: default reward function is earliest

external_parser/unit_tests/test_example_joiner.cc
```diff
@@ -0,0 +1,42 @@
...
+#include <boost/test/unit_test.hpp>
+
+BOOST_AUTO_TEST_CASE(example_joiner_test) {
+  auto vw = VW::initialize("--quiet --binary_parser --cb_explore_adf", nullptr,
+                           false, nullptr, nullptr);
+
+  example_joiner joiner(vw);
+
+  std::string input_files = get_test_files_location();
+  auto interaction_buffer = read_file(input_files + "/cb_v2.fb");
```
cheng-tan, line 14: maybe add a comment of which command is used to generate cb_v2, f-reward_v2 and cb_simple?

external_parser/example_joiner.cc
olgavrou: thanks, changed

external_parser/unit_tests/test_example_joiner.cc
olgavrou: Good point, added a readme file under `unit_tests/test_files` so that we know how each file is generated

cheng-tan (approved):

PR #276, opened 2021-03-12
Fix a bunch of issues found by VS 2019 compiled and  /permissive-

kumpera:
In particular:

Make sure `container_iterator` and `const_container_iterator` correctly
work as random access iterators.

Fix `container_iterator::operator*()const` to return a non-const value.
A const interator is not the same thing as a const_interator. C++
constness is a constant source of joyfull surprises.

Beyond that, fix a couple variable size issues and a missing return code
in a falible function.

### DIFF
--- a/include/container_iterator.h
+++ b/include/container_iterator.h
@@ -48,7 +48,7 @@ namespace reinforcement_learning {
     TElem& operator*() {
       return _coll[_idx];
     }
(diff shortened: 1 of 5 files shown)


jakub-szymanski:
LGTM

olgavrou (approved):

PR #280, opened 2021-03-19
Reward function unit tests and test tool updates

cheng-tan:
TODO:
- [x] add test for earliest reward function
- [x] add test for apprentice mode


test_tools/log_parser/parser.py
```diff
@@ -141,6 +141,7 @@ def dump_preamble_file(file_name, buf):
     dump_event_batch(buf[PREAMBLE_LENGTH : PREAMBLE_LENGTH + preamble["msg_size"]])
 
 MSG_TYPE_HEADER = 0x55555555
+MSG_TYPE_REWARD_FUNCTION = 0x11111111
```
olgavrou, line 144: I think this is named `REWARD_FUNCTION` in the joiner and in the parser.. should we change the name there to be consistent?

external_parser/unit_tests/test_reward_functions.cc
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  BOOST_AUTO_TEST_CASE(average) {
+    float reward = get_float_reward(
+      "cb_v2.fb",
+      "f-reward_3obs_v2.fb",
+      v2::RewardFunctionType_Average
+    );
+
+    BOOST_CHECK_EQUAL(reward, (3 + 4 + 5) / 3);
```
olgavrou, line 63: c++ does integer division if all numbers are integers (e.g. `3/4` == `0`), since we are doing float reward calculation is that accounted for?

olgavrou (approved):

external_parser/example_joiner.cc
```diff
@@ -355,7 +355,8 @@ int example_joiner::process_joined(v_array<example *> &examples) {
     enqueued_time->tm_sec = joined_event->timestamp()->second();
 
     time_t enqueued_time_utc = mktime(enqueued_time);
-    std::cout << "enqueued time utc: " << enqueued_time_utc << std::endl;
+
```
olgavrou, line 358: nit: we can start removing the cout statements now that we have unit tests to check stuff

external_parser/unit_tests/test_reward_functions.cc
kumpera: we need to make at least one number a float. IE `(3.0 + 4 + 5) / 3`

PR #281, opened 2021-03-22
[Binary parser ] decompression and inter-batch caching

olgavrou:
- Supports zstd decompression
- Keeps dedup examples that are the same between dedup payloads (old ones evicted using a lru cache-style implementation)
- More tests


external_parser/example_joiner.cc
```diff
@@ -244,37 +275,30 @@ int example_joiner::process_outcome(const v2::Event &event,
...
   o_event.metadata = {"client_time_utc",
                       metadata.app_id() ? metadata.app_id()->str() : "",
-                      metadata.payload_type(),
-                      metadata.pass_probability(),
+                      metadata.payload_type(), metadata.pass_probability(),
                       metadata.encoding()};
   o_event.enqueued_time_utc = enqueued_time_utc;
 
-  auto outcome = v2::GetOutcomeEvent(event.payload()->data());
+  flatbuffers::DetachedBuffer detached_buffer;
```
cheng-tan, line 282: `detached_buffer` doesn't seem to be used

olgavrou: true, removed

PR #288, opened 2021-04-06
[parser] Update to latest schema and add large file goodies.

kumpera:
The two large file improvements are: `--short` and `--summary`.

`--short`: produce a one line message per event.
`--summary`: print one line per checkpoint message.

### DIFF
--- a/test_tools/log_parser/parser.py
+++ b/test_tools/log_parser/parser.py
@@ -6,10 +6,14 @@ import sys
 import json
 import struct
 import datetime
+import argparse
+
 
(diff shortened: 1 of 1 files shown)


olgavrou (commented):
Do we want to have the joiner.py produce checkpoints or is that something we will skip?

cheng-tan:
> Do we want to have the joiner.py produce checkpoints or is that something we will skip?

I'm updating RewardFunctionInfo to CheckpointInfo and add in the config of learning mode and problem type. I can update the joiner.py, it will be needed for the unit test files.

PR #291, opened 2021-04-12
Update reward function info to checkpoint info

cheng-tan:


external_parser/example_joiner.cc
```diff
@@ -149,6 +150,14 @@ void example_joiner::set_default_reward(float default_reward) {
   _default_reward = default_reward;
 }
 
+void example_joiner::set_learning_mode_config(const v2::LearningModeType learning_mode) {
```
olgavrou, line 153:
```suggestion
void example_joiner::set_learning_mode_config(const v2::LearningModeType& learning_mode) {
```

(same below)

olgavrou (approved):

PR #302, opened 2021-04-21
Fix a bunch of bugs plus an improvement to log_parser that I used to debug this.

kumpera:
parser.py can will now parse all preamble messages in a file.

rl.net.cli was not disposing its live model, now it does.

Introduced some basic logging to the batching process so we can do basic monitoring without having to a debugger or adding printf statements.

The extra logging turned out to be a problem because at-exit flushing was done from async_batch dtor and accessing a bunch of dangling pointers - fun stuff!

I fixed that by moving the flushing to live_model_impl, which owns everything so its dtor can do it safely.

### DIFF
--- a/bindings/cs/rl.net.cli/RLSimulator.cs
+++ b/bindings/cs/rl.net.cli/RLSimulator.cs
@@ -283,7 +283,7 @@ namespace Rl.Net.Cli
         }
     }
 
-    internal class RLSimulator
+    internal class RLSimulator : IDisposable
     {
         private RLDriver driver;
 
@@ -322,5 +322,29 @@ namespace Rl.Net.Cli
                 this.driver.OnError -= value;
             }
         }
+
+        #region IDisposable Support
+        private bool disposedValue = false; // To detect redundant calls
+
+        protected virtual void Dispose(bool disposing)
+        {
+            Console.WriteLine("Disponsing rlsim");
+            if (!disposedValue)
+            {
+                if (disposing)
+                {
+                    this.driver?.Dispose();
+                    this.driver = null;
+                }
+
+                disposedValue = true;
+            }
+        }
+
+        public void Dispose()
+        {
+            Dispose(true);
+        }
+        #endregion
     }
 }
--- a/rlclientlib/live_model_impl.cc
+++ b/rlclientlib/live_model_impl.cc
@@ -397,6 +397,13 @@ namespace reinforcement_learning {
     _learning_mode = learning::to_learning_mode(_configuration.get(name::LEARNING_MODE, value::LEARNING_MODE_ONLINE));
   }
 
+  live_model_impl::~live_model_impl() {
+    if (_interaction_logger)
+	  _interaction_logger->flush();
+	if (_outcome_logger)
+		_outcome_logger->flush();
+  }
+
   int live_model_impl::init_trace(api_status* status) {
     const auto trace_impl = _configuration.get(name::TRACE_LOG_IMPLEMENTATION, value::NULL_TRACE_LOGGER);
     i_trace* plogger;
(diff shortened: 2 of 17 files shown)


rlclientlib/live_model_impl.cc
```diff
@@ -397,6 +397,13 @@ namespace reinforcement_learning {
     _learning_mode = learning::to_learning_mode(_configuration.get(name::LEARNING_MODE, value::LEARNING_MODE_ONLINE));
   }
 
+  live_model_impl::~live_model_impl() {
+    if (_interaction_logger)
```
ataymano, line 401: how flushing in live_model destructor is better than doing it in async_batcher's one?

bindings/cs/rl.net.cli/RLSimulator.cs
```diff
@@ -322,5 +322,29 @@ public void Run(int steps = SimulatorStepProvider.InfinitySteps, int slots = Sim
                 this.driver.OnError -= value;
             }
         }
+
+        #region IDisposable Support
+        private bool disposedValue = false; // To detect redundant calls
```
olgavrou, line 327: just for my education: is this some kind of C# object disposal trick? why not put stuff in `Dispose()`?

rlclientlib/logger/async_batcher.h
```diff
@@ -46,27 +49,30 @@ namespace reinforcement_learning { namespace logger {
...
       api_status* status);
 
-    void flush(); //flush all batches
 
   public:
     async_batcher(i_message_sender* sender,
                   utility::watchdog& watchdog,
                   shared_state_t& shared_state,
                   error_callback_fn* perror_cb,
+                  i_trace *tracer,
```
olgavrou, line 65: nit: should we place the `trace` argument to the end since it is more likely to be missing than `config`?

rlclientlib/live_model_impl.cc
kumpera: Because of liveness. 
live_model dtor owns all the relevant objects and they will be valid during its dtor.
async_batcher doesn't own anything, so it's all potentially dangling pointers when its dtor is called.

I explained it in detail over one of the commits.

```
Add a dtor to live_model_impl and flush the queues from it.

Why flushing from live_model_impl is ok but from async_batcher is not?

async_batcher holds un-owned references to a lot of stuff and the
expectation is that those objects won't be valid during destruction.

Flushing data in async_batcher requires those dangling pointers to
be valid so we cannot do it from the dtor.

Then why is it ok to do it from live_model_impl dtor? As it turns out,
it holds owning references to pretty much everything that's needed for flushing.

Did you say pretty much? Yes, there are a lot of un-owned references in
live_model_impl as well but they look related to initialization only.
Life is too short otherwise.
```

bindings/cs/rl.net.cli/RLSimulator.cs
kumpera: It's not a trick, it's the recommended way to implement IDisposable.

It avoids a lot of common pitfalls associated with implementing it.

Here's an article explaining it:
https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose

rlclientlib/logger/async_batcher.h
kumpera: I find optional arguments in non-API methods to be a terribly bad idea as it makes REALLY EFFING HARD to correctly wire them around.

It's not a compiler error to miss an optional argument and that will lead to stuff mysteriously not working.
There's no ergonomic value in optional argument for non-published APIs.

I made the change this way so i_trace doesn't have to be optional.

rlclientlib/live_model_impl.cc
ataymano: yeah that makes sense, thank you

ataymano (approved):
LGTM. Although looks like there is some consistent issue with indentation (maybe spaces vs tabs?)

kumpera:
> 
> 
> LGTM. Although looks like there is some consistent issue with indentation (maybe spaces vs tabs?)

Yes, formatting is all bad because I used VS which I can't figure out how to make it respect our style. I'll fix it using clang-fmt before committing.

olgavrou (approved):

PR #303, opened 2021-04-22
fb to dsjson log converter

cheng-tan:
TODO:
- [x] Assign reward function, default reward etc.
- [x]  Enable this by passing a flag to external parser
- [x]  Verify original label cost if apprentice is off
- [x]  Format TimeStamp field


external_parser/example_joiner.h
```diff
@@ -82,7 +84,13 @@ class example_joiner {
   // true if there are still event-groups to be processed from a deserialized
   // batch
   bool processing_batch();
+
+  joined_event get_grouped_example();
+  std::vector<const v2::JoinedEvent *> get_grouped_events();
+
   float get_reward();
```
olgavrou, line 91: nit: maybe add a comment somewhere explaining what original reward means

external_parser/parse_example_external.cc
```diff
@@ -13,6 +15,18 @@ bool parser_options::is_enabled() { return binary; }
 std::unique_ptr<parser>
 parser::get_external_parser(vw *all, const input_options &parsed_options) {
   if (parsed_options.ext_opts->binary) {
+    bool binary_to_json = parsed_options.ext_opts->binary_to_json;
+    if (binary_to_json) {
+      std::string infile_name = all->data_filename;
+      std::string outfile_name = infile_name.substr(
```
olgavrou, line 21: should probably never happen, but maybe check that the filename doesn't already end in `dsjson` and if it does log a warning?

olgavrou: or maybe even abort, because it looks like on line 24 we are trying to delete it, so might end up deleting an input file if someone calls this with a dsjson file by mistake

external_parser/example_joiner.cc
```diff
@@ -458,20 +476,28 @@ int example_joiner::process_joined(v_array<example *> &examples) {
...
+        _reward = _original_reward;
       }
     } else {
-      _reward = _reward_calculation(je);
+      _reward = _original_reward;
     }
   }
 
-  try_set_label(je, _reward, examples);
+  if (_binary_to_json) {
```
olgavrou, line 494: could open the file in the constructor of `example_joiner` and just write to it on every call, and then close the file in the destructor

external_parser/example_joiner.cc
```diff
@@ -304,8 +309,19 @@ int example_joiner::process_interaction(const v2::Event &event,
...
+      std::move(meta),
+      std::move(data)
+    };
+
+    if (_binary_to_json) {
+      je.context = context;
+      je.model_id = cb->model_id()->data();
+    }
+
+    std::string line_vec(context, cb->context()->size());
```
olgavrou, line 324: could we instead of keeping a `const char *` in `joined_event` for the context, keep a `std::string` and copy it into the `je` struct after line 324? I think it's best to copy the string otherwise not sure what happens if the json parser performs destructive parsing

external_parser/example_joiner.cc
```diff
@@ -231,7 +240,7 @@ const T *example_joiner::process_compression(const uint8_t *data, size_t size,
   return payload;
 }
 
-void example_joiner::try_set_label(const joined_event &je, float reward,
+void example_joiner::try_set_label(const joined_event &je,
```
olgavrou, line 243: nice :)

external_parser/parse_example_external.cc
```diff
@@ -13,6 +15,22 @@ bool parser_options::is_enabled() { return binary; }
...
+    if (binary_to_json) {
+      std::string infile_path = all->data_filename;
+      std::string infile_name = infile_path.substr(
+        0, infile_path.find_last_of('.'));
+      std::string infile_extension = infile_path.substr(
+        infile_path.find_last_of(".") + 1);
+
+      if (infile_extension == "dsjson") {
+        throw std::runtime_error("input file for --binary_to_json option should"
+        " be binary format");
```
olgavrou, line 28:
```suggestion
        " be binary format, file provided: " + infile_path);
```

external_parser/parse_example_external.cc
```diff
@@ -13,6 +15,22 @@ bool parser_options::is_enabled() { return binary; }
 std::unique_ptr<parser>
 parser::get_external_parser(vw *all, const input_options &parsed_options) {
   if (parsed_options.ext_opts->binary) {
+    bool binary_to_json = parsed_options.ext_opts->binary_to_json;
+    if (binary_to_json) {
+      std::string infile_path = all->data_filename;
```
olgavrou, line 20:
```suggestion
      const auto& infile_path = all->data_filename;
```

external_parser/parse_example_external.cc
```diff
@@ -13,6 +15,22 @@ bool parser_options::is_enabled() { return binary; }
...
+      std::string infile_name = infile_path.substr(
```
olgavrou, line 21:
```suggestion
     const auto& infile_name = infile_path.substr(
```

external_parser/parse_example_external.cc
```diff
@@ -13,6 +15,22 @@ bool parser_options::is_enabled() { return binary; }
...
+        0, infile_path.find_last_of('.'));
+      std::string infile_extension = infile_path.substr(
```
olgavrou, line 23:
```suggestion
      const auto& infile_extension = infile_path.substr(
```

olgavrou (approved):
left some nit suggestions, otherwise LGTM

PR #311, opened 2021-05-06
[Binary parser] more error handling and logging

olgavrou:
- checking and handling errors -> cleanup if something went wrong and continue if you can
  - will skip an interaction if the interaction could not be processed or if one of its outcomes could not be processed
  - will skip an entire batch if a dedup payload can't be processed
- logging using:
  - `warn` if we can cleanup and continue while skipping the interaction
  - `error` when we can cleanup and continue but are skipping an entire batch
  - `critical` when we can't continue and so are exiting

before merging will add some tests/testfiles to follow the error paths


external_parser/example_joiner.cc
```diff
@@ -140,20 +141,25 @@ void example_joiner::return_example_f(void *vw, example *ex) {
   ((example_joiner *)vw)->return_example(ex);
 }
 
-int example_joiner::process_event(const v2::JoinedEvent &joined_event) {
+bool example_joiner::process_event(const v2::JoinedEvent &joined_event) {
```
kumpera, line 144: Don't we need to surface rich info about the errors so the top level caller knows why it failed?

external_parser/example_joiner.cc
```diff
@@ -273,38 +288,66 @@ void example_joiner::try_set_label(const joined_event &je,
...
+}
+
+bool example_joiner::process_interaction(const v2::Event &event,
+                                         const v2::Metadata &metadata,
+                                         const TimePoint &enqueued_time_utc,
+                                         v_array<example *> &examples) {
+
+  if (EnumNamePayloadType(metadata.payload_type()) !=
+      EnumNameProblemType(_problem_type_config)) {
+    VW::io::logger::log_warn(
```
kumpera, line 318: I find it weird that we're calling a global function and not an instance method

external_parser/example_joiner.cc
```diff
@@ -315,52 +358,73 @@ int example_joiner::process_interaction(const v2::Event &event,
...
-    } else {
-      VW::template read_line_json<false>(
-          *_vw, examples, const_cast<char *>(line_vec.c_str()),
-          reinterpret_cast<VW::example_factory_t>(&VW::get_unused_example), _vw,
-          &_dedup_cache.dedup_examples);
+    joined_event je{enqueued_time_utc,
+                    {timestamp_to_chrono(*metadata.client_time_utc()),
+                     metadata.app_id() ? metadata.app_id()->str() : "",
+                     metadata.payload_type(), metadata.pass_probability(),
+                     metadata.encoding(), metadata.id()->str(), learning_mode},
```
kumpera, line 368: did we check if `metadata.id()` was valid before here?

external_parser/example_joiner.cc
```diff
@@ -315,52 +358,73 @@ int example_joiner::process_interaction(const v2::Event &event,
...
+                    std::move(data),
+                    line_vec,
+                    cb->model_id()->data()};
```
kumpera, line 371: Same question, did we check? Plus, model is is string IIRC.

external_parser/example_joiner.cc
olgavrou: I'll check with Peter but I think that that is how the logs are currently implemented, there is no instance method afaik

external_parser/example_joiner.cc
olgavrou: My approach is to keep things as local as possible, logging where appropriate and handling errors in a logical way and at the lowest level possible.

We are lucky here in that if something fails in a lower level call we need to either clean and continue or just continue to the next payload

external_parser/example_joiner.cc
peterychang: There are plans to create an instanced logger, but it requires a lot of piping and since the change was kind of urgent I went with a global logger for now

external_parser/example_joiner.cc
olgavrou: yes it's been checked on a higher level

external_parser/example_joiner.cc
olgavrou: yes checked, good point it is a string, changing

PR #313, opened 2021-05-10
Ataymano/trainer join

ataymano:
per episode join.
still in progress:
- topological sort
- tests


ataymano:
Please ignore - it is already stale. Let me split it into 2 parts:
- common changes around joiner interface PR to master
- multistep logic PR to feature branch

PR #314, opened 2021-05-10
[Binary parser] refactor reward related structs and methods

olgavrou:
move into `reward.h`

### DIFF
--- a/external_parser/example_joiner.cc
+++ b/external_parser/example_joiner.cc
@@ -1,6 +1,7 @@
 #include "example_joiner.h"
 #include "log_converter.h"
 
+#include "generated/v2/CbEvent_generated.h"
 #include "generated/v2/DedupInfo_generated.h"
 #include "generated/v2/Event_generated.h"
 #include "generated/v2/Metadata_generated.h"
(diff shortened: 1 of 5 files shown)


external_parser/example_joiner.cc
```diff
@@ -11,90 +12,16 @@
...
 
 // VW headers
-#include "example.h"
 #include "io/logger.h"
 #include "parse_example_json.h"
 #include "parser.h"
-#include "v_array.h"
-
-namespace RewardFunctions {
-float average(const joined_event &event) {
```
ataymano, line 21: not for this PR, but just a question:
is there a strong reason why this interface takes joined_event as input but not const <outcome_event>&?

ataymano (approved):

external_parser/example_joiner.cc
olgavrou: I think mostly future-proofing since a more complex reward function might need more information. We could make it more strict and only use `outcome_event` if we see we don't use anything else from `joiend_event` @cheng-tan what do you think?

PR #320, opened 2021-05-18
Gets CMake working on Windows

lokitoth:
* Also enables building dotnet bindings via CMake on Windows

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,12 @@
 cmake_minimum_required(VERSION 3.5)
+
+if(WIN32)
+  # Due to needing to configure the CMAKE platform, this needs to be included before the
+  # top-level project() declaration.
+  include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/platforms/win32.cmake)
+  message(STATUS "WinSDK Version: ${CMAKE_SYSTEM_VERSION}")
+endif()
+
 project(reinforcement_learning)
 
 set(CMAKE_CXX_STANDARD 11)
@@ -14,7 +22,7 @@ if (vw_USE_AZURE_FACTORIES)
 endif()
 
 find_package(OpenSSL REQUIRED)
-find_package(Boost COMPONENTS unit_test_framework system program_options thread REQUIRED)
+find_package(Boost COMPONENTS unit_test_framework system program_options thread REQUIRED) # Boost::uuid too, but that cannot be "found" here
 
 # On MacOS, Boost::thread needs to be linked separately.
 if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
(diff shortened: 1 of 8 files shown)


cmake/platforms/win32.cmake
```diff
@@ -0,0 +1,31 @@
...
+if(WIN32)
+  # VW targets Windows 10.0.16299.0 SDK
+  set(CMAKE_SYSTEM_VERSION "10.0.16299.0" CACHE INTERNAL "Windows SDK version to target.")
+
+  # TODO: Enable building .NET bindings via CMake on Linux 
+  option(rlclientlib_BUILD_DOTNET "Build .NET bindings" ON)
+
+  if (rlclientlib_BUILD_DOTNET)
+    cmake_minimum_required(VERSION 3.14)
+    # set(CMAKE_DOTNET_TARGET_FRAMEWORK_VERSION  "v4.5.2" CACHE INTERNAL ".NET Framework SDK version to target.")
```
olgavrou, line 20: remove commented out code

CMakeLists.txt
```diff
@@ -14,7 +22,7 @@ if (vw_USE_AZURE_FACTORIES)
 endif()
 
 find_package(OpenSSL REQUIRED)
-find_package(Boost COMPONENTS unit_test_framework system program_options thread REQUIRED)
+find_package(Boost COMPONENTS unit_test_framework system program_options thread REQUIRED) # Boost::uuid too, but that cannot be "found" here
```
olgavrou, line 25: where is it found?

bindings/cs/CMakeLists.txt
```diff
@@ -1 +1,29 @@
-add_subdirectory(rl.net.native)
+if (rlclientlib_BUILD_DOTNET)
+  set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/Modules/")
+  include(FindDotnet)
+
+  add_subdirectory(rl.net.native)
+
+  add_custom_target(rl.net
```
jackgerrits, line 7: Can you add the sources to these targets so when a VS solution is generated they show up?

CMakeLists.txt
lokitoth: It is a header-only library, which cannot be found as a component by find_package()

bindings/cs/CMakeLists.txt
lokitoth: Done. At that point it makes sense to break them out into separate CMakeLists.txt files. 

It is somewhat of an impedance mismatch between CMake and the dotnet toolchain, as .Net developers will not naturally think to add new sources into the CMakeLists.txt file. That was one of the big gains in Sdk-style projects: including all source in the tree by default.

jackgerrits (approved):

olgavrou (approved):
LGTM :)

PR #322, opened 2021-05-18
add cb-loop to example_gen and make it accept a config file

olgavrou:


test_tools/example_gen/example_gen.cc
```diff
@@ -184,6 +213,22 @@ int take_action(r::live_model& rl, const char *event_id, int action, bool gen_ra
           std::cout << status.get_error_msg() << std::endl;
       break;
     };
+    case CB_LOOP: { // "cb action and random number of float rewards"
+      r::ranking_response response;
+      std::cout << "choose rank for event: " << event_id << std::endl;
+      if(rl.choose_rank(event_id, JSON_CB_CONTEXT, response, &status))
+          std::cout << status.get_error_msg() << std::endl;
+
+      size_t num_of_rewards = get_random_number(rng);
```
kumpera, line 222: This will produce [1,5[ rewards per decision. This is fine for example_gen, but it doesn't follow real world distributions.

kumpera (approved):

test_tools/example_gen/example_gen.cc
olgavrou: Yup, this was mostly in order to get more than one observation so that we can test the reward functions

PR #323, opened 2021-05-18
[Binary parser] add test that compares json and fb joined logs's outputs

olgavrou:


external_parser/unit_tests/test_vw_external_parser.cc
```diff
@@ -73,4 +74,89 @@ BOOST_AUTO_TEST_CASE(cb_dedup_compressed) {
...
+
+    auto vw = VW::initialize("--cb_explore_adf --binary_parser --quiet -f " + fb_model +
+                                 " -d " + full_file_name,
+                             nullptr, false, nullptr, nullptr);
+
+    v_array<example *> examples;
+    examples.push_back(&VW::get_unused_example(vw));
+    multi_ex multi_exs;
+    bool read_payload = false;
+    while (vw->example_parser->reader(vw, examples) > 0) {
```
lalo, line 102: could this logic be replaced by calls similar to this https://github.com/VowpalWabbit/vowpal_wabbit/blob/b5939a84b16d27dfe6ab9cec7360be415899026f/python/pylibvw.cc#L264

olgavrou: yeah probably, let me try (thanks!)

olgavrou: yeah works fine

external_parser/unit_tests/test_files/test_outputs/.gitignore
```diff
@@ -0,0 +1,4 @@
+# Ignore everything in this directory
```
lalo, line 1: nit (don't care a lot): could possibly be migrated to root gitignore file

lalo (approved):

external_parser/unit_tests/test_files/test_outputs/.gitignore
olgavrou: yeah tried that but afaik git doesn't let you add a directory if it is empty so just added a gitignore file to it

PR #329, opened 2021-05-20
[Binary parser] add file format to README

olgavrou:


cheng-tan (approved):
lgtm. do you think it worth mentioning one file can contain only one header but can include multiple checkpoint and regular message. eof is not required?

PR #332, opened 2021-05-21
Enable import of native msbuild projects into CMake-generated solution

lokitoth:
This enables an option `Drlclientlib_DOTNET_USE_MSPROJECT=On` to generate a CMake-based solution file which directly imports the .NET projects (rather than relying on a custom command to drive `dotnet build`), and enables the native debug experiences in Visual Studio.

Also adds a utility target `launch_vs` to start Visual Studio passing the CMake build environment on to it; in particular, it enables passing the `BinaryOutputBase` environment variable to ensure that the output path of the managed assemblies is in the same places as the native interop assemblies, working around the issue that native assemblies do not get properly copied when dependent managed projects are built.

 Usage example (from my  `.vscode/settings.json`):

```
{
    "cmake.configureArgs": [
        "-DCMAKE_TOOLCHAIN_FILE=C:/w/vcpkg/scripts/buildsystems/vcpkg.cmake",
        "-DrlclientlibBUILD_DOTNET=On",
        "-Drlclientlib_DOTNET_USE_MSPROJECT=On"
    ]
}
```

Alternatively, when using CMake on the command line:

```
cmake -DCMAKE_TOOLCHAIN_FILE=C:/w/vcpkg/scripts/buildsystems/vcpkg.cmake 
      -DrlclientlibBUILD_DOTNET=On 
      -Drlclientlib_DOTNET_USE_MSPROJECT=On 
      -B build -S . 
cmake --build build --target launch_vs
```


bindings/cs/rl.net.cli.test/RegressionTest.cs
```diff
@@ -0,0 +1,94 @@
...
+
+
+        [TestMethod]
+        public void Test_PPTDesigner()
+        {
+            
+        }
+
+*/
+
```
jackgerrits, line 11: Should this be here?

bindings/cs/rl.net.cli.test/RegressionTest.cs
```diff
@@ -0,0 +1,94 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
```
jackgerrits, line 12: This file seems like it may have been added by accident?

jackgerrits (approved):
Looks good apart from it looks like an unrelated test file snuck in

bindings/cs/rl.net.cli.test/RegressionTest.cs
lokitoth: Nope, good catch. Also there was a superfluous reference to the corresponding test file in the project.

jackgerrits (approved):

PR #336, opened 2021-05-27
feat: Generate and upload docs for every commit

jackgerrits:
- Generates and uploads C++ and Python docs to https://vowpalwabbit.org/docs
- Removes old header from doxygen docs used to navigate between C++/Python


.github/workflows/build_docs.yml
```diff
@@ -0,0 +1,114 @@
+name: Build Documentation
+
+# Build documentation on:
+# - pushes to master and uploads to latest folder
+# - Released and uploads to folder that matches release tag
+on:
+  push:
+    branches:
```
olgavrou, line 8: is it worth setting this to `pr` so that the pipeline can run for the PR and then when it's ok and working re-set it to `master`?

jackgerrits:
I don't think the action will run from a PR if it isn't yet merged into the target branch. At least that's what it seemed like when I added the original Python CI.

olgavrou (approved):

PR #341, opened 2021-06-02
Ataymano/multistep joiner

ataymano:
multistep joiner.
topological sort + tests in next PR


external_parser/multistep_example_joiner.cc
```diff
@@ -0,0 +1,246 @@
...
+
+  _order.pop();
+  return true;
+}
+
+bool multistep_example_joiner::processing_batch() {
+  return _sorted && !_order.empty();
+}
+
+void multistep_example_joiner::on_new_batch() {
```
olgavrou, line 236: could `on_new_batch` and maybe `on_batch_read` be operations that are done at the end of every batch instead? that way they don't need to be introduced at the binary parser level and each batch can just cleanup

external_parser/parse_example_external.cc
```diff
@@ -38,6 +39,9 @@ parser::get_external_parser(vw *all, const input_options &parsed_options) {
     } else {
```
olgavrou, line 39: what about:

```
if (binary_to_json)
{
      joiner = ...
} else if (multistep)
{
   joiner = VW::make_unique<multistep_example_joiner>(all);
}
else
{
    joiner = VW::make_unique<example_joiner>(all);
}
```

from my understanding to enable this we need to pass both `--binary_parser` and `--multistep` right?

external_parser/multistep_example_joiner.cc
```diff
@@ -0,0 +1,246 @@
...
+                         ? timestamp_to_chrono(*metadata.client_time_utc())
+                         : TimePoint(),
+                        metadata.app_id() ? metadata.app_id()->str() : "",
+                        metadata.payload_type(),
+                        metadata.pass_probability(),
+                        metadata.encoding(),
+                        metadata.id()->str(),
+                        v2::LearningModeType::LearningModeType_Online};
+
+  DecisionServiceInteraction data;
```
olgavrou, line 143: I wonder if some of this could be moved to the parent class.. maybe as a next step/refactor we can look at what we could refactor out

external_parser/multistep_example_joiner.cc
olgavrou: on second thought these could be nice for future potential implementations as a clear indication that we are switching to a new batch

external_parser/parse_example_external.cc
ataymano: yes, that is true

external_parser/multistep_example_joiner.cc
ataymano: yes, there are number of that kind of improvements that I am planning to do in next PRs

olgavrou (approved):

PR #347, opened 2021-06-09
[Binary parser] ccb parsing

olgavrou:
- fill in ccb related methods
- unit tests
- separated fill in label and set cost (not sure about this one)


external_parser/joiners/example_joiner.cc
```diff
@@ -470,7 +465,11 @@ bool example_joiner::process_joined(v_array<example *> &examples) {
     return true;
   }
 
-  je.fill_in_label(examples, reward);
+  je.fill_in_label(examples);
```
olgavrou, line 468: @cheng-tan what do you think about this separation?

cheng-tan (approved):

PR #349, opened 2021-06-09
vw pdrop hotfix pickup

ataymano:
replacing latest vw update with the hotfix from here:
https://github.com/VowpalWabbit/vowpal_wabbit/tree/hotfix-06092021-pdrop_learning


olgavrou:
the patch seems to have removed the check for `pdrop ==1.` was that intentional?

ataymano:
Yes, that part of change was heavily using Eduardo's changing which we are trying to bypass. Removing this check is the simplest thing to do, and is relatively safe for short time hotfix, since client library is never producing pdrop==1.
Can add explicit throw though.

olgavrou:
> Yes, that part of change was heavily using changes from ignoring bad examples which we are trying to bypass. Removing this check is the simplest thing to do, and is relatively safe for short time hotfix, since client library is never producing pdrop==1.
> Can add explicit throw though.

makes sense

lalo (approved):

PR #351, opened 2021-06-17
pdrop fix in binary parser

ataymano:
Not sure about the right way of writing unit test here though.


external_parser/event_processors/joined_event.h
```diff
@@ -88,9 +88,8 @@ struct cb_joined_event : public typed_joined_event {
 
     int index = interaction_data.actions[0];
     auto action = interaction_data.actions[0];
-    auto probability = interaction_data.probabilities[0] *
-                       (1.f - interaction_data.probabilityOfDrop);
-    auto weight = 1.f - interaction_data.probabilityOfDrop;
+    auto probability = interaction_data.probabilities[0];
```
olgavrou, line 91: check if probabilityOfDrop is 1.f and if so maybe log and set weight to zero?

olgavrou:
If you want to add a unit test the way to go would be to generate a payload from example_gen (see [here](https://github.com/VowpalWabbit/reinforcement_learning/tree/master/external_parser/unit_tests/test_files)) with pdrop set to something, because by default I think it is zero, and then [here](https://github.com/VowpalWabbit/reinforcement_learning/blob/master/external_parser/unit_tests/test_example_joiner.cc) would be a good place to add a unit test that just checks for the example label's weight ([e.g.](https://github.com/VowpalWabbit/reinforcement_learning/blob/master/external_parser/unit_tests/test_example_joiner.cc#L56))

But this makes me think that it might be nice to somehow re-use example_gen for unit testing (not this PR, just a general thought)

external_parser/event_processors/joined_event.h
ataymano: hm, not 100% sure if setting weight to zero is safe. Also, dsjson parser is considering that kind of examples invalid and got skipped. Maybe it is better to replicate this behavior here for consistency and explore zero weight as next step?

olgavrou: Consistency is good, that would require `fill_in_label` to indicate back to joiner whether it should abort this example or not

ataymano:
@kumpera , @olgavrou , @cheng-tan 
working on some conflict (mostly in tests), but would love to hear your opinion on changes inside test_tools/log_parser.

I don't like the fact that I had to introduce one more way to generate joined logs, but the problem was that it is hard to generate logs with custom pdrop using existing approach (example_gen + joiner) since pdrop is not under our control in this case.

So I had to add missing pieces to log_gen.py instead.
The reason I put test data generation scrips inside jupyter notebook instead of enhancing log_gen command line arguments is that I am afraid that it can grow uncontrollably once we are going to create more nd more tests with more custom fields. Maybe we can start with focusing on programmatic approach to generate binary logs?

external_parser/event_processors/joined_event.h
cheng-tan: q: does this probability need to multiply by (1-pdrop) ?

external_parser/event_processors/joined_event.h
```diff
@@ -77,16 +77,22 @@ struct cb_joined_event : public typed_joined_event {
       VW::io::logger::log_warn(
           "distribution for event [{}] contains invalid probabilities",
           interaction_data.eventId);
+      return true; //why?
```
cheng-tan, line 80: this is a bit confusing.

external_parser/unit_tests/test_vw_external_parser.cc
```diff
@@ -359,4 +359,65 @@ BOOST_AUTO_TEST_CASE(cb_compare_dsjson_with_fb_models_deferred_actions_w_activat
   BOOST_CHECK_EQUAL_COLLECTIONS(buffer_fb_model.begin(), buffer_fb_model.end(),
                                 buffer_dsjson_model.begin(),
                                 buffer_dsjson_model.end());
+}
+
+BOOST_AUTO_TEST_CASE(cb_pdrop_05_parse) {
+  std::string input_files = get_test_files_location();
+
+  auto buffer =
+      read_file(input_files + "/valid_joined_logs/cb_joined_with_pdrop_05.fb");
```
cheng-tan, line 368: Can you document under `external_parser/unit_tests/test_files/README.md` how this file is generated?

external_parser/event_processors/joined_event.h
```diff
@@ -88,9 +88,8 @@ struct cb_joined_event : public typed_joined_event {
 
     int index = interaction_data.actions[0];
     auto action = interaction_data.actions[0];
-    auto probability = interaction_data.probabilities[0] *
-                       (1.f - interaction_data.probabilityOfDrop);
-    auto weight = 1.f - interaction_data.probabilityOfDrop;
+    auto probability = interaction_data.probabilities[0];
+    auto weight = 1.f / (1.f - interaction_data.probabilityOfDrop);
```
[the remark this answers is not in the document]
ataymano: resolved

external_parser/event_processors/joined_event.h
ataymano: thank you for catching it. reverted.

external_parser/unit_tests/test_vw_external_parser.cc
ataymano: done

external_parser/event_processors/joined_event.h
ataymano: Why do we need to take it into account both in probability and weight? 
Eventually all of them are going to the cost and in some cases (probably in ips) we should see weight / probability multiplier which probably can be mimicked by setting weight to 1 and multiplying probability by (1-pdrop), but not sure why we should do it this way.

test_tools/log_parser/log_gen.py
```diff
@@ -182,30 +197,67 @@ def mk_cb_payload():
...
+
+    builder = flatbuffers.Builder(0)
+    event_id = builder.CreateString(_event_id)
+
+    MetadataStart(builder)
+    MetadataAddId(builder, event_id)
+    MetadataAddClientTimeUtc(builder, mk_timestamp(builder))
+    MetadataAddPayloadType(builder, PayloadType.Outcome)
+    MetadataAddEncoding(builder, EventEncoding.Identity)
+    if _pdrop != 0:
```
kumpera, line 223: Why are you making it conditional here and not for decision?

test_tools/log_parser/parser.py
```diff
@@ -58,10 +58,14 @@ def cast(table, tmp_type):
...
 
 
-def parse_cb(payload):
+def parse_cb(payload, verbose):
     evt = CbEvent.GetRootAsCbEvent(payload, 0)
-    print(f'\tcb: actions:{evt.ActionIdsLength()} model:{evt.ModelId()} lm:{learning_mode_name(evt.LearningMode())} deferred:{evt.DeferredAction()}')
-    print(f'\t\tcontext: {fmt_payload(evt.ContextAsNumpy())}')
+    if not verbose:
+        print(f'\tcb: actions:{evt.ActionIdsLength()} model:{evt.ModelId()} lm:{learning_mode_name(evt.LearningMode())} deferred:{evt.DeferredAction()}')
+        print(f'\t\tcontext: {fmt_payload(evt.ContextAsNumpy())}')
```
kumpera, line 65: We probably don't want to print the context if !verbose, which is the biggest source of console spamming.

test_tools/log_parser/test_gen.ipynb
```diff
@@ -0,0 +1,77 @@
...
+   "source": [
+    "# Generate pdrop test data"
+   ],
+   "metadata": {}
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "source": [
+    "ctx1 = \"\"\"{\"_multi\":[{\"a1\":\"f1\"},{\"a2\":\"f2\"}]}\"\"\"\n",
```
kumpera, line 27: I guess we can't expect that the CLI vocabulary be rich enough to produce the sort of test files we need without making it a DSL. This approach makes more sense.
I don't understand  why did you pick a notebook, they are pretty hard to review compared to regular python files and in this case we don't care for any interactive elements, do we?

kumpera (changes requested):
I guess the library approach now makes sense and it's a good place to go.
We should later move the other files generated with log_converter to be scripted as well.

kumpera:
I don't think we can escape having a way to generate very specific test files that example_gen won't be able to, since it must go through rlclientlib.
My only thing here is the usage of a notebook.

test_tools/log_parser/test_gen.ipynb
ataymano: do you suggest just .py script?

ataymano: yeah I agree. replaced with python file - for now just naively converted from notebook, but I hope we can think later how exactly it should be written.

test_tools/log_parser/log_gen.py
ataymano: here is wrong, forgot to fix. thanks for catching.

external_parser/event_processors/joined_event.h
olgavrou: I think setting one or the other has an effect on reported average loss, comparing results with dsjson parser

olgavrou: looking at this 2 things need to be done here 
1. looks like for dsjson we set the weight for **every example** (see [here](https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/vowpalwabbit/parse_example_json.h#L1664) so we should do the same here too
2. @cheng-tan is right, we need to apply the pdrop to the probability too as it was before (that's how it is set during dsjson joining and before the dsjson format arrives to vw for parsing)

olgavrou: 1. could be done in a separate PR and should also be applied to ccb/cats

ataymano: 1. agree, good catch!
2. let's discuss offline

ataymano: 1. and agree, let me fix it in another PR - that one seems big enough

external_parser/joiners/example_joiner.cc
```diff
@@ -498,7 +498,11 @@ bool example_joiner::process_joined(v_array<example *> &examples) {
     return true;
   }
 
-  je->fill_in_label(examples);
+  if (!je->fill_in_label(examples)) {
+    _current_je_is_skip_learn = true;
```
olgavrou, line 502: Is this skip learn or is this a bug in the data that should be reported (i.e. if you don't set this to true then we will continue learning but this will get logged as an error [here](https://github.com/VowpalWabbit/reinforcement_learning/blob/master/external_parser/parse_example_binary.cc#L268)

ataymano: yeah you are right, it should be reported. let me fix that.

ataymano: fixed

olgavrou (approved):

external_parser/event_processors/joined_event.h
ataymano: 2. discussed offline - looks like the change was propagated from the legacy backend code which should be fixed.

PR #357, opened 2021-06-29
Update bin log spec

kumpera:
This update the spec to support magic and header messages at any position and files without them.


external_parser/parse_example_binary.cc
```diff
@@ -209,12 +172,15 @@ bool binary_parser::read_checkpoint_msg(io_buf *input) {
   _example_joiner->set_problem_type_config(
       checkpoint_info->problem_type_config());
 
+  _got_checkpoint_info = true;
   return true;
 }
 
 bool binary_parser::read_regular_msg(io_buf *input,
```
olgavrou, line 179: instead of passing a bool here could we potentially add a method `bool checkpoint_info_ready()` and call it after calling `read_regular_msg()` ?

kumpera: Good point. `i_joiner` should hold that info since it might have being externally configured prior to the construction of the parser object.

I rather keep the checking side of the code as is today so we entirely avoid parsing the examples in such case.

external_parser/parse_example_binary.cc
```diff
@@ -343,41 +299,47 @@ bool binary_parser::parse_examples(vw *all, v_array<example *> &examples) {
...
-
-  while (payload_type == MSG_TYPE_REGULAR) {
-    if (read_regular_msg(all->example_parser->input.get(), examples)) {
-      return true;
-    }
-
-    // bad payload process the next one
-    if (!advance_to_next_payload_type(all->example_parser->input.get(),
-                                      payload_type)) {
+  while(true) {
```
olgavrou, line 302: can this be `while(advance_to_next_payload_type())`?

external_parser/parse_example_binary.h
```diff
@@ -9,6 +9,7 @@
 
 constexpr size_t BINARY_PARSER_VERSION = 1;
 
+constexpr unsigned int MSG_TYPE_FILEMAGIC = 0x42465756; //'VWFB'
```
olgavrou, line 12: should update `REAME.md` with new file format expectancies

olgavrou (commented):
other than my 2 comments LGTM

external_parser/parse_example_binary.cc
kumpera: it can

external_parser/parse_example_binary.h
kumpera: OMG, I lost the update on the readme :(

kumpera: done

PR #358, opened 2021-06-29
[Binary parser] converter to skip learning

olgavrou:
Created a new converter parser that will be called by vw

The converter parser will in turn call the binary parser but not return to vw until all examples are read (and converted), and then will return false to vw so vw will just exit

if not called with `--quiet` then a run will look like:

```
Num weight bits = 18
learning rate = 0.5
initial_t = 0
power_t = 0.5
using no cache
Reading datafile = <datafile>
num sources = 1
Enabled reductions: gd, scorer, csoaa_ldf, cb_adf, cb_explore_adf_greedy, shared_feature_merger
average  since         example        example  current  current  current
loss     last          counter         weight    label  predict features

finished run
number of examples = 0
weighted example sum = 0.000000
weighted label sum = 0.000000
average loss = n.a.
total feature number = 0
```


external_parser/parse_example_converter.cc
```diff
@@ -0,0 +1,31 @@
...
+namespace external {
+
+binary_json_converter::binary_json_converter(std::unique_ptr<i_joiner> &&joiner)
+    : _parser(std::move(joiner)) {}
+
+binary_json_converter::~binary_json_converter() = default;
+
+bool binary_json_converter::parse_examples(vw *all,
+                                           v_array<example *> &examples) {
+  while (_parser.parse_examples(all, examples)) {
```
kumpera, line 19: Won't this array grown unbounded?

olgavrou: no it gets cleaned up, see [here](https://github.com/VowpalWabbit/reinforcement_learning/pull/358/files#diff-36a22962013a321a397c4ef1a1cb47d690b407e35e9bd49d144913dc64bad082R463)

kumpera (approved):

PR #361, opened 2021-07-06
[Binary parser] reward functions for ccb format

cheng-tan:
**Here are the use cases for ccb log reward calculation:**
1. interaction with slot index(numeric) + outcomes with slot index(numeric) -> join by slot index
2. interaction with slot id(string) + outcomes with slot id(string) -> join by slot id(string). If outcomes slot id(string) doesn't match, outcomes will be dangling.
3. interaction with slot id(string) + outcomes with slot index(numeric) -> join by slot index
4. interaction with slot index(numeric) + outcomes with slot id -> outcomes will be dangling.
5. interaction with slot id(string) + mixed outcome with slot index or slot id -> mapping outcome slot id to slot index, then join by slot index

**Note:**
1. In rl client lib, slot id in interaction event is not required, so case 4 is valid.
2. If there's no matched outcome events for a slot, we assign default reward for that slot. (same as current behavior)

**TODO:**
- [x] add warning for case 2 and case 4 if there are dangling outcomes
- [x] set default reward if there's no matched outcome events


external_parser/event_processors/joined_event.h
```diff
@@ -159,21 +160,37 @@ struct ccb_joined_event : public typed_joined_event {
   }
 
   void set_cost(v_array<example *> &examples, float reward,
-                size_t index = 0) const override {
-    if (examples.size() <= index) {
+                size_t slot_index = 0) const override {
+    if (examples.size() <= slot_index) {
```
ataymano, line 164: Do we still need this check? Looks like slot_index is incomparable with example.size() + we are doing proper comparison with example_index further.

external_parser/event_processors/joined_event.h
```diff
@@ -159,21 +160,37 @@ struct ccb_joined_event : public typed_joined_event {
...
-      VW::io::logger::log_warn(
-          "trying to set index [{}] on a CCB non-slot example", index);
-      return;
+
+    size_t index = 0;
+    for (auto &example : examples) {
+      if (example->l.conditional_contextual_bandit.type !=
+          CCB::example_type::slot) {
+        index++;
+        continue;
```
ataymano, line 176: Do we need it?

external_parser/event_processors/joined_event.h
```diff
@@ -159,21 +160,37 @@ struct ccb_joined_event : public typed_joined_event {
...
+      }
     }
-    examples[index]->l.conditional_contextual_bandit.outcome->cost =
+
+    size_t example_index = index + slot_index;
+    if (example_index >= examples.size()) {
+      VW::io::logger::log_error(
```
ataymano, line 182: Looks like we are just logging error but still continuing execution, and will probably have problems on line 186 during indexing.

external_parser/unit_tests/test_reward_functions.cc
```diff
@@ -38,95 +55,285 @@ float get_float_reward(std::string int_file_name, std::string obs_file_name, v2:
...
+    v2::LearningModeType_Online,
+    v2::ProblemType_CCB
+  );
+
+  BOOST_CHECK_EQUAL(rewards.size(), 2);
+  BOOST_CHECK_EQUAL(rewards.at(0), (2.0 + 4.0) / 2);
+  BOOST_CHECK_EQUAL(rewards.at(1), (2.0 + 2.0) / 2);
+}
+
+BOOST_AUTO_TEST_CASE(set_default_reward_if_slot_has_no_outcome_events)
```
cheng-tan, line 300: add another test if outcome events index is out of bound

external_parser/event_processors/joined_event.h
cheng-tan: yup, nice catch.

olgavrou (approved):

PR #366, opened 2021-07-15
[Binary parser] refactoring rewards

olgavrou:
- moved `metadata_info` into it's own file/namespace since it is being used by both interactions and observations
- moved `outcome_event`'s to `reward.h`
- moved reward calculations from `example_joiner` to `joined_event`s


cheng-tan (approved):
looks good.

PR #375, opened 2021-07-23
[Binary parser] don't log when skip learn, more tests, skip over unknown msg type

olgavrou:
- keep track of event's that are skip learn and don't log an error when we skip over them
- a bit of code cleanup
- added some unit tests (and their file generating changes in `example_gen` and `joiner`)
- if unknown message type just try to read the payload and continue

probably should have been 2 PRs but changes are small and mostly in the testing side


external_parser/parse_example_binary.cc
```diff
@@ -234,16 +259,21 @@ bool binary_parser::read_regular_msg(io_buf *input,
 
   _example_joiner->on_batch_read();
 
+  return process_next_in_batch(examples);
+}
+
+bool binary_parser::process_next_in_batch(v_array<example *> &examples) {
   while (_example_joiner->processing_batch()) {
```
olgavrou, line 266: this `while(_example_joiner->processing_batch())` loop was duplicated so refactored into `binary_parser::process_next_in_batch()`

cheng-tan (approved):

PR #380, opened 2021-07-28
example_gen enhancements

olgavrou:
- epsilon as cli argument
- loops output interactions and observations to files
- added ccb-baseline-loop
- added activations for ccb loops that have deferred events

### DIFF
--- a/test_tools/example_gen/example_gen.cc
+++ b/test_tools/example_gen/example_gen.cc
@@ -40,6 +40,7 @@ static const char *options[] = {
   "action-taken",
   "cb-loop",
   "ccb-loop",
+  "ccb-baseline-loop",
   nullptr
 };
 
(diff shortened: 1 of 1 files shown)


kumpera (approved):
I feel like we're stretching the capacity of example_gen as it was originally created.

It's still fine to have all this new capabilities on it, but we might want to make some of those more orthogonal.

For example: 
The way we control what to produce. In particular the _LOOP variants that generates both decisions and rewards.
If we allow controlling reward and decision generation separately in the CLI args this would not have being an issue. IE `example_gen --decision --reward f`.

Another example:
Data corruption is explicitly controlled as a generation mode and not as an extra flag. Having it that way would allow us to generate a bigger basket of bad examples.

olgavrou:
> I feel like we're stretching the capacity of example_gen as it was originally created.
> 
> It's still fine to have all this new capabilities on it, but we might want to make some of those more orthogonal.
> 
> For example:
> The way we control what to produce. In particular the _LOOP variants that generates both decisions and rewards.
> If we allow controlling reward and decision generation separately in the CLI args this would not have being an issue. IE `example_gen --decision --reward f`.
> 
> Another example:
> Data corruption is explicitly controlled as a generation mode and not as an extra flag. Having it that way would allow us to generate a bigger basket of bad examples.

You are right, this is me sort of abusing `example_gen` to do e2e testing where as I should probably have been using `rl.net` for that

PR #381, opened 2021-07-28
ccb log converter

cheng-tan:


external_parser/joiners/example_joiner.cc
```diff
@@ -433,9 +433,12 @@ bool example_joiner::process_joined(v_array<example *> &examples) {
       if (_vw->example_parser->metrics) {
         je->calculate_metrics(_vw->example_parser->metrics.get());
       }
-      if (_binary_to_json &&
-          je->interaction_metadata.payload_type == v2::PayloadType_CB) {
-        log_converter::build_cb_json(_outfile, *je);
+      if (_binary_to_json) {
+        if (je->interaction_metadata.payload_type == v2::PayloadType_CB) {
```
olgavrou, line 437: could this be:

```
if (_binary_to_json) {
    log_converter::build_json(_outfile, *je);
}
```

and then have `build_json()` check the payload type and call either `build_cb_json` or `build_ccb_json`

external_parser/log_converter.cc
```diff
@@ -123,8 +132,148 @@ void build_cb_json(std::ofstream &outfile,
...
     VW::io::logger::log_error(
-        "convert events: [{}] from binary to json format failed: [{}].",
+        "convert event: [{}] from binary to json format failed: [{}].",
         interaction_data.eventId, e.what());
   }
 }
+
+void build_ccb_json(std::ofstream &outfile,
+                    joined_event::joined_event &je) {
+  std::string event_id = je.interaction_metadata.event_id;
```
olgavrou, line 142: does this need to be a copy or can it be a reference? i.e. `std::string& event_id` or `const std::string& event_id`

olgavrou (approved):

external_parser/log_converter.cc
cheng-tan: replaced with reference

PR #382, opened 2021-07-29
[Binary parser] only convert timestamp to string when we persist metrics

olgavrou:
Moving the timestamp-to-string conversion at the end (i.e. when we have parsed all examples and are exiting and persisting the metrics) for perf reasons

### DIFF
--- a/external_parser/event_processors/joined_event.h
+++ b/external_parser/event_processors/joined_event.h
@@ -359,29 +359,5 @@ struct joined_event {
     typed_data->calc_and_set_cost(examples, default_reward, reward_function,
                                   interaction_metadata, outcome_events);
   }
-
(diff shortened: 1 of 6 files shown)


lalo:
Breaking changes in flatbuffers v2.0.0 that are being automatically pulled

olgavrou:
> Possibly related to breaking changes in flatbuffers v2.0.0 that are being automatically pulled. Not sure though.
> 
> [microsoft/vcpkg@eea00aa](https://github.com/microsoft/vcpkg/commit/eea00aaa919f720e2dfc8b86ebe1402a5c4679be)

Yes seems like it, all PR's fail on the windows CI due to the same issue. Looking into it

Issue #300, opened 2021-04-20
Flaky MacOS CI


olgavrou:
MacOS CI tests fail occasionally without much info, needs to be investigated and fixed

PR #385, opened 2021-07-30
Revert "mac ci: continue on error true (#327)"

lalo:
This reverts commit 763c4c78eb2ea5eecc6901638ac6431e83dbb07b.

Mac CI currently fails. Either we should fix or remove. This would turn back on the failing CI.

Related:
https://github.com/VowpalWabbit/reinforcement_learning/issues/300

Reverts
https://github.com/VowpalWabbit/reinforcement_learning/pull/327

### DIFF
--- a/.pipelines/build-macos.yml
+++ b/.pipelines/build-macos.yml
@@ -11,14 +11,10 @@ pool:
 steps:
 - checkout: self
   submodules: true
-  continueOnError: true
 - bash: .scripts/macos/restore.sh
   displayName: 'Restore dependencies'
-  continueOnError: true
 - bash: .scripts/macos/build.sh
   displayName: 'Build'
-  continueOnError: true
 - bash: .scripts/macos/test.sh
   displayName: 'Run tests'
-  continueOnError: true
 - task: PublishTestResults@2


kumpera:
I just ran into the issue, do you mind if I just fix it instead?

kumpera:
https://github.com/VowpalWabbit/reinforcement_learning/pull/386

kumpera:
Closing as we have a PR to fix the build instead.

lalo:
We technically still need to revert, to remove fail success in future breaks.

kumpera:
> 
> 
> We technically still need to revert, to remove fail success in future breaks.

I don't follow you what you mean. What happens with future breakage?

lalo:
If it breaks in the future again, the CI will pass bc it is set to "continueOnError: true". This reverts that, but this should be merged after your PR.

kumpera:
@lalo I thought this PR was reverting the change the broke mac CI in the first place.

Why did we disable it to begin with?

Disabled CI lanes should be a temporary situation and treated as an urgent matter. They should be deleted otherwise.

olgavrou:
@lalo @kumpera this seems to fix the mac ci: https://github.com/VowpalWabbit/reinforcement_learning/pull/390 so we can then merge this PR too

PR #388, opened 2021-07-30
Switch to SAX parsing get the perf we need.

kumpera:
This applies 3 optimizations:

- don't parse examples
- switch to SAX serializer
- emit context as a raw string avoiding parsing it

The result is:
Initial: 52.4Mbps
This PR: 147Mbps

This PR introduces a more significant improvement in user time, which will be more salient in oversubscribed environments


kumpera (commented):
Remove dead code.
Remove usage EndObject/ EndArray that counts number of entries

external_parser/joiners/example_joiner.cc
```diff
@@ -482,6 +484,10 @@ bool example_joiner::process_joined(v_array<example *> &examples) {
     return false;
   }
 
+  if(_binary_to_json) {
```
olgavrou, line 487: do we now need to remove lines 502 to 505 which also check `_binary_to_json`?

olgavrou: actually I am confused, without the calc_and_set_reward, how are we calculating the reward?

kumpera: I thought the calculation was done on log_converted as it works for original_reward.
The issue is that `calc_and_set_reward` needs to be split in two, one o calc and one to set. The calc part is called always and set only when !binary_to_json.

Working on it

kumpera: split done. it now works as suggested.

olgavrou (approved):

cheng-tan (approved):

PR #389, opened 2021-08-02
try to use a specific vm image

olgavrou:


olgavrou:
looks like we can only use the latest windows `vmImage`

PR #392, opened 2021-08-04
Activations in multistep: first PR with schema changes only

ataymano:


CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 cmake_minimum_required(VERSION 3.5)
+set(CMAKE_POLICY_DEFAULT_CMP0074 NEW)
```
kumpera, line 2: I could not find anything on google about this, what does it do?

kumpera: The documentation on it is here: https://cmake.org/cmake/help/latest/variable/CMAKE_POLICY_DEFAULT_CMPNNNN.html#variable:CMAKE_POLICY_DEFAULT_CMP%3CNNNN%3E

It says: `This variable should not be set by a project in CMake code; use cmake_policy(SET) instead. `

kumpera: Furthermore, this policy was introduced in CMake 3.12. If we depend on this behavior shouldn't we bump the minimum version instead?
CC @jackgerrits

jackgerrits: Yes we should use `cmake_policy`. If we depend on this in the build, then yeah we should bump the version requirement to match it.

PR #396, opened 2021-08-05
[binary_parser] Bug fix: Add pdrop field for cb log converter

cheng-tan:


kumpera (approved):
This looks good, but it could use a test case.

cheng-tan:
> This looks good, but it could use a test case.

yes. I'll merge this pr for now and add a test case once pr #351 is merged.

PR #399, opened 2021-08-09
feat: Add skip_learn and metrics tests for cats binary parser

bassmang:


external_parser/unit_tests/test_metrics.cc
```diff
@@ -228,4 +228,35 @@ BOOST_AUTO_TEST_CASE(
...
+    check_metrics_ca_mixed_deferred_action_events) {
+  std::string infile_name =
+      "valid_joined_logs/ca_mixed_deferred_action_events_20.log";
+  std::string outfile_name =
+      "ca_mixed_deferred_action_events.json";
+
+  std::map<std::string, int> expected_int_metrics = {
+      {"total_predict_calls", 0},
+      {"total_learn_calls", 10},
+      {"total_log_calls", 233},
```
olgavrou, line 243: I'd remove the `total_log_calls` because the test might then break if we add/remove log messages and it will be unrelated with what we are testing

olgavrou (approved):

PR #400, opened 2021-08-10
feat: cats log converter

bassmang:


external_parser/unit_tests/test_log_converter.cc
```diff
@@ -134,3 +138,19 @@ BOOST_AUTO_TEST_CASE(ccb_payload_with_slot_id) {
 }
 
 BOOST_AUTO_TEST_SUITE_END()
+
+BOOST_AUTO_TEST_SUITE(log_converter_ca_format)
+BOOST_AUTO_TEST_CASE(ca_deferred_action_without_activation) {
+  std::string infile_path = "skip_learn/ca/deferred_action_without_activation.fb";
```
cheng-tan, line 144: q: is this joined log has skip learn enabled? The dsjson schema doesn't include a "_skip_learn" field like the cb and ccb schema, which affects cfe.

bassmang: Yes the joined log was generated with `--deferred_action` for the interaction and without `action_taken` in the observation, so skip_learn should be enabled.  Perhaps we should add a _skip_learn field to the dsjson schema for ca. The CaEvent schema includes `deferred_action` and is/set_skip_learn() is defined for all joined events. @olgavrou do you have thoughts?

olgavrou (commented):
There is actually more that we can get from the cat's json format (see the unit tests [here](https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/test/unit_test/cats_user_provided_pdf.cc)) when using `--first_only` so we should take it into account during conversion. This can be done here or in a separate PR (I just realised the input format page in the vw wiki isn't update for this, adding it now)

olgavrou:
> There is actually more that we can get from the cat's json format (see the unit tests [here](https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/test/unit_test/cats_user_provided_pdf.cc)) when using `--first_only` so we should take it into account during conversion. This can be done here or in a separate PR (I just realised the input format page in the vw wiki isn't update for this, adding it now)

This is actually part of the context so it should be included already if needed, we don't need to do anything extra

olgavrou (approved):

PR #404, opened 2021-08-16
Slates parsing and reward calculation

cheng-tan:


external_parser/event_processors/joined_event.h
```diff
@@ -158,27 +158,29 @@ struct cb_joined_event : public typed_joined_event {
...
+  std::map<std::string, int> slot_id_to_index_map;
   std::vector<float> rewards;
   std::vector<float> original_rewards;
   std::map<int, std::vector<reward::outcome_event>> outcomes_map;
 
-  bool skip_learn;
-  float probability_of_drop{0.f};
-
   ~ccb_joined_event() = default;
   // TODO fill in
```
olgavrou, line 176: nit: remove the TODO since it's done

external_parser/event_processors/joined_event.h
```diff
@@ -324,6 +326,77 @@ struct ccb_joined_event : public typed_joined_event {
...
+        // process next slot from interaction_data vector
+        slot_index++;
+      }
+    }
+    return true;
+  }
+
+  void set_cost(v_array<example *> &examples, float reward,
+                size_t slot_offset = 0) const override {
+    if (examples[0]->l.slates.type ==  VW::slates::example_type::shared) {
```
olgavrou, line 376: so if it isn't shared that means we won't set a cost but we will have set the probabilities and actions?

external_parser/unit_tests/test_vw_external_parser.cc
```diff
@@ -87,6 +87,69 @@ BOOST_AUTO_TEST_CASE(ccb_simple) {
...
+    BOOST_CHECK_EQUAL(examples[8]->indices.size(), 0); // newline example
+
+    multi_ex multi_exs;
+    for (auto *ex : examples) {
+      multi_exs.push_back(ex);
+    }
+
+    vw->learn(multi_exs);
+
+    examples[0]->pred.decision_scores = {
```
olgavrou, line 134: nice! maybe add a comment saying that the slates clear example expects the pred to be set

external_parser/event_processors/joined_event.h
```diff
@@ -324,6 +326,77 @@ struct ccb_joined_event : public typed_joined_event {
...
+
+  bool is_skip_learn() const override { return multi_slot_interaction.skip_learn; }
+  void set_skip_learn(bool sl) override { multi_slot_interaction.skip_learn = sl; }
+
+  void set_apprentice_reward() override {}
+
+  bool fill_in_label(v_array<example *> &examples) const override {
+    size_t slot_index = 0;
+
+    for (auto *ex : examples) {
```
olgavrou, line 344: Do we also need to fill the weight for every example, same as we do on ccb fill_in_label?

external_parser/event_processors/joined_event.h
olgavrou: I think setting this to `examples[0]` without checking if it is `shared` is ok based on [here](https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/vowpalwabbit/parse_slates_example_json.h#L158) and [here](https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/vowpalwabbit/parse_slates_example_json.h#L246)

external_parser/joiners/multistep_example_joiner.cc
```diff
@@ -226,8 +220,8 @@ bool multistep_example_joiner::process_joined(v_array<example *> &examples) {
     return false;
   }
 
-  const auto reward = _reward_calculation(joined.outcome_events);
-  try_set_label(joined, reward, examples);
+  joined.calc_reward(_loop_info.default_reward, _reward_calculation.value());
```
olgavrou, line 223: flipping `fill_in_label` and `calc_reward` like that makes the code much cleaner <3

olgavrou (approved):

external_parser/event_processors/joined_event.h
```diff
@@ -374,19 +366,10 @@ struct ca_joined_event : public typed_joined_event {
...
-  void set_cost(v_array<example *> &examples, float reward,
-          size_t index = 0) const override {
-    if (std::isnan(interaction_data.action)) {
-      return;
-    }
-
-    examples[index]->l.cb_cont.costs[0].cost = -1.f * reward;
-  }
-
   bool should_calculate_reward(
```
olgavrou, line 373: so all the `should_calculate_reward`'s functionality has been moved to the reward functions, except for cats?

cheng-tan: forgot to remove this one, `should_calculate_reward` functionality should be handled by reward functions for all problem types.

PR #405, opened 2021-08-18
Benchmarks and github action to run them in CI

olgavrou:


.github/workflows/run_benchmarks.yml
```diff
@@ -0,0 +1,61 @@
...
+        uses: actions/upload-artifact@v2
+        with:
+          name: benchmark-compare
+          path: benchmark/tools/
+      - name: Install benchmark compare requirements
+        shell: bash
+        run: /usr/local/miniconda/envs/test-python36/bin/pip install -r benchmark/tools/requirements.txt
+      - run: rm -rf benchmark build rlclientlib/generated/ # generated or downloaded files
+      - uses: actions/checkout@v1
+        # temp for first build
```
olgavrou, line 42: lines 43 and 44 will run the branch against the branch since the benchamarks do not exist in master, should be removed before merging so that future branches run against master

benchmarks/README.md
```diff
@@ -0,0 +1,15 @@
+install google benchmark: https://github.com/google/benchmark
+
+by default google benchmark is built in Debug mode so you might want to specify Release mode when building/installing
+
+build rl benchmarks:
+
+```
+make -j rl_benchmarks
```
bassmang, line 8:
```suggestion
make -j $(nproc) rl_benchmarks
```

benchmarks/benchmarks_common.h
```diff
@@ -0,0 +1,101 @@
...
+
+class cb_decision_gen {
+  int shared_features, action_features, actions_per_decision;
+  std::vector<std::string> actions_set;
+  prng rand;
+  std::string temp_str;
+  bool passthrough;
+
+  std::string mk_feature_vector(int count, uint32_t max_idx) {
+    std::stringstream str;
```
bassmang, line 34:
```suggestion
    std::ostringstream str;
```

benchmarks/benchmarks_common.h
```diff
@@ -0,0 +1,101 @@
...
+      : shared_features(shared_features), action_features(action_features),
+        actions_per_decision(actions_per_decision), rand(initial_seed),
+        passthrough(passthrough) {
+    for (int i = 0; i < total_actions; ++i) {
+      actions_set.push_back(
+          mk_feature_vector(action_features, action_features * 3));
+    }
+  }
+  const char *gen_example() {
+    std::stringstream str;
```
bassmang, line 66:
```suggestion
    std::ostringstream str;
```

benchmarks/benchmarks_common.h
```diff
@@ -0,0 +1,101 @@
...
+#include <string>
+#include <vector>
+
+// We use this instead of rand to ensure it's xplat
+int pseudo_random(int seed) {
+  constexpr uint64_t CONSTANT_A = 0xeece66d5deece66dULL;
+  constexpr uint64_t CONSTANT_C = 2147483647;
+
+  uint64_t val = CONSTANT_A * seed + CONSTANT_C;
+  return (int)(val & 0xFFFFFFFF);
```
bassmang, line 11:
```suggestion
  return static_cast<int>(val & 0xFFFFFFFF);
```

bassmang (approved):

benchmarks/benchmark_cb_v2.cc
```diff
@@ -0,0 +1,115 @@
...
+  }
+}
+
+// x shared features
+// x features per action
+// x actions per example
+// x actions in total
+// example count
+// compression
+// dedup
```
bassmang, line 89: It could be helpful to be more specific about each input, ie `example count` -> `Total number of examples`

.scripts/linux/install-benchmarks.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+set -e
+set -x
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+REPO_DIR=$SCRIPT_DIR/../../
+cd $REPO_DIR
+
+# Check out the library.
+git clone https://github.com/google/benchmark.git
```
kumpera, line 10: Should we incorporate this step into the docker image we use?

benchmarks/benchmark_cb_v2.cc
```diff
@@ -0,0 +1,115 @@
...
+BENCHMARK_CAPTURE(bench_cb, non_dedupable_payload, 20, 10, 50, 2000, 500, false,
```
kumpera, line 90: Are we capturing the cost of cb_decision_gen construction into the benchmark?

benchmarks/benchmarks_common.h
```diff
@@ -0,0 +1,101 @@
...
+        ++added;
+        added_idx.insert(idx);
+      }
+    }
+    str << "}";
+    return str.str();
+  }
+
+public:
+  cb_decision_gen(int shared_features, int action_features,
```
kumpera, line 54: In general, only templates or small inline methods should be in headers.

benchmarks/benchmark_cb_v2.cc
kumpera: If we don't move the unique action construction to outside the benchmarks, won't we favor the heavy dedup case?

kumpera (changes requested):

benchmarks/benchmarks_common.h
olgavrou: true

benchmarks/benchmark_cb_v2.cc
olgavrou: The benchmark captures the cost of generating the next example you are right. We could generate the `count` examples outside of the benchmark and then just use them in the rank call

.github/workflows/run_benchmarks.yml
```diff
@@ -0,0 +1,61 @@
+name: Run Benchmarks
```
jackgerrits, line 1: Eventually it would probably be useful to factor out this action into a repo for easier sharing: https://docs.github.com/en/actions/learn-github-actions/sharing-workflows-with-your-organization

thats for later though

benchmarks/benchmark_cb_v2.cc
olgavrou: fixed

benchmarks/benchmark_cb_v2.cc
olgavrou: added more explanation let me know if you think it's still unclear

benchmarks/README.md
olgavrou: done

benchmarks/benchmarks_common.h
olgavrou: split files

.scripts/linux/install-benchmarks.sh
olgavrou: Depends if we view the docker image as a CI tool or a vw distribution mechanism

benchmarks/benchmarks_common.h
olgavrou: done

benchmarks/benchmarks_common.h
olgavrou: done

PR #410, opened 2021-08-25
set policy if exists in cmake version available

olgavrou:


CMakeLists.txt
```diff
@@ -1,5 +1,7 @@
 cmake_minimum_required(VERSION 3.5)
-cmake_policy(SET CMP0074 NEW)
+if(POLICY CMP0074)
```
jackgerrits, line 2: Does the build script correctly function if this policy is not available? It may be better to just update the required version to 3.12

olgavrou: I believe it does, this policy is about how cmake discovers packages afaik

If you build with a newer version of cmake it still functions properly, it just wants the new policy to be set to either `NEW` or `OLD`

cheng-tan (approved):

PR #414, opened 2021-09-07
Add apim authorization

slahabar:
Adds ability for users of rlclient lib to authenticate to apim using the instance url  and the instance key. 

With move to multitenancy (with shared eventhub) we are introducing two new apis /logs/interactions and /logs/observations to directly log the binary payload to the shared eventhub through apim. 

The existing functionality of logging directly to a event hub should not be affected if the users are still using event hub connection string.

[Front end proxy](https://microsoft.sharepoint.com/:w:/r/teams/personalizr/_layouts/15/doc.aspx?sourcedoc=%7B36e8f5b9-99c3-4fa7-b933-20833f05af32%7D&action=edit&share=IQG59eg2w5mnT7kzIIM_Ba8yAaS3aScvU07fDc7NZIYrZZA&cid=af40fc7a-bb79-4a74-b4de-579d7902c6aa)


slahabar:
Fixing build...

include/constants.h
```diff
@@ -13,6 +13,7 @@ namespace reinforcement_learning {  namespace name {
       const char *const  INITIAL_EPSILON         = "initial_exploration.epsilon";
       const char *const  LEARNING_MODE           = "rank.learning.mode";
       const char* const  PROTOCOL_VERSION             = "protocol.version";
+      const char* const  API_KEY                 = "api.key";
```
johnhuang01, line 16: test

susanhuhu:
Is it expected that interaction and observation will go through different APIM hosts?

rlclientlib/utility/eventhub_http_authorization.h
```diff
@@ -11,17 +12,18 @@ namespace reinforcement_learning {
...
+    ~eventhub_http_authorization() = default;
     
-    int init(api_status* status);
-    int get(std::string& authorization, api_status* status);
+    virtual int init(api_status* status) override;
+    virtual int get_http_headers(http_headers& headers, api_status* status) override;
 
   private:
     int check_authorization_validity_generate_if_needed(api_status* status);
+    int get(std::string& authorization, api_status* status);
```
tyclintw, line 26: `get` is not a descriptive enough function name. Please elaborate to describe the functionality.

rlclientlib/utility/apim_http_authorization.cc
```diff
@@ -0,0 +1,18 @@
...
+    const std::string& api_key)
+    : _api_key(api_key) {
+  }
+
+  int apim_http_authorization::init(api_status* status) {
+    return error_code::success;
+  }
+
+  int apim_http_authorization::get_http_headers(http_headers& headers, api_status* status) {
+    headers.add(_XPLATSTR("apim-subscription-id"), _api_key.c_str());
```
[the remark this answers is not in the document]
tyclintw: We'll need support for bearer tokens as well.

tyclintw (changes requested):

rlclientlib/azure_factories.cc
```diff
@@ -15,11 +18,15 @@ namespace reinforcement_learning {
   int observation_sender_create(i_sender** retval, const u::configuration&, error_callback_fn*, i_trace* trace_logger, api_status* status);
   int interaction_sender_create(i_sender** retval, const u::configuration&, error_callback_fn*, i_trace* trace_logger, api_status* status);
   int decision_sender_create(i_sender** retval, const u::configuration&, error_callback_fn*, i_trace* trace_logger, api_status* status);
-
+  int observation_api_sender_create(i_sender** retval, const u::configuration& cfg, error_callback_fn* error_cb, i_trace* trace_logger, api_status* status);
```
orenmichaely, line 21: I think it would be helpful to add a comment explaining the difference between interaction/observation_sender_create and interaction/observation_api_sender_create

rlclientlib/azure_factories.cc
```diff
@@ -40,41 +47,97 @@ namespace reinforcement_learning {
     return url;
   }
 
+  int observation_api_sender_create(i_sender** retval, const u::configuration& cfg, error_callback_fn* error_cb, i_trace* trace_logger, api_status* status) {
```
homezcx, line 50: nit: it seems to be a lot of shared code except OBSERVATION_APIM_TASKS_LIMIT, OBSERVATION_APIM_MAX_HTTP_RETRIES, 
INTERACTION_APIM_TASKS_LIMIT,
INTERACTION_APIM_MAX_HTTP_RETRIES

homezcx: also for the method observation_sender_create, interaction_sender_create

rlclientlib/logger/http_transport_client.cc
```diff
@@ -58,7 +55,7 @@ namespace reinforcement_learning {
       }
 
       // If the response is not the expected code then it has failed. Retry if possible otherwise report background error.
-      if(code != status_codes::Created) {
+      if(code != status_codes::Created && code != status_codes::NoContent) {
```
homezcx, line 58: is NoContent considered as success?

slahabar: Yes
HTTP Status 204 (No Content) indicates that the server has successfully fulfilled the request and that there is no content to send in the response payload body.

rlclientlib/azure_factories.cc
slahabar: Added comments to these functions.

slahabar:
@susanhuhu - Regarding your comment 
Is it expected that interaction and observation will go through different APIM hosts? Not at the moment. 
However, keeping it separate would make testing with different versions easier. This was the ask during design review with Rajan and Alexey.

rlclientlib/logger/http_transport_client.h
```diff
@@ -15,18 +15,18 @@
 
 namespace reinforcement_learning {
   class i_trace;
+  class i_authorization;
 
   // The eventhub_client send string data in POST requests to an HTTP endpoint.
```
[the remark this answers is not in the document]
slahabar: Fixed.

rlclientlib/rlclientlib.vcxproj
```diff
@@ -292,17 +292,20 @@ $(flatcPath) -o "$(SolutionDir)rlclientlib\generated\v2" --cpp "$(SolutionDir)rl
...
     <ClInclude Include="model_mgmt\restapi_data_transport.h" />
     <ClInclude Include="utility\http_client.h" />
     <ClInclude Include="utility\http_helper.h" />
-    <ClInclude Include="utility\http_authorization.h" />
+    <ClInclude Include="utility\authorization.h" />
+    <ClInclude Include="utility\eventhub_http_authorization.h" />
+    <ClInclude Include="utility\apim_http_authorization.h" />
     <ClCompile Include="azure_factories.cc" />
-    <ClCompile Include="logger\eventhub_client.cc" />
+    <ClInclude Include="logger\http_transport_client.cc" />
```
ataymano, line 303: ClInclude -> ClCompile

slahabar: Fixed

rlclientlib/utility/apim_http_authorization.cc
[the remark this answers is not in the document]
slahabar: 1) Fixed the key name. 

2) We should be able to support bearer token as well by implementing the same interface that accepts token endpoint, apikey and returns the Authorization header with "Bearer AuthToken". 

E.g.,
curl -v -X POST \
"https://YOUR-REGION.api.cognitive.microsoft.com/sts/v1.0/issueToken" \
-H "Content-type: application/x-www-form-urlencoded" \
-H "Content-length: 0" \
-H "Ocp-Apim-Subscription-Key: YOUR_SUBSCRIPTION_KEY"

curl -X POST 'https://api.cognitive.microsofttranslator.com/translate?api-version=3.0&from=en&to=de' \
-H 'Authorization: Bearer YOUR_AUTH_TOKEN' \
-H 'Content-Type: application/json' \
--data-raw '[{ "text": "How much for the cup of coffee?" }]' | json_pp

The implementation would be similar to event hub authorization headers, where we cache the authorization token and refresh if it has expired.

rlclientlib/utility/eventhub_http_authorization.h
slahabar: Fixed.

rlclientlib/azure_factories.cc
slahabar: Refactored the interaction/observation api sender code I added in this PR.

include/constants.h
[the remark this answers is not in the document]
slahabar: fixed

include/constants.h
```diff
@@ -28,6 +29,9 @@ namespace reinforcement_learning {  namespace name {
       const char *const  INTERACTION_USE_COMPRESSION = "interaction.send.use_compression";
       const char *const  INTERACTION_USE_DEDUP = "interaction.send.use_dedup";
       const char *const  INTERACTION_QUEUE_MODE = "interaction.queue.mode";
+      const char *const  INTERACTION_API_HOST = "interaction.api.host";
```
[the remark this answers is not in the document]
slahabar: fixed

include/constants.h
```diff
@@ -42,6 +46,9 @@ namespace reinforcement_learning {  namespace name {
       const char *const  OBSERVATION_SENDER_IMPLEMENTATION    = "observation.sender.implementation";
       const char *const  OBSERVATION_USE_COMPRESSION = "observation.send.use_compression";
       const char *const  OBSERVATION_QUEUE_MODE = "observation.queue.mode";
+      const char *const  OBSERVATION_API_HOST = "observation.api.host";
```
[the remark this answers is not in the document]
slahabar: fixed

include/constants.h
```diff
@@ -75,6 +82,8 @@ namespace reinforcement_learning {  namespace value {
       const char *const INTERACTION_EH_SENDER = "INTERACTION_EH_SENDER";
       const char *const OBSERVATION_FILE_SENDER = "OBSERVATION_FILE_SENDER";
       const char *const INTERACTION_FILE_SENDER = "INTERACTION_FILE_SENDER";
+      const char* const OBSERVATION_API_SENDER = "OBSERVATION_API_SENDER";
+      const char* const INTERACTION_API_SENDER = "INTERACTION_API_SENDER";
```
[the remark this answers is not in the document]
slahabar: fixed

rlclientlib/azure_factories.cc
```diff
@@ -40,41 +47,89 @@ namespace reinforcement_learning {
...
+  int create_apim_http_api_sender(i_sender** retval, const u::configuration& cfg, const char* api_host, int tasks_limit, int max_http_retries, error_callback_fn* error_cb, i_trace* trace_logger, api_status* status)
+  {
+    std::string url;
+    url.append(api_host);
+
+    i_http_client* client;
+    RETURN_IF_FAIL(create_http_client(url.c_str(), cfg, &client, status));
+
+    i_authorization* authorization = new apim_http_authorization(cfg.get(name::API_KEY, "dummykey"));
+    *retval = new http_transport_client(
```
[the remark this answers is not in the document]
slahabar: Updated to use templates. I am getting below warnings and was wondering what is the best practice for placing template declaration and definition in different files. @rajan-chari 

3>e:\repo\upstream_rl\rlclientlib\logger\http_transport_client.h(58): note: see declaration of 'reinforcement_learning::http_transport_client<reinforcement_learning::apim_http_authorization>::http_request_task::send_request'
3>e:\repo\upstream_rl\rlclientlib\logger\http_transport_client.h(94): warning C4661: 'reinforcement_learning::http_transport_client<reinforcement_learning::eventhub_http_authorization>::http_request_task::http_request_task(reinforcement_learning::i_http_client *,web::http::http_headers,const reinforcement_learning::http_transport_client<reinforcement_learning::eventhub_http_authorization>::http_request_task::buffer &,size_t,reinforcement_learning::error_callback_fn *,reinforcement_learning::i_trace *)': no suitable definition provided for explicit template instantiation request
3>e:\repo\upstream_rl\rlclientlib\logger\http_transport_client.h(40): note: see declaration of 'reinforcement_learning::http_transport_client<reinforcement_learning::eventhub_http_authorization>::http_request_task::http_request_task'
3>e:\repo\upstream_rl\rlclientlib\logger\http_transport_client.h(94): warning C4661: 'int reinforcement_learning::http_transport_client<reinforcement_learning::eventhub_http_authorization>::http_request_task::join(void)': no suitable definition provided for explicit template instantiation request
3>e:\repo\upstream_rl\rlclientlib\logger\http_transport_client.h(56): note: see declaration of 'reinforcement_learning::http_transport_client<reinforcement_learning::eventhub_http_authorization>::http_request_task::join'
3>e:\repo\upstream_rl\rlclientlib\logger\http_transport_client.h(94): warning C4661: 'Concurrency::task<web::http::status_code> reinforcement_learning::http_transport_client<reinforcement_learning::eventhub_http_authorization>::http_request_task::send_request(size_t)': no suitable definition provided for explicit template instantiation request
3>e:\repo\upstream_rl\rlclientlib\logger\http_transport_client.h(58): note: see declaration of 'reinforcement_learning::http_transport_client<reinforcement_learning::eventhub_http_authorization>::http_request_task::send_request'

rlclientlib/logger/http_transport_client.h
```diff
@@ -90,4 +89,7 @@ namespace reinforcement_learning {
     i_trace* _trace;
     error_callback_fn* _error_callback;
   };
+
+  template class http_transport_client<apim_http_authorization>;
```
rajan-chari, line 93: This does not seem necessary here.  Any reason for this?

rlclientlib/logger/http_transport_client.h
```diff
@@ -90,4 +89,7 @@ namespace reinforcement_learning {
...
+  template class http_transport_client<eventhub_http_authorization>;
```
rajan-chari, line 94: Does not seem necessary

rajan-chari (commented):
@slahabar , the following two lines at the end of http_transport_client.h are not needed.  Once removed the warnings go away for me.

  template class http_transport_client<apim_http_authorization>;
  template class http_transport_client<eventhub_http_authorization>;

rlclientlib/logger/http_transport_client.h
slahabar: image

slahabar: If I remove these definitions and build rl_sim_cpp I get linking errors. Google search suggested this. Not sure if there is a better way. rlclientlib builds fine. @rajan-chari

slahabar: 1>Done building project "rlclientlib.vcxproj".
2>------ Build started: Project: rl_sim_cpp, Configuration: Debug x64 ------
2>rlclientlib.lib(azure_factories.obj) : error LNK2019: unresolved external symbol "public: __cdecl reinforcement_learning::http_transport_client<class reinforcement_learning::apim_http_authorization>::http_transport_client<class reinforcement_learning::apim_http_authorization>(class reinforcement_learning::i_http_client *,unsigned __int64,unsigned __int64,class reinforcement_learning::i_trace *,class reinforcement_learning::error_callback_fn *)" (??0?$http_transport_client@Vapim_http_authorization@reinforcement_learning@@@reinforcement_learning@@QEAA@PEAVi_http_client@1@_K1PEAVi_trace@1@PEAVerror_callback_fn@1@@Z) referenced in function "int __cdecl reinforcement_learning::create_apim_http_api_sender(class reinforcement_learning::i_sender * *,class reinforcement_learning::utility::configuration const &,char const *,int,int,class reinforcement_learning::error_callback_fn *,class reinforcement_learning::i_trace *,class reinforcement_learning::api_status *)" (?create_apim_http_api_sender@reinforcement_learning@@YAHPEAPEAVi_sender@1@AEBVconfiguration@utility@1@PEBDHHPEAVerror_callback_fn@1@PEAVi_trace@1@PEAVapi_status@1@@Z)
2>rlclientlib.lib(azure_factories.obj) : error LNK2019: unresolved external symbol "public: __cdecl reinforcement_learning::http_transport_client<class reinforcement_learning::eventhub_http_authorization>::http_transport_client<class reinforcement_learning::eventhub_http_authorization>(class reinforcement_learning::i_http_client *,unsigned __int64,unsigned __int64,class reinforcement_learning::i_trace *,class reinforcement_learning::error_callback_fn *)" (??0?$http_transport_client@Veventhub_http_authorization@reinforcement_learning@@@reinforcement_learning@@QEAA@PEAVi_http_client@1@_K1PEAVi_trace@1@PEAVerror_callback_fn@1@@Z) referenced in function "int __cdecl reinforcement_learning::interaction_sender_create(class reinforcement_learning::i_sender * *,class reinforcement_learning::utility::configuration const &,class reinforcement_learning::error_callback_fn *,class reinforcement_learning::i_trace *,class reinforcement_learning::api_status *)" (?interaction_sender_create@reinforcement_learning@@YAHPEAPEAVi_sender@1@AEBVconfiguration@utility@1@PEAVerror_callback_fn@1@PEAVi_trace@1@PEAVapi_status@1@@Z)
2>E:\Repo\upstream_rl\x64\Debug\rl_sim_cpp.exe : fatal error LNK1120: 2 unresolved externals
2>Done building project "rl_sim_cpp.vcxproj" -- FAILED.

.scripts/macos/restore.sh
```diff
@@ -3,7 +3,5 @@
 set -e
 set -x
 
-brew update
```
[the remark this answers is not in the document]
slahabar: Update step is failing the Mac OS build. @jackgerrits did the same for vw. Got the lead from him to remove this step.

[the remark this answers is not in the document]
jackgerrits: This should be resolved in master and this PR updated rather than including the changes also here to reduce noise.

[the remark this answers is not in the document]
slahabar: Thanks. will remove it.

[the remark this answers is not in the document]
jackgerrits: Merged in master

rlclientlib/azure_factories.cc
```diff
@@ -40,41 +46,73 @@ namespace reinforcement_learning {
     return url;
   }
 
+  int create_apim_http_api_sender(i_sender** retval, const u::configuration& cfg, const char* api_host, int tasks_limit, int max_http_retries, error_callback_fn* error_cb, i_trace* trace_logger, api_status* status)
+  {
+    std::string url;
```
ataymano, line 51: why do we need this?

rlclientlib/utility/authorization.h
```diff
@@ -0,0 +1,15 @@
+#pragma once
+
+#include <cpprest/http_headers.h>
+
+using namespace web::http;
+
+namespace reinforcement_learning {
+  class api_status;
+  class i_authorization {
```
ataymano, line 9: why do we need this? Looks like there is no usage/children of this class. Is eventhub_http_authorization supposed to be inherited from here?

slahabar: Not anymore. I deleted the reference from the project file. Missed deleting the file. Deleted now.

rlclientlib/logger/http_transport_client.h
ataymano: I believe we need to move http_transport_client.cc to http_transport_client.h in order to fix that.
It will mess the diff though. Maybe we can leave these lines for now and create another refactoring PR (http_transport_client.h -> http_transport_client.cc + instantiation removal) immediately after that.
@slahabar , @rajan-chari , what do you think?

rlclientlib/azure_factories.cc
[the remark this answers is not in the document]
slahabar: got rid of the warnings by moving template class http_transport_client<apim_http_authorization>;
template class http_transport_client<eventhub_http_authorization>; to the .cc file.

ataymano (approved):

rlclientlib/logger/http_transport_client.h
rajan-chari: Sounds good @ataymano , @slahabar

include/constants.h
```diff
@@ -13,6 +15,14 @@ namespace reinforcement_learning {  namespace name {
       const char *const  INITIAL_EPSILON         = "initial_exploration.epsilon";
       const char *const  LEARNING_MODE           = "rank.learning.mode";
       const char* const  PROTOCOL_VERSION             = "protocol.version";
+      const char* const  HTTP_API_KEY            = "http.api.key";
+      const char *const  INTERACTION             = "interaction";
+      const char *const  OBSERVATION             = "observation";
+      const char *const  CONFIG_SECTION          = "config.section";
+      const char *const  EH_HOST                 = ".eventhub.host";
```
jackgerrits, line 22: I don't understand the change here? I was under the impression the idea of `reinforcement_learning::name` was the each constant mapped 1:1 with keys that could exist in the configuration JSON? Can you explain the intention here?

rlclientlib/azure_factories.cc
```diff
@@ -15,11 +17,15 @@ namespace reinforcement_learning {
   int observation_sender_create(i_sender** retval, const u::configuration&, error_callback_fn*, i_trace* trace_logger, api_status* status);
```
jackgerrits, line 17: Why is `observation_sender_create` a different operation to `observation_api_sender_create`? If the parameters are the same is the difference not in the concrete implementation of the sender?

rlclientlib/live_model_impl.cc
```diff
@@ -421,8 +421,9 @@ namespace reinforcement_learning {
     i_sender* ranking_data_sender;
 
     // Use the name to create an instance of raw data sender for interactions
+    _configuration.set(name::CONFIG_SECTION, name::INTERACTION);
```
jackgerrits, line 424: Mutating the configuration in order to interpret the values internally prior to calling an `init` function seems like a dangerous pattern to me. What is the rationale here?

rlclientlib/logger/http_transport_client.h
```diff
@@ -5,28 +5,30 @@
...
 
+using namespace web::http;
+
 namespace reinforcement_learning {
   class i_trace;
 
-  // The eventhub_client send string data in POST requests to an HTTP endpoint.
-  // It handles authorization headers specific for the Azure event hubs.
-  class eventhub_client : public i_sender {
+  template <typename T>
```
jackgerrits, line 23: What is `T`? Please use a more descriptive parameter name

rlclientlib/rlclientlib.vcxproj.filters
```diff
@@ -136,7 +143,9 @@
     <None Include="schema\v2\CaEvent.fbs" />
     <None Include="schema\v2\MultiSlotEvent.fbs" />
     <None Include="schema\v2\Event.fbs" />
-    <None Include="schema\v2\DedupInfo.fbs" />
+    <None Include="ClassDiagram.cd" />
```
jackgerrits, line 146: Was this intentionally committed?

rlclientlib/utility/apim_http_authorization.cc
```diff
@@ -0,0 +1,15 @@
+#pragma once
+
+#include "apim_http_authorization.h"
+
+namespace reinforcement_learning {
+  int apim_http_authorization::init(const utility::configuration& config, api_status* status, i_trace* trace) {
+    _api_key = config.get(name::HTTP_API_KEY, "dummykey");
```
jackgerrits, line 7: If the API key is missing should this not be an error? We cannot proceed in this case

rlclientlib/utility/apim_http_authorization.h
```diff
@@ -0,0 +1,29 @@
...
+  class apim_http_authorization {
+  public:
+    apim_http_authorization() = default;
+    ~apim_http_authorization() = default;
+
+    int init(const utility::configuration& config, api_status* status, i_trace* trace);
+    int get_http_headers(http_headers& headers, api_status* status);
+
+  private:
+    apim_http_authorization(const apim_http_authorization&) = delete;
```
jackgerrits, line 21: If these are deleted they should not be private. Deleted and public is what you want here

rlclientlib/utility/config_utility.cc
```diff
@@ -110,7 +110,13 @@ namespace reinforcement_learning { namespace utility { namespace config {
       { "InitialCommandLine"        , name::MODEL_VW_INITIAL_COMMAND_LINE},
       { "initialCommandLine"        , name::MODEL_VW_INITIAL_COMMAND_LINE},
       { "ProtocolVersion"           , name::PROTOCOL_VERSION},
-      { "protocolVersion"           , name::PROTOCOL_VERSION}
+      { "protocolVersion"           , name::PROTOCOL_VERSION},
```
jackgerrits, line 113: This is a legacy mapping. Why are we adding more values to this? It is only intended for compat. New keys should only ever be used in a single form with no need for mapping

rlclientlib/utility/eventhub_http_authorization.h
```diff
@@ -0,0 +1,58 @@
...
+      std::string& authorization_string /* out */,
+      long long& valid_until /* out */,
+      api_status* status,
+      i_trace* trace);
+
+    // cannot be copied or assigned
+    eventhub_http_authorization(const eventhub_http_authorization&) = delete;
+    eventhub_http_authorization(eventhub_http_authorization&&) = delete;
+    eventhub_http_authorization& operator=(const eventhub_http_authorization&) = delete;
+    eventhub_http_authorization& operator=(eventhub_http_authorization&&) = delete;
```
jackgerrits, line 44: Deleted functions should be public

jackgerrits (changes requested):

rlclientlib/live_model_impl.cc
slahabar: For this particular one, I did pair with Rajan and he suggested that we should config as a state and mutable not a static settings.

rlclientlib/utility/apim_http_authorization.cc
slahabar: Configuration never throw errors. This is consistent with the pattern through out. There is no change in behavior introduced in this PR.

rlclientlib/azure_factories.cc
slahabar: One uses event hub and another uses http apim auth and different clients and mapped to different types. Not sure what you are suggesting here. Sender works with both auth.

    sender_factory.register_type(value::OBSERVATION_EH_SENDER, observation_sender_create);
    sender_factory.register_type(value::INTERACTION_EH_SENDER, interaction_sender_create);
    sender_factory.register_type(value::OBSERVATION_HTTP_API_SENDER, observation_api_sender_create);
    sender_factory.register_type(value::INTERACTION_HTTP_API_SENDER, interaction_api_sender_create);

rlclientlib/live_model_impl.cc
jackgerrits: @rajan-chari could you explain the rationale then?

rlclientlib/utility/apim_http_authorization.cc
jackgerrits: Just because a configuration hasn't thrown an error yet doesn't seem to be a valid reason to not do it here. Clearly the contract of this function is that it can fail, otherwise it would be a `void` function.

If the API key is not supplied and the `apim_http_authorization` object is in use is that an error? Can RLClientlib operate as normal in that case?

rlclientlib/azure_factories.cc
jackgerrits: Is it possible to use both at the same time? Or just one?

rlclientlib/rlclientlib.vcxproj.filters
slahabar: Autogenerated. Getting rid of ClassDiagram.cd.

rlclientlib/logger/http_transport_client.h
slahabar: Using TAuthorization.

include/constants.h
slahabar: Well, I needed these constants. Moved it to different namespace for readability.

rlclientlib/utility/config_utility.cc
slahabar: Fixed.

rlclientlib/live_model_impl.cc
slahabar: @rajan-chari @jackgerrits can we please sync tomorrow and close this comment? I have done all the functional testing and ready to merge the PR. This is blocking the merge. If there are any other suggestions, I would prefer we create separate item outside of this PR and address separately since this was already reviewed and approved and is a relatively bigger PR to manage.

rlclientlib/utility/apim_http_authorization.cc
jackgerrits: @slahabar this issue is unresolved

rlclientlib/live_model_impl.cc
rajan-chari: Normally to construct/initialize objects,  parameters can simply be added to a constructor on init function.  In the case of factory based construction, adding parameters requires changing signature of the factory itself.  The issue there is that this results in create/init methods having to accept a parameter that may or may not be relevant.   This seems acceptable in cases where the extra parameter is a generic.  For example a logger. 

In this case there is a single factory method, that uses this parameter to select different configurations used for construction.  (interaction vs observation).  So adding an extra parameter to config seems like a reasonable choice given the trade off.

I do agree that we are changing a pattern here.  One mitigation might be to move name::CONFIG_SECTION to a different (internal_constants.h ?) file which would allow us to formulate a policy as follows:  
a) Do not modify externally visible name/values passed in via config
b) Adding/Modifying new name/values is allowed as long as these names are in internal_constants.h

jackgerrits: I like that plan. I think the main thing I was referring to is that we should keep the user inputted configuration in-tact. Adding additional values/arguments is necessary to facilitate the construction like you said.

Moving it to an internal_constants.h hopefully makes it clear to the user and to us working on it that this is not a config value that they should be setting externally but we use it internally for construction of things.

rlclientlib/utility/apim_http_authorization.cc
rajan-chari: If the apim key is invalid, http_transport will report an error and the user has to handle it either in a) the background error handler or b) as warnings in foreground rank/report call.  This condition needs to be handled in any case since apim keys can be rotated and therefore become invalid.  However, from a user perspective, this is graceful system degradation since inference is still possible, even with bad apim keys and web pages are still being served etc.

Alternatively, treating this as an init failure means that rlclientlib itself will fail to initialize and the object will not be usable.  From an operational perspective, this requires catastrophic intervention.   i.e. shit has hit the fan and and I get a call because my website is down.

Given that this condition will be detected and handled for other reasons, this approach seems like an acceptable trade-off to me in this particular case.

That is not to say that missing config values cannot ever constitute a failure.

jackgerrits: It's a separate issue to have an expired or invalid key, and I agree because of rotation and many other things an invalid API key can sneak in here. I am referring to requiring a config value, valid or not. Just to be clear I am not proposing this function to "validate" an API key, just to check something was given.

I am just proposing the init failure is if **no** config value is supplied (not supplying an **invalid** value). Using a default API key of `dummykey` will never work. The fact you forgot to set a config value will be surfaced as a 403 in a background thread and is significantly harder to debug/diagnose than an immediate "this config field is required but not supplied".

If I were debugging rlclientlib I would prefer a foreground error stating "HTTP_API_KEY config required but was not supplied" instead of a background error stating "403 when trying to connect to APIM"

rajan-chari: Makes sense.  Should return an error here.

slahabar: Added check for api key being passed. Throws error if there is none.

rlclientlib/live_model_impl.cc
slahabar: Moved the constants to internal_constants.h file.

jackgerrits (approved):

PR #426, opened 2021-09-20
Multistep rewards

ataymano:


external_parser/parse_example_external.cc
```diff
@@ -94,18 +62,19 @@ void apply_cli_overrides(std::unique_ptr<i_joiner>& joiner, vw *all, const input
...
   }
   if(all->options->was_supplied("reward_function")) {
     v2::RewardFunctionType reward_function;
-    if(!str_to_reward_function(parsed_options.ext_opts->reward_function, reward_function)) {
+    if(!str_to_enum(parsed_options.ext_opts->reward_function, reward_functions, v2::RewardFunctionType_MIN, reward_function)) {
       throw std::runtime_error("Invalid argument to --problem_type " + parsed_options.ext_opts->reward_function);
     }
     joiner->set_reward_function(reward_function, true);
   }
+  joiner->apply_cli_overrides(all, parsed_options);
```
olgavrou, line 77: is this something that the `joiner` needs to know about? can't it be handled at this level?

external_parser/joiners/multistep_example_joiner.cc
```diff
@@ -280,7 +307,8 @@ bool multistep_example_joiner::process_joined(v_array<example *> &examples) {
     return false;
   }
 
-  joined.calc_reward(_loop_info.default_reward, _reward_calculation.value());
+  // TODO: remove
```
olgavrou, line 310: is this to be removed in this PR? the access to the `joined.typed_data` pointer was introduced for binary to dsjson conversion, as that needs to know the ins and outs of the data but in general the pattern we followed was to delegate work to the `joined_event` and in turn to the typed data in `joined_event.h` instead of using that pointer in the joiner

external_parser/joiners/multistep_example_joiner.h
```diff
@@ -23,6 +24,36 @@
 
 namespace v2 = reinforcement_learning::messages::flatbuff::v2;
 
+enum multistep_reward_funtion_type {
```
olgavrou, line 27: reward stuff could be moved to `reward.h` or to a `reward_multistep.h` that imports `reward.h` so that it can access `outcome` struct

external_parser/parse_example_external.cc
ataymano: This PR is about adding multistep specific parameter "multistep_reward" which is controlling behavior specific to multistep_joiner. I agree that it does not look great, but not sure about other ways to generalize this logic across multistep/non multistep joiner implementations.

external_parser/joiners/multistep_example_joiner.cc
```diff
@@ -1,18 +1,21 @@
...
 #include "event_processors/typed_events.h"
+#include "utils.h"
 
 #include <limits.h>
 #include <time.h>
 #include <map>
 #include <stack>
 #include <queue>
 #include <tuple>
+#include <cstdio>
```
zwd-ms, line 18: clean up?

external_parser/joiners/multistep_example_joiner.h
```diff
@@ -52,14 +84,16 @@ class multistep_example_joiner : public i_joiner {
     const v2::Metadata &meta;
     const event_t &event;
   };
+  void set_multistep_reward_function(const multistep_reward_funtion_type type, bool sticky);
 
 private:
   bool populate_order();
   reward::outcome_event
-  process_outcome(const Parsed<v2::OutcomeEvent> &event_meta);
+  process_outcome(const TimePoint timestamp, const v2::Metadata &metadata, const v2::OutcomeEvent& event);
```
zwd-ms, line 92: use `TimePoint&` reference?

external_parser/unit_tests/test_vw_external_parser.cc
```diff
@@ -980,3 +980,117 @@ BOOST_AUTO_TEST_CASE(multistep_unordered_episodes) {
   clear_examples(examples, vw);
   VW::finish(*vw);
 }
+
+BOOST_AUTO_TEST_CASE(multistep_2_episodes_suffix_mean) {
+  // order within episode is not guaranteed. TODO: fix strict ordering
```
zwd-ms, line 985: clean up?

external_parser/unit_tests/test_vw_external_parser.cc
```diff
@@ -980,3 +980,117 @@ BOOST_AUTO_TEST_CASE(multistep_unordered_episodes) {
...
+  BOOST_CHECK_EQUAL(seenB, true);
+  BOOST_CHECK_EQUAL(seenC, true);
+  BOOST_CHECK_EQUAL(seenD, true);
+
+  clear_examples(examples, vw);
+  VW::finish(*vw);
+}
+
+BOOST_AUTO_TEST_CASE(multistep_2_episodes_suffix_sum) {
+  // order within episode is not guaranteed. TODO: fix strict ordering
```
zwd-ms, line 1042: clean up TODO?

external_parser/parse_example_external.cc
olgavrou: ok maybe we could create a multistep `parse_example_multistep` like we have a `parse_example_converter` and rename `parse_example_external` to something else, and have the options available there so that each parser can set it's own options

might be outside of the scope of this PR

external_parser/utils.h
```diff
@@ -0,0 +1,32 @@
...
+
+namespace VW {
+namespace external {
+
+bool stricmp(const char* first, const char* second);
+
+template<typename enum_t>
+bool str_to_enum(
+  const std::string &str,
+  const std::vector<std::pair<const char *, enum_t>> enum_map,
```
zwd-ms, line 18: why not using a real map here?

zwd-ms (approved):

external_parser/joiners/multistep_example_joiner.h
ataymano: Not sure if we need to have access to outcome here. If original reward function is the function of outcome structs, then multistep_reward is function over graph (just list in this simple prototype) with results of reward function in the vertices. It could change though, since this code is more prototypish than the rest of external_parser. That was the reason to:
- make it as function of vector<float>
- keep it inside multistep_joiner for now.

external_parser/joiners/multistep_example_joiner.cc
ataymano: thanks for noticing, fixed

external_parser/joiners/multistep_example_joiner.h
ataymano: thanks for noticing, fixed

external_parser/unit_tests/test_vw_external_parser.cc
ataymano: fixed

external_parser/unit_tests/test_vw_external_parser.cc
ataymano: fixed

external_parser/utils.h
ataymano: no specific reason, i believe. fixed

external_parser/joiners/multistep_example_joiner.cc
ataymano: just removed.

external_parser/parse_example_external.cc
ataymano: yes I am going to schedule some refactoring/ cleaning up work for myself and will try to address it  there.

olgavrou (approved):

PR #427, opened 2021-09-22
Move http_transport_client template implementation code to .h file + rename internal constants namespace

slahabar:
Moves http_transport_client template implementation code to .h file. (no change in code)
Rename internal constants namespace.


rlclientlib/logger/http_transport_client.h
```diff
@@ -89,4 +99,137 @@ namespace reinforcement_learning {
     i_trace* _trace;
     error_callback_fn* _error_callback;
   };
+
+  template <typename TAuthorization>
+  http_transport_client<TAuthorization>::http_request_task::http_request_task(
```
jackgerrits, line 104: Can you please place the function implementations below the class declaration (still in the .h file since its a template) so that the class functions can be read without needing to sift through the impl?

rlclientlib/logger/http_transport_client.h
```diff
@@ -89,4 +99,137 @@ namespace reinforcement_learning {
     i_trace* _trace;
     error_callback_fn* _error_callback;
   };
+
+  template <typename TAuthorization>
```
ataymano, line 103: There are benefits of separating definition and implementation, but it is probably more consistent with the code base to keep them together. It also helps to get rid of "template <typename TAuthorization>" copy-paste.

jackgerrits: Leaving the entire implementation inline means you cannot look at the class definition to see what the interface is without reading past many lines of implementation code though?

include/internal_constants.h
```diff
@@ -1,6 +1,6 @@
 #pragma once
 
-namespace reinforcement_learning { namespace constant {
+namespace reinforcement_learning { namespace internal_name {
```
ataymano, line 3: why do we need this?

rlclientlib/logger/http_transport_client.h
ataymano: yeah I agree with this. My only argument was consistency, but actually you are right - we are not inlining it.

include/internal_constants.h
ataymano: sorry, let me reformulate:
1) why do we need to rename constants to internal_constants?
2) why is the namespace name internal_name (although file name is internal_constants)?

slahabar: Ask was to use better namespace name than "constant". I am open to better name suggestions.

ataymano: this file looks like not only about constants, but about client config as well. maybe 'config_constants"?

rlclientlib/logger/http_transport_client.h
slahabar: Thanks for confirming. Resolving conversation.

include/internal_constants.h
slahabar: updated to suggested

rajan-chari:
looks good.  Thanks

rajan-chari (approved):

ataymano (approved):

PR #429, opened 2021-10-05
feat: Adds new metric DsjsonSumOriginalCostFirstSlot that aggregrates…

slahabar:
… original label cost for the first slot

### DIFF
--- a/external_parser/event_processors/joined_event.h
+++ b/external_parser/event_processors/joined_event.h
@@ -30,6 +30,7 @@ struct typed_joined_event {
 
   virtual void calculate_metrics(dsjson_metrics*) {}
   virtual float get_sum_original_reward() const = 0;
+  virtual float get_sum_original_reward_first_slot() const = 0;
 };
 
 struct cb_joined_event : public typed_joined_event {
@@ -119,6 +120,10 @@ struct cb_joined_event : public typed_joined_event {
   }
 
   float get_sum_original_reward() const override {
+    return original_reward;
+  }
+
+  float get_sum_original_reward_first_slot() const override {
     return original_reward;
   }
 };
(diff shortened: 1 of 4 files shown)


external_parser/joiners/example_joiner.cc
```diff
@@ -540,6 +541,7 @@ void example_joiner::persist_metrics() {
         _joiner_metrics.number_of_skipped_events;
 
     _vw->example_parser->metrics->DsjsonSumCostOriginal = _joiner_metrics.sum_cost_original;
+    _vw->example_parser->metrics->DsjsonSumCostOriginal = _joiner_metrics.sum_cost_original_first_slot;
```
slahabar, line 544: Testing code

slahabar: Dependent on PR https://github.com/VowpalWabbit/vowpal_wabbit/pull/3359

external_parser/event_processors/joined_event.h
```diff
@@ -30,6 +30,7 @@ struct typed_joined_event {
 
   virtual void calculate_metrics(dsjson_metrics*) {}
   virtual float get_sum_original_reward() const = 0;
+  virtual float get_sum_original_reward_first_slot() const = 0;
```
olgavrou, line 33: this virtual function should not be introduced here since `first slot` is something that isn't applicable to all reductions (cb/cats)

what can be done is write an override for `calculate_metrics` in `ccb_joined_event` and call `calculate_metrics` from example_joiner line 464 instead (now that I look at it `sum_cost_original` should also go in there, but that is outside the scope of this PR)

olgavrou: Actually it might be fine to leave it like this, as long as we assume that all reductions have one slot (maybe add a comment above)

PR #430, opened 2021-10-05
feat: Enable subsampling of events via configuration

peterychang:


rlclientlib/utility/config_helper.cc
```diff
@@ -55,6 +68,8 @@ async_batcher_config get_batcher_config(const configuration &config, const char
   res.send_queue_max_capacity = get_int(config, section, name::SEND_QUEUE_MAX_CAPACITY_KB, 16 * 1024) * 1024;
   res.queue_mode = to_queue_mode_enum(get_str(config, section, name::QUEUE_MODE, value::QUEUE_MODE_DROP));
   res.batch_content_encoding = config.get_bool(section, name::USE_DEDUP, false) ? value::CONTENT_ENCODING_DEDUP : value::CONTENT_ENCODING_IDENTITY;
+  // TODO: what should happen if the rate is not (0, 1]?
```
olgavrou, line 71: should we maybe put a warning/throw something here until we figure out what happens if the rate is not (0,1]?

unit_test/async_batcher_test.cc
```diff
@@ -61,6 +64,31 @@ class test_droppable_event : public event {
...
+    return *this;
+  }
+
+  bool try_drop(float drop_prob, int _drop_pass) override {
+    auto prob = static_cast<float>(std::atof(_seed_id.c_str()));
+
+    // logic because floating point comparison is dumb. In this case, 0.7 > 0.7 == true
+    const float tol = 1e-5f;
+    bool is_equal = (prob-drop_prob)*(prob-drop_prob) < tol*tol;
+    return (prob > drop_prob) && !VW::math::are_same(prob, drop_prob, 1e-5f);
```
jackgerrits, line 87:
```suggestion
    constexpr float FLOAT_TOL= 1e-5f;
    return (prob > drop_prob) && !VW::math::are_same(prob, drop_prob, FLOAT_TOL);
```

unit_test/async_batcher_test.cc
```diff
@@ -84,12 +112,26 @@ namespace reinforcement_learning { namespace logger {
...
+  template <>
+  struct json_event_serializer<config_drop_event> {
+    using serializer_t = json_event_serializer<config_drop_event>;
+
+    static int serialize(config_drop_event& evt, std::ostream& out, api_status* status) {
+      out << evt.get_event_id();
+      return error_code::success;
+    }
+
+    static size_t size_estimate(const config_drop_event& evt) { return 1; }
```
jackgerrits, line 125: Isn't the size the length of the event id?

unit_test/async_batcher_test.cc
```diff
@@ -61,6 +64,31 @@ class test_droppable_event : public event {
...
+public:
+  config_drop_event() {}
+  config_drop_event(const std::string& id) : event(id.c_str(), timestamp{}) {}
+
+  config_drop_event(config_drop_event&& other) : event(std::move(other)) {}
+  config_drop_event& operator=(config_drop_event&& other)
+  {
+    if (&other != this) event::operator=(std::move(other));
+    return *this;
+  }
```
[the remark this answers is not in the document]
peterychang: This was just copied from other test events

unit_test/async_batcher_test.cc
peterychang: I'm not sure, I just copied it from the previous examples. It probably is? But I assume this estimate is only used to try and optimize the number of memory allocations, I dont think anything will break from it being wrong

unit_test/async_batcher_test.cc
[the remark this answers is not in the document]
kumpera: That doesn't make it any less 🤮

unit_test/async_batcher_test.cc
kumpera: Size estimate can be anything so any positive value should work.

unit_test/async_batcher_test.cc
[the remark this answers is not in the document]
jackgerrits: Then perhaps it should be removed from the others too?

PR #438, opened 2021-11-18
move generic_event logic to logger thread

peterychang:
There are still several major challenges I need to work out for this PR; namely dealing with the observation workflows for generic_events, simplifying the explicit template type for several objects (currently its `std::function<int(EVENT_TYPE&, api_status*)>>`, and untangling a circular dependency between generic_event and i_logger_extensions

TODO:
- [x] refactor generic_event
- [x] fix calls in logger facade
- [x] fix generic_event observation workflows
- [x] simplify template types if possible


rlclientlib/logger/async_batcher.h
```diff
@@ -17,35 +17,41 @@
...
 
   // This class takes uses a queue and a background thread to accumulate events, and send them by batch asynchronously.
   // A batch is shipped with TSender::send(data)
-  template<typename TEvent, template<typename> class TSerializer = json_collection_serializer>
-  class async_batcher: public i_async_batcher<TEvent> {
+  // TFunc : a function with the prototype int f(Event& evt, api_status* status)
+  //       return value: Error code
+  //       evt    : Event type out-param
+  //       status : Optional status object
+  template<typename TEvent, template<typename> class TSerializer = json_collection_serializer, typename TFunc = std::function<int(TEvent&, api_status*)>>
```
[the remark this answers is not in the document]
peterychang: Hm.. I think you're right, we can probably use a typedef here. I'll give that a try once I finish the changes

rlclientlib/logger/async_batcher.h
```diff
@@ -121,31 +134,33 @@ namespace reinforcement_learning { namespace logger {
     return error_code::success;
   }
 
-  template<typename TEvent, template<typename> class TSerializer>
-  int async_batcher<TEvent, TSerializer>::append(TEvent& evt, api_status* status) {
-    return append(std::move(evt), status);
+  template<typename TEvent, template<typename> class TSerializer, typename TFunc>
+  int async_batcher<TEvent, TSerializer, TFunc>::append(TFunc& func, const char* evt_id, size_t size_estimate, api_status* status) {
```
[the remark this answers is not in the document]
peterychang: `func` should be pushed onto the queue. I forgot to make the change in the code

rlclientlib/logger/event_logger.cc
```diff
@@ -2,37 +2,82 @@
...
 namespace reinforcement_learning { namespace logger {
+using namespace std::placeholders;
   int interaction_logger::log(const char* event_id, const char* context, unsigned int flags, const ranking_response& response, api_status* status, learning_mode learning_mode) {
     const auto now = _time_provider != nullptr ? _time_provider->gmt_now() : timestamp();
-    return append(ranking_event::choose_rank(event_id, context, flags, response, now, 1.0f, learning_mode), status);
+    // std::bind being used here to avoid needing to copy the actual event. Lambdas allow capture statements
+    // in C++14, so we can remove the bind usage at that point
+    auto evt_fn = std::bind(
+      [](ranking_event& out_evt, api_status* status, ranking_event in_evt)->int {
+        out_evt = std::move(in_evt);
```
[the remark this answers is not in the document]
peterychang: We could have a heap object, but we'd either need to use a raw/shared pointer (in the case of a pure lambda), or keep the bind call if we use a unique_ptr. I dislike the idea of passing around a raw pointer and shared_ptr is a bit overkill.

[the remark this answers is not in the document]
jackgerrits: Is the function called right away? As long as it is not possible for anything to throw between creating the func and calling the lambda you could release a raw pointer into the capture and wrap it back in a unique ptr as the first thing in the lambda.

```c++
auto my_thing = std::make_unique<...>();
auto* raw_thing = my_thing.release();
auto lambda = [raw_thing](...){
    auto my_thing_again = std::unique_ptr<...>(raw_thing);
};
```

rlclientlib/logger/event_logger.cc
```diff
@@ -2,37 +2,82 @@
 #include "ranking_event.h"
 #include "err_constants.h"
 #include "time_helper.h"
+
+#include <functional>
 namespace reinforcement_learning { namespace logger {
+using namespace std::placeholders;
```
olgavrou, line 8: the loggers that don't use the generic event's are v1 loggers (right @ataymano ?) Is this change mean to affect v1 and v2 logging, or just v2?

rlclientlib/logger/event_logger.h
```diff
@@ -124,13 +128,34 @@ class multi_slot_logger : public event_logger<multi_slot_decision_event> {
...
+          {
+            RETURN_IF_FAIL(in_evt.transform(ext, serializer, status, args...));
+            out_evt = std::move(in_evt);
+            return error_code::success;
+          },
+          _1,
+          _2,
+          generic_event(event_id, now, type, context, _app_id)
+        );
+      return append(std::move(evt_fn), event_id, 1, status);
```
olgavrou, line 155: will all these pointers (`api_status`, `ext`) be now used in an asynchronous manner, withought locking, by both threads? I am wondering this will cause a problem? i.e. thread 1 sets status to fail, thread 2 sets status to success, thread 1 returns now incorrect status

peterychang: This workflow already exists in the code. I'm not sure if its safe, but it would need to be a larger refactor to fix any issues with it

rlclientlib/ranking_event.h
```diff
@@ -81,7 +81,9 @@ namespace reinforcement_learning {
   class decision_ranking_event : public event {
   public:
     decision_ranking_event();
+    decision_ranking_event(const decision_ranking_event& other) = default;
```
rajan-chari, line 84: Peter mentioned, this can be deleted now and set to default.  (Reminder)

rlclientlib/logger/async_batcher.h
```diff
@@ -25,27 +30,33 @@ namespace reinforcement_learning { namespace logger {
...
   class async_batcher: public i_async_batcher<TEvent> {
   public:
     using shared_state_t = typename TSerializer<TEvent>::shared_state_t;
+    using TFunc = std::function<int(TEvent&, api_status*)>;
 
     int init(api_status* status) override;
 
-    int append(TEvent&& evt, api_status* status = nullptr) override;
-    int append(TEvent& evt, api_status* status = nullptr) override;
+    int append(TFunc&& func, string_view evt_id, TEvent* event, api_status* status = nullptr) override;
```
rajan-chari, line 58: To check: evt_id may not be needed?

rlclientlib/live_model_impl.h
```diff
@@ -104,6 +104,10 @@ namespace reinforcement_learning
     std::unique_ptr<model_management::i_data_transport> _transport{nullptr};
     std::unique_ptr<model_management::i_model> _model{nullptr};
 
+    // These objects need to be handled VERY carefully. _interaction_logger will spawn a thread
+    // which will contain a pointer to the _logger_extensions object. The thread's lifetime is
+    // tied to the _interaction_logger object. If any of these conditions change, the _logger_extentions
```
jackgerrits, line 109: Why not just make it a shared object now so we don't have this scary message?

rlclientlib/logger/event_logger.cc
```diff
@@ -2,28 +2,72 @@
...
+
+#include <functional>
+#include <memory>
 namespace reinforcement_learning { namespace logger {
   int interaction_logger::log(const char* event_id, string_view context, unsigned int flags, const ranking_response& response, api_status* status, learning_mode learning_mode) {
     const auto now = _time_provider != nullptr ? _time_provider->gmt_now() : timestamp();
-    return append(ranking_event::choose_rank(event_id, context, flags, response, now, 1.0f, learning_mode), status);
+    // using shared_ptr because we can't move a unique_ptr in C++11
+    // We should replace them in C++14
+    auto evt_sp = std::make_shared<ranking_event>();
```
rajan-chari, line 13: Switch to raw pointer here.  The usage and life time is very clear.  The overhead of shared_ptr seems unnecessary given we are doing this for perf reasons.

rajan-chari (approved):
A few minor things to be reviewed.  Looks good to me after that.

jackgerrits (commented):
I am not very familiar with the code this is changing so I am finding reviewing this a bit tricky. The main question I have is, now that the code is done do we understand what the performance improvements are over the previous implementation? Can we measure this to get a picture?

rlclientlib/logger/event_logger.cc
olgavrou: I think we should avoid switching to a raw pointer at all costs. Shared pointers mostly cause a perf hit if they are actually shared this one here seems to just being moved around.

If there is a measurable perf hit then we can consider it

jackgerrits: +1 to Olga's comment. Safety should be the default approach unless there is strong evidence of a performance issue.

peterychang: discussed outside the PR: A smart object of some kind is required because the functor will never get run for dropped events (eg: when being pruned) which would mean the objects never get destroyed. Rajan to investigate alternatives when convenient

rlclientlib/live_model_impl.h
peterychang: > I am not very familiar with the code this is changing so I am finding reviewing this a bit tricky. The main question I have is, now that the code is done do we understand what the performance improvements are over the previous implementation? Can we measure this to get a picture?

I'll try to get some performance metrics, but it'll be a bit tricky since I'll either have to figure out how to stub out the logger sink or we'll have I/O noise to deal with

peterychang: > Why not just make it a shared object now so we don't have this scary message?

I could, I just found it to be unnecessary in this case. @rajan-chari How would you feel about switching this case to a shared_ptr?

rajan-chari: The concern for proper cleanup order is prudent.  rlclientlib thread and object lifetimes were carefully designed for proper cleanup.  i.e. thread are signaled to cleanup and join during shutdown before all resources are deallocated.  If we find that this is not the case, it should be fixed.

rajan-chari: I believe in the next sprint, we are auditing suspect shutdown hygiene issue in rlclientlib.  Might be a good time to confirm that things still work as designed.

rajan-chari: The general idea is that live_model_impl has ownership of all these resources and should clean them up after proper shutdown of all background threads.

rajan-chari: unique_ptr signals this contract and design.  I'd like to leave that in place.

PR #439, opened 2021-12-07
Merge multistep feature branch.

zwd-ms:


include/model_mgmt.h
```diff
@@ -77,6 +78,7 @@ namespace reinforcement_learning { namespace model_management {
       virtual int choose_continuous_action(const char* features, float& action, float& pdf_value, std::string& model_version, api_status* status = nullptr) = 0;
       virtual int request_decision(const std::vector<const char*>& event_ids, const char* features, std::vector<std::vector<uint32_t>>& actions_ids, std::vector<std::vector<float>>& action_pdfs, std::string& model_version, api_status* status = nullptr) = 0;
       virtual int request_multi_slot_decision(const char* event_id, const std::vector<std::string>& slot_ids, const char* features, std::vector<std::vector<uint32_t>>& actions_ids, std::vector<std::vector<float>>& action_pdfs, std::string& model_version, api_status* status = nullptr) = 0;
+      virtual int choose_rank_multistep(uint64_t rnd_seed, const char* features, const episode_history* history, std::vector<int>& action_ids, std::vector<float>& action_pdf, std::string& model_version, api_status* status = nullptr) = 0;
```
ataymano, line 81: probably better to change history to const &

rlclientlib/multistep.cc
```diff
@@ -0,0 +1,44 @@
+#include "multistep.h"
+#include "err_constants.h"
+
+namespace reinforcement_learning {
+  void episode_history::update(const char* event_id, const char* previous_event_id, const char* context, const ranking_response& resp) {
+    _depths[event_id] = this->get_depth(previous_event_id) + 1;
+  }
+
+  std::string episode_history::get_context(const char* previous_event_id, const char* context) const {
+    return R"({"episode":{"depth":")" + std::to_string(this->get_depth(previous_event_id) + 1) + "\"}," + std::string(context + 1);
```
ataymano, line 10: 1. Probably better to discuss with broader audience whether it is ok for now to introduce hard-coded namespace.
2. If so, make json manipulations cleaner

zwd-ms: yes agreed.

include/model_mgmt.h
zwd-ms: done

include/multistep.h
```diff
@@ -0,0 +1,60 @@
...
+#include <utility>
+#include <vector>
+#include <string>
+#include <map>
+#include <memory>
+
+#include "ranking_response.h"
+
+namespace reinforcement_learning {
+  class episode_history {
```
ataymano, line 15: Should be an interface in the future.

rlclientlib/multistep.cc
ataymano: More detailed explanation of the problems with this piece of code:
1) episode history class is supposed to be splitted into interface (which is introducing extension point) and implementation (existing one which is tracking #step is just one of possible implementations).
For existing implementation:
1) Injecting new feature into hardcoded namespace/feature_name is definitely a hack (obvious problem is existence of the same namespace in the context).
2) json manipulation is clearly bad.

Given that this code is still considered as prototype and not interfering with any of cb/ccb stack, it is fine to me to commit as is and work on improvements incrementally, but can you please share your thoughts (@olgavrou @jackgerrits @lokitoth )?

external_parser/unit_tests/test_vw_external_parser.cc
```diff
@@ -247,128 +247,6 @@ BOOST_AUTO_TEST_CASE(ccb_compare_dsjson_with_fb_models) {
                                 buffer_dsjson_model.end());
 }
```
olgavrou, line 249: are these tests removed by mistake?

rlclientlib/multistep.cc
olgavrou: Since it doesn't affect any path that is in use I think it's fine to make changes and cleanups incrementally

olgavrou (approved):

external_parser/unit_tests/test_vw_external_parser.cc
zwd-ms: they were already added back if you compare all commits.

PR #446, opened 2022-02-07
Token auth&model api

Tparuchuri:


README.md
```diff
@@ -88,7 +88,7 @@ cmake --build build --target all -j 4
 
 #### Dependencies
 ```cmd
-vcpkg install --triplet x64-windows zlib boost-system boost-program-options boost-test boost-align boost-foreach boost-math boost-uuid cpprestsdk flatbuffers
+vcpkg install --triplet x64-windows zlib boost-system boost-program-options boost-locale boost-test boost-align boost-foreach boost-math boost-uuid cpprestsdk flatbuffers
```
jackgerrits, line 91: Why is pulling in a new dependency required?

Tparuchuri: To support Token auth new changes are implemented to allow HTTP header name from config. In order to convert it to http_headers::key_type, wstring conversion is required. Boost locale is supported in both linux and windows, but as it's not currently available in our project have to add it in dependency. Tried using codecvt but checks are failing in linux. Looking for other alternatives

jackgerrits: What is the type of `http_headers::key_type`? It seems to be std::string based on the docs: [1](https://microsoft.github.io/cpprestsdk/classweb_1_1http_1_1http__headers.html#a00428d8e7a1ee544227efdae54550835) [2](https://microsoft.github.io/cpprestsdk/namespaceutility.html)

Is the conversion from or to wide string?

rlclientlib/model_mgmt/restapi_data_transport.cc
```diff
@@ -39,14 +43,23 @@ namespace reinforcement_learning { namespace model_management {
...
-      .then([&](http_response response) {
-      if ( response.status_code() != 200 )
-        RETURN_ERROR_ARG(_trace, status, http_bad_status_code, _httpcli->get_url());
-
+    auto request_task = _httpcli->request(request).then([&](http_response response) {
+      if ( response.status_code() != 200)
+      {
+        if ( response.status_code() == 404)
+        {
+          RETURN_IF_FAIL(get_data_info_using_get(response, status));
```
jackgerrits, line 55: Why immediately call get if head fails?

rlclientlib/model_mgmt/restapi_data_transport.cc
```diff
@@ -69,17 +82,39 @@ namespace reinforcement_learning { namespace model_management {
...
+          }
+          response = get_response;
+          return error_code::success;
+      });
+      try {
+          return request_task.get();
+      }
+      catch (const std::exception& e) {
+          RETURN_ERROR_LS(_trace, status, exception_during_http_req) << e.what() << "\n URL: " << _httpcli->get_url();
+      }
```
jackgerrits, line 102: This seems to be mostly duplicated with get_data_info. Why not factor out into a `make_request_foo` and factor out the HTTP method being used?

rlclientlib/model_mgmt/restapi_data_transport.cc
Tparuchuri: while using modelBlobUri it has head api call where as /model api in frontEnd doesn't have HEAD call at the moment. Instead if 404 is received we will give it a try with GET method

rlclientlib/model_mgmt/restapi_data_transport.cc
Tparuchuri: updated logi

rlclientlib/azure_factories.cc
```diff
@@ -33,14 +35,32 @@ namespace reinforcement_learning {
...
+    if (model_uri == nullptr)
+    {
+      RETURN_ERROR(trace_logger, status, http_model_uri_not_provided);
+    }
+    i_http_client* client;
+    RETURN_IF_FAIL(create_http_client(model_uri, config, &client, status));
+
+    //Initialize header
+    http_headers header;
+    header_authorization* api_obj = new header_authorization();
```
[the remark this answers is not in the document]
Tparuchuri: No specific reason. it's no longer needed

jackgerrits (changes requested):

rlclientlib/utility/header_authorization.cc
```diff
@@ -0,0 +1,25 @@
...
+    #ifdef _WIN32
+    _http_api_header_key_name = u::conversions::utf8_to_utf16(config.get(name::HTTP_API_HEADER_KEY_NAME, value::HTTP_API_DEFAULT_HEADER_KEY_NAME));
+    #else
+    _http_api_header_key_name = config.get(name::HTTP_API_HEADER_KEY_NAME, value::HTTP_API_DEFAULT_HEADER_KEY_NAME);
+    #endif
+    return error_code::success;
+  }
+
+  int header_authorization::insert_authorization_header(http_headers& headers, api_status* status) {
+    headers.add(_http_api_header_key_name, _api_key.c_str());
```
ataymano, line 22: although looks like we are trying not to call insert_authorization_header if init wasn't successful, it would be cleaner to return some error code here (http_api_key_not_provided) as well if class was not properly initialized

include/model_mgmt.h
```diff
@@ -71,6 +71,10 @@ namespace reinforcement_learning { namespace model_management {
       CA
     };
 
+    enum class model_source {
+        AZURE,
+        HTTPAPI
```
johnhuang01, line 76: Should we use HTTP_API?

johnhuang01 (approved):
Left a minor comment.

include/model_mgmt.h
Tparuchuri: modified

rlclientlib/utility/header_authorization.cc
Tparuchuri: added check

rlclientlib/model_mgmt/restapi_data_transport.cc
```diff
@@ -39,14 +44,21 @@ namespace reinforcement_learning { namespace model_management {
...
-    auto request_task = _httpcli->request(methods::HEAD)
-      // Handle response headers arriving.
-      .then([&](http_response response) {
-      if ( response.status_code() != 200 )
-        RETURN_ERROR_ARG(_trace, status, http_bad_status_code, _httpcli->get_url());
-
+    auto request_task = _httpcli->request(request).then([&](http_response response) {
+      if (response.status_code() != 200){
+        if (response.status_code() == 404 && _retry_get_data){
+          _retry_get_data = false;
```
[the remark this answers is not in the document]
Tparuchuri: added comment in code. Basically, we will try with HEAD call first if it fails with 404, we will try try with GET only once

rlclientlib/azure_factories.cc
```diff
@@ -51,8 +51,7 @@ namespace reinforcement_learning {
     }
     i_http_client* client;
     RETURN_IF_FAIL(create_http_client(model_uri, config, &client, status));
-
-    *retval = new m::restapi_data_transport(client, config, m::model_source::HTTPAPI, trace_logger);
+    *retval = new m::restapi_data_transport(std::move(std::unique_ptr<i_http_client>(client)), config, m::model_source::HTTP_API, trace_logger);
```
jackgerrits, line 54: This is already an rvalue you should not need to move it.

Tparuchuri: My bad. Initially tested it by assigning to a variable. Corrected it

jackgerrits (approved):
LGTM

PR #447, opened 2022-02-14
Send the actual number of interaction events received to front end

Tparuchuri:


unit_test/async_batcher_test.cc
```diff
@@ -147,16 +167,53 @@ BOOST_AUTO_TEST_CASE(flush_timeout) {
...
 
+BOOST_AUTO_TEST_CASE(flush_timeout_with_counter_ENABLE) {
+  std::vector<std::string> items;
+  std::vector<int> events_count;
+  auto s = new message_sender(items, events_count);
+  size_t timeout_ms = 100; //set a short timeout
+  error_callback_fn error_fn(expect_no_error, nullptr);
+  utility::watchdog watchdog(nullptr);
+  utility::async_batcher_config config;
+  config.send_high_water_mark = 262143;
```
johnhuang01, line 178: Is it a reason to use 262143 here? Also line 180, 8192 is also just a randomly picked number?

rlclientlib/logger/async_batcher.h
```diff
@@ -153,9 +164,37 @@ namespace reinforcement_learning { namespace logger {
 
     RETURN_IF_FAIL(collection_serializer.finalize(status));
 
+    if (_events_counter_status == events_counter_status::ENABLE) {
+      _buffer_end_event_number = evt.get_event_number();
+      *original_count = (_buffer_end_event_number - _buffer_start_event_number);
+      if (_queue.size() == 0) {
+        //reset counter values if queue is empty to prevent overflow incase of too many events
+        *original_count+= reset_counter() - _buffer_end_event_number;
```
johnhuang01, line 172: Nit: add a space before +=

rlclientlib/logger/async_batcher.h
```diff
@@ -136,10 +145,12 @@ namespace reinforcement_learning { namespace logger {
   int async_batcher<TEvent, TSerializer>::fill_buffer(
                                                       std::shared_ptr<utility::data_buffer>& buffer,
                                                       size_t& remaining,
-                                                      api_status* status)
+                                                      api_status* status,
+                                                      unsigned int* original_count)
   {
     TEvent evt;
     TSerializer<TEvent> collection_serializer(*buffer.get(), _batch_content_encoding, _shared_state);
+    if (_events_counter_status == events_counter_status::ENABLE) { _buffer_start_event_number = _buffer_end_event_number; }
```
johnhuang01, line 153: Can you move 
_buffer_start_event_number = _buffer_end_event_number;
in line 168 below?

rlclientlib/logger/async_batcher.h
```diff
@@ -153,9 +164,37 @@ namespace reinforcement_learning { namespace logger {
 
     RETURN_IF_FAIL(collection_serializer.finalize(status));
 
+    if (_events_counter_status == events_counter_status::ENABLE) {
+      _buffer_end_event_number = evt.get_event_number();
```
johnhuang01, line 168: Can we change _buffer_start_event_number from a member variable to a local variable as it is only used in this method?

rlclientlib/logger/async_batcher.h
```diff
@@ -199,14 +245,15 @@ namespace reinforcement_learning { namespace logger {
...
   {}
 
   template<typename TEvent, template<typename> class TSerializer>
   async_batcher<TEvent, TSerializer>::~async_batcher() {
-    // Stop the background procedure the queue before exiting
-    _periodic_background_proc.stop();
-    if (_queue.size() > 0) {
-      flush();
-    }
+      // Stop the background procedure the queue before exiting
```
johnhuang01, line 253: Let's keep the original 2-space indentation.

rlclientlib/logger/async_batcher.h
```diff
@@ -153,9 +164,37 @@ namespace reinforcement_learning { namespace logger {
...
   }
 
+  template<typename TEvent, template<typename> class TSerializer>
+  void async_batcher<TEvent, TSerializer>::increment_counter()  {
+    std::unique_lock<std::mutex> mlock(_counter_mutex);
+    if (_events_counter_status == events_counter_status::ENABLE) { ++_event_number; }
+  }
+
+  template<typename TEvent, template<typename> class TSerializer>
+  unsigned int async_batcher<TEvent, TSerializer>::reset_counter() {
```
johnhuang01, line 188: Maybe change the name to reset_counter_get_count?

tyclintw (changes requested):

include/sender.h
```diff
@@ -9,8 +9,12 @@ namespace reinforcement_learning {
     using buffer = std::shared_ptr<utility::data_buffer>;
     virtual int init(const utility::configuration& config, api_status* status) = 0;
     int send(const buffer& data, api_status* status = nullptr) { return v_send(data, status); }
+    int send(const buffer& data, unsigned int number_of_events, api_status* status = nullptr) { return v_send(data, number_of_events,status); }
```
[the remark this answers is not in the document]
Tparuchuri: added space

rlclientlib/utility/config_helper.cc
```diff
@@ -47,6 +47,14 @@ static const char* get_str(const configuration &config, const char *section, con
   return config.get(property, defval);
 }
 
+static const char* get_default_sender_implmentation(const char* section)
+{
+  if (section == OBSERVATION_SECTION) {
```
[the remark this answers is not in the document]
Tparuchuri: used std::strcmp

rlclientlib/utility/config_helper.h
```diff
@@ -8,6 +8,12 @@ namespace reinforcement_learning {
     BLOCK//queue block if it is full
   };
 
+  //this enum sets the counter for number of events behaviour in aysnc_batcher
+  enum class events_counter_status {
```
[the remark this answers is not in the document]
Tparuchuri: This enum is required, bcz it's set to ENABLE only certain conditions are met like for now protocol version should be V2 and interaction implementation sender has to be INTERACTION_HTTP_API_SENDER

rlclientlib/logger/async_batcher.h
Tparuchuri: added

rlclientlib/logger/async_batcher.h
Tparuchuri: changed

rlclientlib/logger/async_batcher.h
Tparuchuri: updated

rlclientlib/generic_event.cc
```diff
@@ -31,7 +31,11 @@ namespace reinforcement_learning {
 
   const char* generic_event::get_id() const { return _id.c_str(); }
 
-  const char* generic_event::get_app_id() const { return _app_id.c_str(); }  
+  const char* generic_event::get_app_id() const { return _app_id.c_str(); } 
+
+  const unsigned int generic_event::get_event_number() { return _event_number; }
```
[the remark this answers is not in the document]
Tparuchuri: renamed

rlclientlib/logger/async_batcher.h
Tparuchuri: updated

unit_test/async_batcher_test.cc
Tparuchuri: this is something that was set for already existed test cases. I have used the same parameters but with event_counter_status ENABLE to make sure new implementation is not impacting the existing logic

rlclientlib/logger/async_batcher.h
```diff
@@ -153,9 +162,39 @@ namespace reinforcement_learning { namespace logger {
 
     RETURN_IF_FAIL(collection_serializer.finalize(status));
 
+    if (_events_counter_status == events_counter_status::ENABLE) {
+      unsigned int _buffer_start_event_number;
```
johnhuang01, line 166: Can you remove _ in the variable name? A local variable should not have _ prefix based on the pattern we have.

rlclientlib/logger/async_batcher.h
```diff
@@ -80,6 +84,9 @@ namespace reinforcement_learning { namespace logger {
     utility::object_pool<utility::data_buffer> _buffer_pool;
     const char* _batch_content_encoding;
     float _subsample_rate;
+    events_counter_status _events_counter_status;
+    unsigned int _number_of_events = 0;
```
olgavrou, line 88: could these be initialized in the constructor with the other member variables?

rlclientlib/logger/async_batcher.h
```diff
@@ -52,9 +52,13 @@ namespace reinforcement_learning { namespace logger {
...
     int fill_buffer(std::shared_ptr<utility::data_buffer>& retbuffer,
       size_t& remaining,
-      api_status* status);
+      api_status* status,
+      unsigned int* original_count);
 
     void flush(); //flush all batches
+    void increment_counter();
+    unsigned int reset_counter_get_count();
+    std::mutex _counter_mutex;
```
olgavrou, line 61: super nit: maybe move the mutex to be with the other member variables below to follow the structure of the file

rlclientlib/logger/http_transport_client.h
```diff
@@ -174,6 +179,9 @@ namespace reinforcement_learning {
   template <typename TAuthorization>
   int http_transport_client<TAuthorization>::init(const utility::configuration& config, api_status* status) {
     RETURN_IF_FAIL(_authorization.init(config, status, _trace));
+    _config = new utility::configuration(config);
```
olgavrou, line 182: do we need to create a new configuration object here just so that we can get a setting? could we not just re-use the `config` to do `const char* section = config->get(config_constants::CONFIG_SECTION, config_constants::INTERACTION);`?

rlclientlib/model_mgmt/restapi_data_transport.cc
```diff
@@ -51,8 +51,8 @@ namespace reinforcement_learning { namespace model_management {
     // Build request URI and start the request.
     auto request_task = _httpcli->request(request).then([&](http_response response) {
       if (response.status_code() != 200){
-        //if the call using HEAD fails with 404 try with GET only once and return the results of GET request call
-        if (response.status_code() == 404 && _retry_get_data){
+        //if the call using HEAD fails try with GET only once and return the results of GET request call
+        if (_retry_get_data){
```
olgavrou, line 55: what is the motivation behind removing this 404 check?

johnhuang01 (approved):

rlclientlib/logger/async_batcher.h
```diff
@@ -136,11 +144,12 @@ namespace reinforcement_learning { namespace logger {
   int async_batcher<TEvent, TSerializer>::fill_buffer(
                                                       std::shared_ptr<utility::data_buffer>& buffer,
                                                       size_t& remaining,
-                                                      api_status* status)
+                                                      api_status* status,
```
olgavrou, line 147: it looks like `original_count` can be passed in as a reference here instead of a pointer (no need to dereference or do a nullptr check)

rlclientlib/logger/http_transport_client.h
Tparuchuri: we need the config when we are trying to create new HTTP client with the new URL (in case we are appending events_counts in the URL and creating new HTTP client)

rlclientlib/model_mgmt/restapi_data_transport.cc
Tparuchuri: there are two ways we are getting the model. One is using the azure blob uri and the other is using model API in front end. azure blob uri support HEAD request where as model API doesn't at the moment. So initially when we try with head using model API it throws 404 but when tested using local host it's throwing 405. So I thought irrespective of error code we give another try with GET instead of HEAD when the call with HEAD fails

rlclientlib/logger/http_transport_client.h
olgavrou: If the new client is passed into the `http_transport_client` constructor same as the current `_client` then it can be constructed in `azure_factories.cc` where the config is known. That way we don't need to create a new config object and store it here

olgavrou: connects with this comment: https://github.com/VowpalWabbit/reinforcement_learning/pull/447/files/b08410ab9bc18a301d2939ec68f7cdd37fe11501#r807140690

rlclientlib/logger/http_transport_client.h
```diff
@@ -199,6 +207,22 @@ namespace reinforcement_learning {
 
   template <typename TAuthorization>
   int http_transport_client<TAuthorization>::v_send(const buffer& post_data, api_status* status) {
+      return v_send(post_data, _client.get(), status);
+  }
+
+  template <typename TAuthorization>
+  int http_transport_client<TAuthorization>::v_send(const buffer& post_data, unsigned int event_counts, api_status* status) {
+    if (event_counts != 0 &&  _event_counter_status == events_counter_status::ENABLE) {
+      i_http_client* new_client;
```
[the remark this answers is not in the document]
Tparuchuri: need new http client as every time we need to create new URI based on the events count we get, instead of passing a pointer passing reference now

rlclientlib/logger/async_batcher.h
Tparuchuri: passing as reference instead of pointer

rlclientlib/logger/async_batcher.h
```diff
@@ -136,11 +144,12 @@ namespace reinforcement_learning { namespace logger {
...
+                                                      unsigned int& original_count)
```
ataymano, line 148: this repo has convention of having "api_status* status" argument as the last one

rlclientlib/logger/http_transport_client.h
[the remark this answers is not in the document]
olgavrou: ok well in that case I think this would be a good in-between solution:

- have a member variable of type `unique_ptr` called something like `client_with_event_count`
- have a local raw pointer variable `new_client` to be passed into `create_http_client` and initialized
- call `client_with_event_count.reset(new_client)` which should delete the previous http client and store the freshly 
allocated one in the unique ptr
- pass the unique ptr `client_with_event_count` into `v_send`

rlclientlib/logger/http_transport_client.h
olgavrou: if the configuration needs to be copied here it should be a unique_ptr so that its memory is cleaned up during destruction

rlclientlib/logger/http_transport_client.h
```diff
@@ -41,6 +43,8 @@ namespace reinforcement_learning {
     ~http_transport_client();
   protected:
     int v_send(const buffer& data, api_status* status) override;
+    int v_send(const buffer& data, unsigned int event_counts, api_status* status) override;
```
ataymano, line 46: It seems like abstraction leakage here. Maybe it is possible to introduce original_event_count as a field of event batch fralbuffers message and track it there?

PR #448, opened 2022-02-15
send total number of events (events sent as part of buffer + events dropped) to front end for billing

Tparuchuri:


rlclientlib/generic_event.h
```diff
@@ -40,6 +40,10 @@ namespace reinforcement_learning {
 
     const char* get_app_id() const;
 
+    const unsigned int get_number_of_events();
```
ataymano, line 43: unsigned int

rlclientlib/generic_event.cc
```diff
@@ -13,17 +13,17 @@ namespace reinforcement_learning {
...
     : _id(id)
     , _client_time_gmt(ts)
     , _payload_type(type)
     , _payload(std::move(payload))
     , _pass_prob(pass_prob)
     , _content_type(content_type) 
-    , _app_id(app_id) {}
-
+    , _app_id(app_id)
+    , _number_of_events(0) {}
```
ataymano, line 26: number_of_events as property of event may be a bit unclear - maybe something like event_index?

rlclientlib/ranking_event.h
```diff
@@ -22,6 +22,8 @@ namespace reinforcement_learning {
     event& operator=(event&&) = default;
     virtual ~event() = default;
     float get_pass_prob() const;
+    const unsigned int get_number_of_events();
```
ataymano, line 25: unsigned int

rlclientlib/logger/async_batcher.h
```diff
@@ -151,11 +158,43 @@ namespace reinforcement_learning { namespace logger {
       }
     }
 
-    RETURN_IF_FAIL(collection_serializer.finalize(status));
-
+    if (_events_counter_status == events_counter_status::ENABLE) {
+      unsigned int buffer_start_event_number;
```
ataymano, line 162: unsigned int buffer_start_event_number  = _buffer_end_event_number looks a bit cleaner (the same about original_count)

rlclientlib/logger/async_batcher.h
```diff
@@ -151,11 +158,43 @@ namespace reinforcement_learning { namespace logger {
...
+      unsigned int original_count;
+      buffer_start_event_number = _buffer_end_event_number;
+      _buffer_end_event_number = evt.get_number_of_events();
+      original_count = (_buffer_end_event_number - buffer_start_event_number);
+      if (_queue.size() == 0) {
```
ataymano, line 167: not sure about this part:
1) for high traffic loop (which we are targeting with this "if") _queue.size() will never be 0 so the trick probably will never work
2) reset_counter_get_count() is obtaining lock
Maybe we can just make this counter uint64_t?

olgavrou: I'll agree here, with the counter as uint64 we will most likely restart any rlclientlib instance before overflowing the counter

ataymano: limit is ~ 10^19. so for 1 million RPS setting we have 10^19 / (365 * 24 * 3600) > 10^11 years until restart

rlclientlib/ranking_event.h
```diff
@@ -22,6 +22,8 @@ namespace reinforcement_learning {
     event& operator=(event&&) = default;
     virtual ~event() = default;
     float get_pass_prob() const;
+    unsigned long long int get_event_index();
```
ataymano, line 25: can we make it uint64_t here as well (for consistency)?

rlclientlib/serialization/json_serializer.h
```diff
@@ -117,6 +117,8 @@ namespace reinforcement_learning { namespace logger {
       return error_code::success;
     }
 
+    int finalize(api_status* status, unsigned long long int original_event_count) { return finalize(status); }
```
ataymano, line 120: uint64_t

rlclientlib/logger/async_batcher.h
```diff
@@ -80,6 +81,9 @@ namespace reinforcement_learning { namespace logger {
     utility::object_pool<utility::data_buffer> _buffer_pool;
     const char* _batch_content_encoding;
     float _subsample_rate;
+    events_counter_status _events_counter_status;
+    unsigned long long int _event_index = 0;
```
ataymano, line 85: uint64_t

rlclientlib/logger/async_batcher.h
```diff
@@ -105,6 +109,7 @@ namespace reinforcement_learning { namespace logger {
       }
     }
     
+    if (_events_counter_status == events_counter_status::ENABLE) { evt.set_event_index(_event_index); }
```
ataymano, line 112: actually there is a problem here:
increment_counter_from_thread_1 -> increment_counter_from_thread_2 -> set_event_index_from_thread1 -> set_event_index_from_thread_2
in this case both events will get the same index

rlclientlib/dedup.cc
```diff
@@ -269,6 +269,8 @@ struct dedup_collection_serializer
     return error_code::success;
   }
 
+  int finalize(api_status* status, unsigned long long int original_event_count) { return finalize(status); }
```
ataymano: uint64_t

rlclientlib/generic_event.h
```diff
@@ -57,5 +61,6 @@ namespace reinforcement_learning {
     float _pass_prob = 1.0;
     event_content_type _content_type;
     std::string _app_id;
+    unsigned long long int _event_index;
```
ataymano: uint64_t

rlclientlib/model_mgmt/restapi_data_transport.cc
```diff
@@ -51,8 +51,8 @@ namespace reinforcement_learning { namespace model_management {
     // Build request URI and start the request.
     auto request_task = _httpcli->request(request).then([&](http_response response) {
       if (response.status_code() != 200){
-        //if the call using HEAD fails with 404 try with GET only once and return the results of GET request call
-        if (response.status_code() == 404 && _retry_get_data){
+        //if the call using HEAD fails try with GET only once and return the results of GET request call
```
Tparuchuri, line 54: This is not related to this change. But this is to retry with GET method call once when HEAD method call fails. Model API in front end doesn't have HEAD method call for now, but might be added later. So irrespective of kind of model URI try with HEAD first and if it fails try with GET

rlclientlib/logger/async_batcher.h
```diff
@@ -151,11 +156,28 @@ namespace reinforcement_learning { namespace logger {
...
+      RETURN_IF_FAIL(collection_serializer.finalize(status, original_count));
+    }
+    else {
+      RETURN_IF_FAIL(collection_serializer.finalize(status));
+    }
     return error_code::success;
   }
 
+  template<typename TEvent, template<typename> class TSerializer>
+  void  async_batcher<TEvent, TSerializer>::update_event_with_index(TEvent evt,TEvent* item)  {
```
ataymano, line 172: why do you need to have second argument?
what is wrong with just
update_event_index(TEvent&& evt) {
...
     evt.set_event_index(_evbent_index)
}

rlclientlib/logger/async_batcher.h
Tparuchuri: now incrementing the event_index and updating it in event are performed under lock so no two threads can access simultaneously

rlclientlib/logger/async_batcher.h
```diff
@@ -96,16 +101,16 @@ namespace reinforcement_learning { namespace logger {
...
-
+    TEvent evt_with_index;
+    update_event_with_index(std::move(evt), &evt_with_index);
     // If subsampling rate is < 1, then run subsampling logic
     if(_subsample_rate < 1) {
-      if(evt.try_drop(_subsample_rate, constants::SUBSAMPLE_RATE_DROP_PASS)) {
+      if(evt_with_index.try_drop(_subsample_rate, constants::SUBSAMPLE_RATE_DROP_PASS)) {
         // If the event is dropped, just get out of here
-        return error_code::success;
+        return error_code::success;;
```
olgavrou, line 110: nit: double `;`

rlclientlib/logger/async_batcher.h
```diff
@@ -151,11 +156,28 @@ namespace reinforcement_learning { namespace logger {
       }
     }
 
-    RETURN_IF_FAIL(collection_serializer.finalize(status));
-
+    if (_events_counter_status == events_counter_status::ENABLE) {
+      uint64_t buffer_start_event_number = _buffer_end_event_number;
```
johnhuang01, line 160: Can we change buffer_start_event_number to buffer_start_event_index and _buffer_end_event_number to _buffer_end_event_index? You are accessing _event_index.

rlclientlib/serialization/json_serializer.h
```diff
@@ -117,6 +117,8 @@ namespace reinforcement_learning { namespace logger {
       return error_code::success;
     }
 
+    int finalize(api_status* status, uint64_t original_event_count) { return finalize(status); }
```
olgavrou, line 120: we are not passing in original_event_count to finalize here?

rlclientlib/logger/async_batcher.h
```diff
@@ -77,9 +78,13 @@ namespace reinforcement_learning { namespace logger {
     queue_mode_enum _queue_mode;
     std::condition_variable _cv;
     std::mutex _m;
+    std::mutex _counter_mutex;
     utility::object_pool<utility::data_buffer> _buffer_pool;
     const char* _batch_content_encoding;
     float _subsample_rate;
+    events_counter_status _events_counter_status;
+    uint64_t _event_index = 0;
```
johnhuang01, line 86: Do you really need uint64_t? Can unit32_t be enough?

rlclientlib/schema/v2/Event.fbs
```diff
@@ -10,6 +10,7 @@ table Event {
 
 table BatchMetadata {
     content_encoding: string; //valid values: IDENTITY and DEDUP
+	original_event_count: uint64;
```
johnhuang01, line 13: uint32 should be enough. Right?

rlclientlib/logger/async_batcher.h
```diff
@@ -151,11 +156,28 @@ namespace reinforcement_learning { namespace logger {
       }
     }
 
-    RETURN_IF_FAIL(collection_serializer.finalize(status));
-
+    if (_events_counter_status == events_counter_status::ENABLE) {
+      uint64_t buffer_start_event_number = _buffer_end_event_number;
+      _buffer_end_event_number = evt.get_event_index();
+      uint64_t original_count = (_buffer_end_event_number - buffer_start_event_number);
```
johnhuang01, line 162: Can you rename this variable to original_event_count?

rlclientlib/logger/async_batcher.h
Tparuchuri: As per this conversation, uint64_t seems more reasonable

rlclientlib/schema/v2/Event.fbs
Tparuchuri: As per this conversation https://github.com/VowpalWabbit/reinforcement_learning/pull/448#discussion_r807543777, uint64_t seems more reasonable

rlclientlib/logger/async_batcher.h
ataymano: 50 RPS is quite realistic.
It will give us 50 * 3600 * 24 ~ 4 millions
uint32 limit is 4 billions which means 3 years of traffic with 50 RPS (or less with higher traffic).
it seems like it should be enough in most of the cases but quite close to realistic limits, so maybe let's just do it safe?

rlclientlib/serialization/json_serializer.h
Tparuchuri: original event count is required only in case of Protocol V2 and Interaction HTTP API sender which uses FB serializer in remaining cases we don't need them

rlclientlib/logger/async_batcher.h
Tparuchuri: Yes, after using TEvent&& evt instead of TEvent second argument is no longer required.

rlclientlib/logger/async_batcher.h
johnhuang01: I missed your previous comments. I am fine with the justification.

rlclientlib/model_mgmt/restapi_data_transport.cc
johnhuang01, line 54: Nit: can you add a , after fails?
if the call using HEAD fails, try with GET only once and return the results of GET request call

johnhuang01 (approved):
very minor comment

rlclientlib/generic_event.cc
```diff
@@ -33,6 +33,10 @@ namespace reinforcement_learning {
 
   const char* generic_event::get_app_id() const { return _app_id.c_str(); }  
 
+  uint64_t generic_event::get_event_index() { return _event_index; }
```
ataymano, line 36:
```suggestion
  uint64_t generic_event::get_event_index() const { return _event_index; }
```

rlclientlib/generic_event.h
```diff
@@ -40,6 +40,10 @@ namespace reinforcement_learning {
 
     const char* get_app_id() const;
 
+    uint64_t get_event_index();
```
ataymano, line 43:
```suggestion
    uint64_t get_event_index() const;
```

rlclientlib/logger/async_batcher.h
```diff
@@ -96,15 +101,14 @@ namespace reinforcement_learning { namespace logger {
 
   template<typename TEvent, template<typename> class TSerializer>
   int async_batcher<TEvent, TSerializer>::append(TEvent&& evt, api_status* status) {
-
+    update_event_with_index(std::move(evt));
```
ataymano, line 104: Just thought a bit more - looks like there is still a problem here:
Following sequence:
1) original_count = 1, queue = [{index = 1}]
2) update_event from thread 1:       event with index=2
3) update_event from thread 2:       event with index=3
4) push to queue from thread 2:      queue = [{index=1}, {index=3}]
5) push to queue from thread 1:      queue = [{index=1}, {index=3], {index=2}]
We got events unordered by indices which is bad since it breaks original_count calculation

ataymano (approved):

olgavrou (approved):

PR #454, opened 2022-03-04
refactor: const char* context_json to string_view

olgavrou:
- Replace `const char* context_json` with `string_view context_json` 
- Updated rl.net.live_model.cc and cs bindings to reflect this change
- Updated python bindings to reflect this change
- changed `rl_string_view.h` to include and reuse vw's string view implementation

Next step could be to do the same change for event_id


rlclientlib/multistep.cc
```diff
@@ -2,12 +2,13 @@
...
 namespace reinforcement_learning {
-  void episode_history::update(const char* event_id, const char* previous_event_id, const char* context, const ranking_response& resp) {
+  void episode_history::update(const char* event_id, const char* previous_event_id, string_view context, const ranking_response& resp) {
     _depths[event_id] = this->get_depth(previous_event_id) + 1;
   }
 
-  std::string episode_history::get_context(const char* previous_event_id, const char* context) const {
-    return R"({"episode":{"depth":")" + std::to_string(this->get_depth(previous_event_id) + 1) + "\"}," + std::string(context + 1);
+  std::string episode_history::get_context(const char* previous_event_id, string_view context) const {
+    // is the +1 here important?
```
olgavrou, line 10: @ataymano

include/rl_string_view.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
+
+// Rely on the VW implementation of string_view
+#include "../ext_libs/vowpal_wabbit/vowpalwabbit/vw_string_view.h"
```
jackgerrits, line 4: A public header cant have a local repo dependent include path like this. This wont work when installed. 

It should be either `vowpalwabbit/vw_string_view.h` or `vw_string_view.h` with the appropriate include path made available via the VowpalWabbit::vw dependency

rlclientlib/ranking_event.h
```diff
@@ -61,11 +62,11 @@ namespace reinforcement_learning {
     learning_mode get_learning_mode() const;
 
   public:
-    static ranking_event choose_rank(const char* event_id, const char* context,
+    static ranking_event choose_rank(const char* event_id, string_view context,
```
jackgerrits, line 65: Should all strings be string views?

rlclientlib/ranking_event.h
```diff
@@ -123,11 +124,11 @@ namespace reinforcement_learning {
     const std::string& get_event_id() const;
 
   public:
-    static multi_slot_decision_event request_decision(const std::string& event_id, const char* context,
+    static multi_slot_decision_event request_decision(const std::string& event_id, string_view context,
```
jackgerrits, line 127: Should event_id be a string_view?

olgavrou: see comment in description :)

include/rl_string_view.h
olgavrou: we should change future_compat.h then too? how does it work with that?

rlclientlib/ranking_event.h
jackgerrits: Ah I see. I vote for event_id being done as well, otherwise it feels inconsistent.

olgavrou: yeah I agree but it's too much for one PR I think

include/rl_string_view.h
jackgerrits: Yep, we need to fix that one too.

olgavrou: OK I can fix both together in a separate PR too, how does that sound?

jackgerrits: Sounds good!

bindings/cs/rl.net.native/rl.net.native.vcxproj
```diff
@@ -57,7 +57,7 @@
       <Optimization>Disabled</Optimization>
       <PreprocessorDefinitions>_DEBUG;_CONSOLE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <SDLCheck>false</SDLCheck>
-      <AdditionalIncludeDirectories>$(ProjectDir)..\..\..\include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+      <AdditionalIncludeDirectories>$(ProjectDir)..\..\..\include;$(SolutionDir)\ext_libs\vowpal_wabbit\vowpalwabbit;$(SolutionDir)\ext_libs\vowpal_wabbit\explore;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
```
lokitoth, line 60: Is there a corresponding change that needs to be made to the CMakeLists.txt file here?

lokitoth (approved):

bindings/cs/rl.net.native/rl.net.native.vcxproj
jackgerrits: These transitive includes are automatically added by CMake. For some reason msbuild doesnt want to propagate include directories from included projects

PR #456, opened 2022-03-07
multistep history test

ataymano:


olgavrou (commented):
does the test file also have to be added to the test cmakelist.txt?

ataymano:
> does the test file also have to be added to the test cmakelist.txt?

yes it should - thanks for noticing

olgavrou (approved):

PR #469, opened 2022-03-25
refactor: event id from const char* to string_view

olgavrou:


lokitoth (approved):

rlclientlib/multistep.cc
```diff
@@ -2,27 +2,27 @@
 #include "err_constants.h"
 
 namespace reinforcement_learning {
-  void episode_history::update(const char* event_id, const char* previous_event_id, string_view context, const ranking_response& resp) {
-    _depths[event_id] = this->get_depth(previous_event_id) + 1;
+  void episode_history::update(string_view event_id, string_view previous_event_id, string_view context, const ranking_response& resp) {
+    _depths[std::string(event_id)] = this->get_depth(previous_event_id) + 1;
```
jackgerrits, line 6: This is not great... We get heterogeneous lookup in c++14 for map I believe. Doesn't help us here though

rlclientlib/multistep.cc
```diff
@@ -2,27 +2,27 @@
...
   }
 
-  std::string episode_history::get_context(const char* previous_event_id, string_view context) const {
+  std::string episode_history::get_context(string_view previous_event_id, string_view context) const {
     return R"({"episode":{"depth":")" + std::to_string(this->get_depth(previous_event_id) + 1) + "\"}," + std::string(context.data() + 1);
```
jackgerrits, line 10: Not part of this PR, but this should probably at least use a string stream

jackgerrits (approved):

PR #490, opened 2022-06-23
Add support for logging number of VW instances in versioned_object_pool

byronxu99:
- `versioned_object_pool` now takes an optional `i_trace*` for logging
- It can log number of objects created when `get_or_create()` or `update_factory()` called
- Simplified implementation of the object pool, use `std::function` instead of pointer to factory, and use `unique_ptr` more often instead of just `new`
- Fixed double-free bug in `model_management::model_data`
- Add tests for logging and fix issues with existing tests


rlclientlib/utility/versioned_object_pool.h
```diff
@@ -145,31 +113,48 @@ class versioned_object_pool
...
   {
+    int objects_count = 0;
+    int new_version = 0;
+
     std::lock_guard<std::mutex> lock(_mutex);
-    _impl->return_to_pool(obj);
+    objects_count = _impl->size();
+    new_version = _impl->version() + 1;
+
+    TRACE_DEBUG(
```
ataymano, line 143: Probably can be TRACE_INFO

rlclientlib/utility/versioned_object_pool.h
```diff
@@ -145,31 +113,48 @@ class versioned_object_pool
...
+  std::unique_ptr<TObject, TObjectDeleter> get_or_create()
   {
     std::lock_guard<std::mutex> lock(_mutex);
-    return _impl->get_or_create();
+    // in the deleter function, capture a copy of the version at time of object creation
+    int current_version = _impl->version();
+    auto pool_obj = _impl->get_or_create();
+    int objects_count = _impl->size();
+
+    TRACE_DEBUG(
```
ataymano, line 126: CAn we also add TRACE_INFO for the cases when objects_count got increased?

byronxu99:
Changed TRACE_DEBUG to TRACE_INFO and log if new object created or existing object returned

ataymano (approved):

PR #493, opened 2022-07-08
Remove dependencies on VW headers

byronxu99:
rlclientlib public headers should not include VW headers.

- `future_compat.h` includes VW `future_compat.h` for `HAS_STD14`, `HAS_STD17`, etc. Fix: copy over definitions of these variables. 
- `rl_string_view.h` includes `VW::string_view`. Fix: use `std::string_view` if C++17 or newer, otherwise use `string_view_lite`. 
- Remove VW headers from being included in Nuget package.

### DIFF
--- a/include/rl_string_view.h
+++ b/include/rl_string_view.h
@@ -1,9 +1,15 @@
 #pragma once
 
-// Rely on the VW implementation of string_view
-#include "vw/common/string_view.h"
-
+#ifdef HAS_STD17
 namespace reinforcement_learning
 {
-using string_view = VW::string_view;
+#include <string_view>
+using std::string_view;
 }
+#else
+#include "nonstd/string_view.h"
+namespace reinforcement_learning
+{
+using string_view = nonstd::string_view;
+}
+#endif
--- a/include/rl_string_view.h
+++ b/include/rl_string_view.h
@@ -1,15 +1,15 @@
 #pragma once
 
 #ifdef HAS_STD17
+#  include <string_view>
 namespace reinforcement_learning
 {
-#include <string_view>
 using std::string_view;
 }
 #else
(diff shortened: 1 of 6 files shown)


jackgerrits:
It was an intentional decision to use the VW::common header to avoid code duplication. Is it a blocker to package those headers too in the nuget?

byronxu99:
The Nuget works just fine with packaging VW headers.

The motivation for this change was Rajan pointing out that rlclientlib library public interface shouldn't refer to VW. Particularly that `live_model` is supposed to abstract away VW internals. So exposing VW headers in the Nuget package would run afoul of this design intention.

In my opinion, duplicating 20-ish lines of code is also less inelegant than forming a dependency between the headers of these two libraries.

jackgerrits:
The thought was that the common functionality could grow beyond the two headers as in currently and it would facilitate sharing. But wanting this interface to not contain any reference to VW is reasonable. LGTM

include/rl_string_view.h
```diff
@@ -1,9 +1,17 @@
...
-// Rely on the VW implementation of string_view
-#include "vw/common/string_view.h"
+#include "future_compat.h"
 
+#ifdef HAS_STD17
+#  include <string_view>
 namespace reinforcement_learning
 {
-using string_view = VW::string_view;
+using std::string_view;
```
jackgerrits, line 9: I believe string view lite delegates internally to the STL implementation if it exists so this isn't really necessary unless we push this all the way through and not even include/search/install string-view-lite if c++17 is being used.

byronxu99: You're right. Looks like VW always uses string-view-lite regardless of C++ version, so I'll make it do that here to maintain consistency.

jackgerrits (approved):

PR #497, opened 2022-08-11
fixing bugs in multistep workflow/tools

peterychang:
Fixes to the workflow and the python parser tool.

I wanted to get these in since the entire multistep workflow is broken in master due to a bug in logger_facade

### DIFF
--- a/rlclientlib/live_model_impl.cc
+++ b/rlclientlib/live_model_impl.cc
@@ -518,11 +518,8 @@ int live_model_impl::init_loggers(api_status* status)
 
   // TODO: Use a specific episode message type (for now it is the same with the observation logger, using
   // observation_logger_facade).
-  if (
-    _configuration.get(name::EPISODE_EH_HOST, nullptr) != nullptr
-      ||
-    _configuration.get(name::EPISODE_FILE_NAME, nullptr) != nullptr
-  )
+  if (_configuration.get(name::EPISODE_EH_HOST, nullptr) != nullptr ||
+      _configuration.get(name::EPISODE_FILE_NAME, nullptr) != nullptr)
   {
     // Get the name of raw data (as opposed to message) sender for episodes.
     const auto* const episode_sender_impl =
--- a/rlclientlib/live_model_impl.cc
+++ b/rlclientlib/live_model_impl.cc
@@ -518,7 +518,11 @@ int live_model_impl::init_loggers(api_status* status)
 
   // TODO: Use a specific episode message type (for now it is the same with the observation logger, using
   // observation_logger_facade).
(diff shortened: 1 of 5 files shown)


rlclientlib/logger/logger_facade.cc
```diff
@@ -124,9 +124,9 @@ int interaction_logger_facade::log(const char* episode_id, const char* previous_
 
       std::string event_id(response.get_event_id());
       std::string model_id(response.get_model_id());
-      std::string previous_id_str(previous_id);
+      std::string previous_id_str(previous_id ? previous_id : "");
```
peterychang, line 127: I'm not 100% sure this is the desired behavior? rl_sim sends nullptr for the previous id, which crashed here, but is an empty string the right replacement?

rlclientlib/live_model_impl.cc
```diff
@@ -518,7 +518,7 @@ int live_model_impl::init_loggers(api_status* status)
 
   // TODO: Use a specific episode message type (for now it is the same with the observation logger, using
   // observation_logger_facade).
-  if (_configuration.get(name::EPISODE_EH_HOST, nullptr) != nullptr)
+  if (_configuration.get(name::EPISODE_SENDER_IMPLEMENTATION, nullptr) != nullptr)
```
peterychang, line 521: Which conditions should we be checking exactly? I know only checking for the EH host is wrong since it breaks file sending. But if USE_AZURE_FACTORIES is true then this check could also fail

rlclientlib/logger/logger_facade.cc
olgavrou: I believe that nullptr should be ok here and flatbuffers will translate it to an empty string on deserialization

rlclientlib/live_model_impl.cc
olgavrou: @zwd-ms do you know?

zwd-ms: I think what this should really check is whether the client is sending multistep events. Only multistep needs the episode logger. The if condition here checks if it is multistep, then initialize the episode logger.
So looks like the check should be whether the configuration file has either eventhub or file logger config for episode (meaning it is using multistep), then do the initialization.

rlclientlib/logger/logger_facade.cc
zwd-ms: Previously this leads to `string(nullptr)`, but it is undefined behavior, and looks like some compiler changes make it runtime error now and will crash. So this change should be good.

zwd-ms (approved):

PR #499, opened 2022-08-16
feat: Episode fb type

peterychang:
This would allow a backend joiner to work generically. VW currently drops any payload it doesn't care about in the multistep workflow (ie: events that are not Multistep or Outcome), which is exactly what we want here

### DIFF
--- a/rlclientlib/logger/logger_facade.cc
+++ b/rlclientlib/logger/logger_facade.cc
@@ -126,8 +126,8 @@ int interaction_logger_facade::log(const char* episode_id, const char* previous_
       std::string model_id(response.get_model_id());
       std::string previous_id_str(previous_id ? previous_id : "");
 
-      return _v2->log(episode_id, context, _multistep_serializer.type, _ext_p, _multistep_serializer,
-          status, previous_id_str, flags, action_ids, probabilities, event_id, model_id);
+      return _v2->log(episode_id, context, _multistep_serializer.type, _ext_p, _multistep_serializer, status,
+          previous_id_str, flags, action_ids, probabilities, event_id, model_id);
     }
     default:
       return protocol_not_supported(status);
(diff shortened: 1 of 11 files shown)


rlclientlib/logger/logger_facade.cc
```diff
@@ -351,5 +351,42 @@ int observation_logger_facade::report_action_taken(const char* primary_id, const
       return protocol_not_supported(status);
   }
 }
+
+
+//TODO: Do we need an EPISODE_SECTION for the config? Just use OBSERVATION_SECTION for now
```
zwd-ms, line 356: The config for Episode were incomplete comparing with interaction/observation: https://github.com/VowpalWabbit/reinforcement_learning/blob/b7b33f8f52634644ebf686c5427058ef1132174d/include/constants.h#L22
The missing ones are used here so we'll need to complete those configs first if we don't want to reuse observation.

rlclientlib/schema/v2/Metadata.fbs
```diff
@@ -1,6 +1,6 @@
 namespace reinforcement_learning.messages.flatbuff.v2;
 
-enum PayloadType : ubyte { CB, CCB, Slates, Outcome, CA, DedupInfo, MultiStep }
+enum PayloadType : ubyte { CB, CCB, Slates, Outcome, CA, DedupInfo, MultiStep, Episode }
```
zwd-ms, line 3: @ataymano Perhaps need to organize the naming a bit:
- Episode -> episode id
- Multistep -> interaction event in an episode
- Outcome -> observation event in an episode

zwd-ms (approved):

rlclientlib/schema/v2/MultiStepEvent.fbs
```diff
@@ -13,4 +13,9 @@ table MultiStepEvent {
     deferred_action:bool = false;
 }
 
+table EpisodeEvent {
+    episode_id:string;
```
ataymano, line 17: we already have this id in meta - do we need to duplicate it here?

peterychang: Technically this entire object is a dummy, we could leave it completely empty (if thats allowed). I just put the id in there since it might be useful at some point in time

ataymano (approved):

PR #502, opened 2022-08-22
Pass nullptr to flatbuffers for empty string to preserve legacy behavior

peterychang:


rlclientlib/serialization/payload_serializer.h
```diff
@@ -204,8 +204,10 @@ struct multistep_serializer : payload_serializer<generic_event::payload_type_t::
     std::string context_str(context);
     copy(context_str.begin(), context_str.end(), std::back_inserter(_context));
 
-    auto fb = v2::CreateMultiStepEventDirect(fbb, event_id.c_str(), previous_id.c_str(), &action_ids, &_context,
-        &probabilities, model_id.c_str(), flags & action_flags::DEFERRED);
+    // TODO: we really shouldn't be checking previous_id.empty(). But to preserve behavior
+    //       for compatibility with older builds of VW, keep it in for now.
+    auto fb = v2::CreateMultiStepEventDirect(fbb, event_id.c_str(), previous_id.empty() ? nullptr : previous_id.c_str(),
```
rajan-chari, line 209: what does flatc generated code in CreateMultiStepEventDirect do with the nullptr?

peterychang: I believe it puts a nullptr into the flatbuffer, so that when the object is deserialized you get a null in the object (which we check for)

peterychang: The actual code is
`auto previous_id__ = previous_id ? _fbb.CreateString(previous_id) : 0;`

So flatbuffers are nullptr aware (in fact the default for all objects are nullptr)

rajan-chari (approved):

PR #505, opened 2022-09-14
Add audit option

tyclintw:
Add the ability to specify audit output in the client configuration. When the AuditEnabled property is set to true, all audit output will be written to an audit file located at "{AUDIT_OUTPUT_PATH}/{eventId}". It is up to the caller to consume this output.

File IO was used to prevent changes to rlclientlib DLL surface. 

Bulk of changes are in vw_model as the audit is vw specific. Previously the vw model forcibly preventing any audit output during json parsing.

### DIFF
--- a/include/constants.h
+++ b/include/constants.h
@@ -3,9 +3,9 @@
 namespace reinforcement_learning
 {
 #ifdef _WIN32
-  const char* const delimiter = "\\";
+const char* const delimiter = "\\";
 #else
-  const char* const delimiter = "/";
+const char* const delimiter = "/";
 #endif
 
 namespace name
--- a/include/constants.h
+++ b/include/constants.h
@@ -3,9 +3,9 @@
 namespace reinforcement_learning
 {
 #ifdef _WIN32
-const char* const delimiter = "\\";
+const char* const PATH_DELIMITER = "\\";
 #else
-const char* const delimiter = "/";
+const char* const PATH_DELIMITER = "/";
 #endif
 
 namespace name
--- a/include/constants.h
+++ b/include/constants.h
@@ -133,7 +133,7 @@ const char* const QUEUE_MODE_BLOCK = "BLOCK";
 const bool DEFAULT_MODEL_BACKGROUND_REFRESH = true;
 const int DEFAULT_VW_POOL_INIT_SIZE = 4;
 const int DEFAULT_PROTOCOL_VERSION = 1;
-const char* DEFAULT_AUDIT_OUTPUT_PATH = "audit";
+const char* const DEFAULT_AUDIT_OUTPUT_PATH = "audit";
 
 const char* get_default_episode_sender();
 const char* get_default_observation_sender();
--- a/include/constants.h
+++ b/include/constants.h
@@ -2,6 +2,12 @@
 
 namespace reinforcement_learning
 {
+#ifdef _WIN32
+  const char* const delimiter = "\\";
+#else
+  const char* const delimiter = "/";
+#endif
+
 namespace name
 {
 const char* const APP_ID = "appid";
@@ -18,6 +24,8 @@ const char* const LEARNING_MODE = "rank.learning.mode";
 const char* const PROTOCOL_VERSION = "protocol.version";
 const char* const HTTP_API_KEY = "http.api.key";
 const char* const HTTP_API_HEADER_KEY_NAME = "http.api.header.key.name";
+const char* const AUDIT_ENABLED = "audit.enabled";
+const char* const AUDIT_OUTPUT_PATH = "audit.output.path";
 
 // Episode
 const char* const EPISODE_EH_HOST = "episode.eventhub.host";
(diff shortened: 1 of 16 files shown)


olgavrou:
@tyclintw there are some google benchmarks in the benchmarks directory, you could extend to include with audit if you want to see the impact of this feature (preferably taking #506 into account)

olgavrou (commented):
it would be nice if this had a unit test in live_model_test, checking for file creation / file path

include/constants.h
```diff
@@ -2,6 +2,12 @@
 
 namespace reinforcement_learning
 {
+#ifdef _WIN32
+const char* const delimiter = "\\";
```
jackgerrits, line 6: Nit: use an uppercase name for constant

rlclientlib/vw_model/safe_vw.cc
```diff
@@ -318,6 +367,24 @@ bool safe_vw::is_CB_to_CCB_model_upgrade(const std::string& args) const
   return local_model_type == mm::model_type_t::CCB && inbound_model_type == mm::model_type_t::CB;
 }
 
+const char* safe_vw::get_audit_data() const
+{
+  if (_vw->audit) { return &(*(_vw->audit_buffer))[0]; }
```
jackgerrits, line 372: I am not certain that `audit_buffer` is a null terminated string. Would be good to either ensure or assert. Better yet this function could return a string_view as you definitely know the size

I think you can do `_vw->audit_buffer.data()`

include/model_mgmt.h
```diff
@@ -74,7 +74,7 @@ class i_model
 {
 public:
   virtual int update(const model_data& data, bool& model_ready, api_status* status = nullptr) = 0;
-  virtual int choose_rank(uint64_t rnd_seed, string_view features, std::vector<int>& action_ids,
+  virtual int choose_rank(const char* event_id, uint64_t rnd_seed, string_view features, std::vector<int>& action_ids,
```
jackgerrits, line 77: Should be a string_view instead of const char*

rlclientlib/vw_model/safe_vw.h
```diff
@@ -40,11 +40,15 @@ class safe_vw
 
   bool is_compatible(const std::string& args) const;
   bool is_CB_to_CCB_model_upgrade(const std::string& args) const;
+  const char* get_audit_data() const;
```
jackgerrits, line 43: Can you add a comment clarifying that a call to `init` or any of the other rank etc functions will invalidate this pointer

rlclientlib/vw_model/vw_model.cc
```diff
@@ -162,6 +173,26 @@ int vw_model::request_multi_slot_decision(const char* event_id, const std::vecto
   }
 }
 
+const std::string vw_model::add_optional_audit_flag(const std::string& command_line) const
```
jackgerrits, line 176: Const is not needed on value based return value
```suggestion
std::string vw_model::add_optional_audit_flag(const std::string& command_line) const
```

rlclientlib/vw_model/vw_model.cc
```diff
@@ -162,6 +173,26 @@ int vw_model::request_multi_slot_decision(const char* event_id, const std::vecto
...
+  if (_audit) { return command_line + " --audit"; }
+
+  return command_line;
+}
+
+void vw_model::write_audit_log(const char* event_id, const char* audit_buffer) const
+{
+  if (event_id != nullptr && audit_buffer != nullptr)
+  {
+    std::string filename = _audit_output_path + reinforcement_learning::delimiter + std::string(event_id);
```
jackgerrits, line 187: Do we need to ensure that this path does not reference some important system path and overwrite it?

include/model_mgmt.h
olgavrou: @jackgerrits this is consistent with event_id's in other function signatures, the PR where we changed all event_id's to string_views has yet to be merged

jackgerrits: Ah okay, I thought it had been merged.

rlclientlib/vw_model/vw_model.cc
ataymano: +1. Also, there is no limitation on entity_id characters - some of them can be problematic for paths ('/','\',...)

tyclintw: The path itself can be define by the caller to it's up to them to use an appropriate path. I will add path sanitization.

rlclientlib/vw_model/vw_model.cc
```diff
@@ -162,6 +173,32 @@ int vw_model::request_multi_slot_decision(const char* event_id, const std::vecto
...
+std::string vw_model::add_optional_audit_flag(const std::string& command_line) const
+{
+  if (_audit) { return command_line + " --audit"; }
+
+  return command_line;
+}
+
+void vw_model::write_audit_log(const char* event_id, string_view audit_buffer) const
+{
+  if (event_id != nullptr && audit_buffer.size() > 0)
```
jackgerrits, line 185:
```suggestion
  if (event_id != nullptr && !audit_buffer.empty())
```

olgavrou (approved):

PR #522, opened 2022-10-24
build: build and package cross platform dotnet core bindings

jackgerrits:
- Version number is monotonically increasing per run, but we can certainly change this.
  - Commit hash which produced the nuget is stored in the nuget in the nuspec file
- Symbols are available as artifacts for the run under the name `symbols`
- Binaries in the package are built with symbols but those are stripped and placed in the separate `symbols` artifact to not bloat the nuget package size.
- rename `rl.net.native` to `rlnetnative` since with .net framework if the dll name contains a period it will not append .dll in the dllimport attribute
  - See relevant issue: https://github.com/dotnet/runtime/issues/7223

TODO:
- [x] Flesh out test stage some more


nuget/dotnet/rl.net.nuspec
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0"?>
+<package>
+  <metadata>
+    <title>RL.Net</title>
```
olgavrou, line 4: since the assembly name changed to `rl.net` do we want this to change too?

jackgerrits: The namespace of the package itself is still Rl.Net so I think it is more consistent to keep it as is here. The assembly name itself was changed to lower case as not all filesystems are case sensitive and now that this is cross platform that was proving troublesome.

.github/workflows/dotnet_nugets.yml
```diff
@@ -0,0 +1,172 @@
+# Changing this name will cause github.run_number to be reset and affect the
```
byronxu99, line 1: rlclientlib has the same version number script as VW - located at `.scripts/version_number.py` - so we should be able to generate a consistently increasing semver number without reference to `github.run_number`

nuget/dotnet/rl.net.nuspec
```diff
@@ -0,0 +1,30 @@
...
+
+    <dependencies>
+      <group targetFramework="netstandard2.0">
+        <dependency id="NETStandard.Library" version="[2.0.0,)" />
+      </group>
+    </dependencies>
+
+  </metadata>
+  <files>
+    <file src="rl.net.dll" target="lib/netstandard2.0" />
```
byronxu99, line 23: `lib\netstandard2.0` directory is  justfor .NET framework, I believe. It may be the case that .NET Core will magically fall back onto the contents here if it can't find the right directories.

.NET Core should be
* `ref\netstandard2.0` compile-time reference containing managed DLL
* `runtimes\{platform}\lib\netstandard2.0` runtime managed DLL, identical copies for each platform
* `runtimes\{platform}\native` runtime native library that's different for each platform

When I was working on the VW nuget, I also ran into issues with using forward slashes in the .nuspec file. It went away when I changed everything to backslashes. Just a heads up if paths aren't behaving as expected.

.github/workflows/dotnet_nugets.yml
jackgerrits: This repo does not use tagged versions therefore that version script is not going to produce a version which can be picked up by nuget as something newer. The incrementing run number is part of the semver prerelease info and therefore doesnt participate in versioning

.github/workflows/dotnet_nugets.yml
```diff
@@ -0,0 +1,172 @@
...
+      fail-fast: false
+      matrix:
+        config:
+          - { os: "windows-latest", runtime_id: "win-x64", vcpkg_target_triplet: "x64-windows-static" }
+          - { os: "ubuntu-latest", runtime_id: "linux-x64", vcpkg_target_triplet: "x64-linux" }
+          - { os: "macos-latest", runtime_id: "osx-x64", vcpkg_target_triplet: "x64-osx" }
+    runs-on: ${{matrix.config.os}}
+    steps:
+      - uses: actions/checkout@v2
+      - run: |
```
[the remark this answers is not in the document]
jackgerrits: I intentionally don't want to do a recursive checkout because it is very slow and not needed by the build.

byronxu99 (approved):

PR #523, opened 2022-10-27
Add local joining and training for federated learning

byronxu99:
Can't figure out how to test the joiner itself, but it mostly just calls functions from external_parser.

~One test is currently failing, still looking into it~ Fixed


rlclientlib/federation/vw_local_joiner.cc
```diff
@@ -0,0 +1,191 @@
...
+      auto* ex_to_return = examples.back();
+      examples.pop_back();
+      VW::finish_example(*_joiner_workspace, *ex_to_return);
+      continue;
+    }
+    VW::setup_examples(*_joiner_workspace, examples);
+
+    // We must remove the trailing newline example.
+    auto* newline_ex = examples.back();
+    examples.pop_back();
```
ataymano, line 102: Aren't we concatenating multiple multi_exs into single one here?

rlclientlib/federation/vw_local_joiner.cc
```diff
@@ -0,0 +1,191 @@
...
+  std::string cmd_str = config.get(name::MODEL_VW_INITIAL_COMMAND_LINE,
+      "--cb_explore_adf --driver_output_off --epsilon 0.2 --power_t 0 -l 0.001 --cb_type mtr -q :: "
+      "--preserve-performance-counters");
+  auto cmd_list = VW::split_command_line(cmd_str);
+  auto options = VW::make_unique<VW::config::options_cli>(cmd_list);
+  _joiner_workspace = VW::initialize_experimental(std::move(options));
+  _joiner = VW::make_unique<example_joiner>(_joiner_workspace.get());
+  _joiner->set_problem_type_config(messages::flatbuff::v2::ProblemType_CB);
+  _joiner->set_learning_mode_config(messages::flatbuff::v2::LearningModeType_Online);
+  _joiner->set_reward_function(messages::flatbuff::v2::RewardFunctionType_Earliest);
```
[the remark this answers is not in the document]
byronxu99: Added the config options themselves

rlclientlib/federation/vw_local_joiner.cc
byronxu99: This was from Jack's code I believe, maybe he can help clarify here.

jackgerrits: Yeah it seems like it could be alexey.  But I don't remember what's going on on the other end here. I don't think I am using the driver so newline examples are not relevant. I would not take the code as it was as correct since it was a very quick prototype

ataymano: I see. @byronxu99, maybe we can add test that is checking this? like adding to interactions with different ids and validating that we have 2 multi_ex on other end (now it looks like we will get single one).

rlclientlib/federation/joined_log_provider.h
```diff
@@ -26,6 +27,11 @@ struct i_joined_log_provider
 {
   virtual ~i_joined_log_provider() = default;
 
+  // Add events to the joiner to be processed when invoke_join() is called.
+  // The contents of the binary data buffer is implementation dependent.
+  RL_ATTR(nodiscard)
+  virtual int add_events(const std::vector<std::shared_ptr<utility::data_buffer>>& events, api_status* status = nullptr) = 0;
```
jackgerrits, line 33: The interface of the log provider should not have the capacity to add logs to it. How the logs are produced should be an implementation detail of the concrete class which implements this interface

rlclientlib/federation/vw_local_joiner.cc
```diff
@@ -0,0 +1,222 @@
...
+    // TODO handle error
+    // Can't return error code in constructor
+  }
+
+  std::string cmd_str = config.get(name::MODEL_VW_INITIAL_COMMAND_LINE,
+      "--cb_explore_adf --driver_output_off --epsilon 0.2 --power_t 0 -l 0.001 --cb_type mtr -q :: "
+      "--preserve-performance-counters");
+  auto cmd_list = VW::split_command_line(cmd_str);
+  auto options = VW::make_unique<VW::config::options_cli>(cmd_list);
+  _joiner_workspace = VW::initialize_experimental(std::move(options));
```
jackgerrits, line 29: Does it make sense for the joiner to need a `Workspace` instance? I think we should probably avoid this

rlclientlib/federation/vw_local_joiner.cc
```diff
@@ -0,0 +1,222 @@
...
+
+  // clear output state
+  chunk_reader.reset(nullptr);
+  _output_example_buffer = std::make_shared<std::vector<char>>();
+
+  // output the next example in the batch
+  if (!_examples.empty())
+  {
+    io_buf io_writer;
+    VW::details::cache_temp_buffer temp_buffer;
```
jackgerrits, line 182: Rather than using cache format I had it my head that this reader would expose binary log based examples that had been EUD joined so that the consumer of this reader could just make the output available as the input to a driver

rlclientlib/federation/vw_trainable_model.cc
```diff
@@ -33,6 +33,24 @@ void learn_and_finish_examples(VW::workspace& vw, VW::multi_ex& examples)
...
+{
+  if (examples.empty()) { return; }
+
+  if (vw.l->is_multiline())
+  {
+    vw.finish_example(examples);
+  }
+  else
+  {
+    for (auto example : examples) { vw.finish_example(*example); }
```
jackgerrits, line 46: vw.finish_example(...) is overloaded for multi ex and should be called directly

byronxu99: I tried it and it didn't work. Turns out `finish_example` does a check for if the learner is multiline or not.

https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/vowpalwabbit/core/src/global_data.cc#L198

jackgerrits: That check is necessary, if you call the incorrect version based on the learner that exists it will fail.

rlclientlib/federation/sender_joined_log_provider.cc
```diff
@@ -0,0 +1,244 @@
...
+  {
+    flatbuffers::FlatBufferBuilder fbb;
+    std::vector<flatbuffers::Offset<messages::flatbuff::v2::JoinedEvent>> events;
+
+    const auto& item = *it;
+    const auto interaction_timestamp = std::get<0>(it->first);
+    const auto interaction_time = interaction_timestamp.to_time_point();
+    if (interaction_time <= eud_cutoff)
+    {
+      const auto reward_cutoff = interaction_time + _eud_offset;
```
ataymano, line 151: Let's think if we need this. I believe existing joiner is treating EUD not as hard cutoff but as "we cannot guarantee for data after EUD to be joined" which is guaranteed by line 149 already.

rlclientlib/federation/vw_trainable_model.cc
```diff
@@ -0,0 +1,441 @@
...
+  }
+}
+}  // namespace
+
+namespace reinforcement_learning
+{
+int trainable_vw_model::create(std::unique_ptr<trainable_vw_model>& output, const utility::configuration& config,
+    i_trace* trace_logger, api_status* status)
+{
+  int protocol_version = config.get_int(name::PROTOCOL_VERSION, 0);
```
ataymano, line 93: kind of nit since it is not affecting final result here, but in other places default version is 1, so it is maybe better to be consistent about it.

rlclientlib/federation/sender_joined_log_provider.cc
```diff
@@ -0,0 +1,244 @@
...
+  if (res->metadata()->content_encoding()->str() != "IDENTITY")
+  { RETURN_ERROR_LS(_trace_logger, status, invalid_argument) << "Can only handle IDENTITY encoding"; }
+
+  for (auto serialized_event : *res->events())
+  {
+    const auto* serialized_payload = serialized_event->payload();
+    const auto* event = flatbuffers::GetRoot<messages::flatbuff::v2::Event>(serialized_payload->data());
+
+    std::string event_id = event->meta()->id()->str();
+    std::vector<uint8_t> joined_event_data(
```
rajan-chari, line 218: Is there a way to avoid mem copy?

rajan-chari: Perhaps a class the holds on to the shared_ptr<buffer> and provides helper methods to access data?

rlclientlib/live_model_impl.cc
```diff
@@ -54,6 +58,10 @@ int live_model_impl::init(api_status* status)
 {
   RETURN_IF_FAIL(init_trace(status));
   RETURN_IF_FAIL(init_model(status));
+
```
rajan-chari, line 61: This seems like an abstraction mismatch.  Perhaps local_loop initialization needs to be refactored from remote_loop init.  local_loop init can then impose conditions on acceptable configurations.

rlclientlib/live_model_impl.cc
```diff
@@ -674,7 +682,30 @@ int live_model_impl::init_model_mgmt(api_status* status)
   // Initialize transport for the model using transport factory
   const auto* const tranport_impl = _configuration.get(name::MODEL_SRC, value::get_default_data_transport());
   m::i_data_transport* ptransport = nullptr;
-  RETURN_IF_FAIL(_t_factory->create(&ptransport, tranport_impl, _configuration, status));
+  RETURN_IF_FAIL(_t_factory->create(&ptransport, tranport_impl, _configuration, _trace_logger.get(), status));
+
+#ifdef RL_BUILD_FEDERATION
+  try
+  {
+    local_loop_controller* local_loop_transport = dynamic_cast<local_loop_controller*>(ptransport);
```
rajan-chari, line 690: Seems like an abstraction mismatch.  See earlier comment on local_loop init.

rajan-chari:
Some notes:
1) Some of the use cases we are discussing with partners limitations on multi threading (single thread only).  In these cases, it seems like exposing a conslidated version of periodic_background_proc .. _proc->run_iteration() for all background threads might be needed.  
2) If it becomes an issue, as it might with in line local joining and training, we may have to change "joiner -> fb -> vw::learn(binary_data)" path to "joiner -> rlclientlib_payload -> vw::learn(examples)"

rajan-chari:
Looks pretty cool!

rajan-chari (approved):
LGTM for merge once all comments are resolved.

examples/rl_sim_cpp/rl_sim.cc
```diff
@@ -602,7 +602,7 @@ std::string rl_sim::create_event_id()
   return oss.str();
 }
 
-rl_sim::rl_sim(boost::program_options::variables_map vm) : _options(std::move(vm)), _loop_kind(CB)
+rl_sim::rl_sim(const boost::program_options::variables_map& vm) : _options(vm), _loop_kind(CB)
```
rajan-chari, line 605: Out of scope for this PR... but would be good to get rid boost::program_options at some point since VW is almost free.

include/constants.h
```diff
@@ -26,6 +26,7 @@ const char* const HTTP_API_KEY = "http.api.key";
 const char* const HTTP_API_HEADER_KEY_NAME = "http.api.header.key.name";
 const char* const AUDIT_ENABLED = "audit.enabled";
 const char* const AUDIT_OUTPUT_PATH = "audit.output.path";
+const char* const EUD_DURATION = "eud.duration";
```
olgavrou, line 29: I feel like these belong in their own federated learning constants file

olgavrou: that should also not be available if not compiled with federated learning compile time flag set to ON

rlclientlib/factory_resolver.cc
```diff
@@ -170,6 +197,13 @@ void factory_initializer::register_default_factories()
         const char* file_name = c.get(name::INTERACTION_FILE_NAME, "interaction.fb.data");
         return file_sender_create(retval, c, file_name, cb, trace_logger, status);
       });
+
+  // Register a default factory for LOCAL_LOOP_SENDER that returns an error
```
olgavrou, line 201: does this also need to go behind an `ifdef`?

rlclientlib/federation/eud_utils.h
```diff
@@ -0,0 +1,64 @@
...
+
+inline int parse_eud(
+    reinforcement_learning::string_view eud_str, std::chrono::seconds& out, reinforcement_learning::api_status* status)
+{
+  std::vector<std::string> components;
+  VW::tokenize(':', eud_str, components);
+  if (components.size() != 3 || std::any_of(components.begin(), components.end(), [](const std::string& component) {
+        return component.empty() || !std::all_of(component.begin(), component.end(), ::isdigit);
+      }))
+  { RETURN_ERROR_ARG(nullptr, status, invalid_argument, "invalid format of eudduration"); }
```
olgavrou, line 49: nit:
```suggestion
  { RETURN_ERROR_ARG(nullptr, status, invalid_argument, "invalid format of eud duration"); }
```

rlclientlib/federation/local_loop_controller.h
```diff
@@ -0,0 +1,69 @@
...
+{
+public:
+  RL_ATTR(nodiscard)
+  static int create(std::unique_ptr<local_loop_controller>& output, const utility::configuration& config,
+      i_trace* trace_logger = nullptr, api_status* status = nullptr);
+
+  // Get model data in binary format
+  // This will perform joining and training on any observed events, and then return the updated model
+  RL_ATTR(nodiscard)
+  virtual int get_data(model_management::model_data& data, api_status* status = nullptr) override;
```
olgavrou, line 33: if this will perform joining and training and then return the updated model it should have a name that indicates that it does so

olgavrou: never mind, I see you are overriding `i_data_transport`

include/constants.h
byronxu99: I think it's simpler to have all the configuration options in one file.

I've pushed a change to put most of the options in an `#ifdef` with the federated learning flag.

rlclientlib/factory_resolver.cc
byronxu99: This is just to print out a more helpful error message when the option is used incorrectly, so I'm putting it outside of the ifdef. That way, you'll still get the error message if you load a config into a rlclientlib compiled without FL.

PR #524, opened 2022-10-31
ignore: Debug dotnet nuget

jackgerrits:


jackgerrits:
For future reference -DCMAKE_MSVC_RUNTIME_LIBRARY=MultiThreaded needs to be changed for debug I think

PR #529, opened 2022-11-04
build nugets on tag push

jackgerrits:


bassmang (approved):

.github/workflows/dotnet_nugets.yml
```diff
@@ -6,6 +6,8 @@ name: .NET Nugets
 on:
   workflow_dispatch:
   push:
+    tags:
+      - '*'
```
byronxu99, line 10: The VW version checks that the tag looks like a SemVer number. Do you think this is also good to have here?
```
    tags:
      - '[0-9]+.[0-9]+.[0-9]+'
      - '[0-9]+.[0-9]+.[0-9]+-*'
```

byronxu99:
Shouldn't we also build the native nuget upon each new tag?
https://github.com/VowpalWabbit/reinforcement_learning/blob/master/.github/workflows/build_nuget.yml

.github/workflows/dotnet_nugets.yml
jackgerrits: I don't think we expect to use different tags, so it doesn't feel necessary to me

byronxu99 (approved):

PR #533, opened 2022-11-09
Draft impl of local joining

jackgerrits:


rlclientlib/federation/sender_joined_log_joiner.h
```diff
@@ -0,0 +1,248 @@
+#pragma once
```
byronxu99, line 1: shouldn't this be split into separate header and source files?

rlclientlib/federation/sender_joined_log_joiner.h
```diff
@@ -0,0 +1,248 @@
...
+  uint32_t size = buffer.size();
+  output.reserve(output.size() + 4);
+  output.insert(std::end(output), reinterpret_cast<uint8_t*>(&size), reinterpret_cast<uint8_t*>(&size)+4);
+
+  output.reserve(output.size() + size);
+  output.insert(std::end(output), buffer.data(), buffer.data()+size);
+  return 0;
+}
+
+struct sender_joined_log_provider : public i_joined_log_provider, i_sender
```
byronxu99, line 112: does `sender_joined_log_provider` need to inherit `i_sender`? because we're only sending to it through a proxy object that contains a pointer to the joined_log_provider.

jackgerrits: Yeah it doesn't need it. Will remove

rlclientlib/federation/sender_joined_log_provider.cc
```diff
@@ -0,0 +1,222 @@
...
+      _interactions.erase(it++);
+    }
+    else
+    {
+      // TODO make sure this is correct. As in does the timestamp ordering mean that all further events are within eud.
+      // break;
+      ++it;
+    }
+
+    auto joined_payload = reinforcement_learning::messages::flatbuff::v2::CreateJoinedPayloadDirect(fbb, &events);
```
byronxu99, line 172: should this part go inside the if statement?

jackgerrits: Good catch definitely should

PR #535, opened 2022-11-10
Introduce backoff if http_transport_client receives 429 response, tidy retry task scheduling.

TrevorChristman:
Today http_transport_client immediately retries its connection attempt if the REST attempt is unsuccessful for any reason.  Recently Personalizer was modified to quickly return 429 (too many requests) response if the server is overloaded.  Also EventHub has always supported 429 responses.  Immediate retries means that http_transport_client will quickly exhaust all its configured retry attempts if the foreign server is under heavy load.

This change modifies the http_request's onComplete handler to sleep for a second before retry if the HTTP request fails.   I reworked the retry logic accept a configured maximum retry timespan, and now all failures retry until either the retry attempts or retry timespan is exhausted.

I also modified http_transport_client to participate better in the ppl task ecosystem.  The old code did an inline retry and blocked inside the original task's onComplete while waiting for the results retry.  This change kicks off the retry as a replacement task.

Today destructing RL.net model objects blocks until all training data has been submitted.  If clients use a very high retry count then destruction might block for a long time.  This was true before my changes so I didn't address it here.  Probably we should plumb some kind of cancellation request and queue status notifications around the training data submission pipeline. 

TESTING DONE
-------------------------------------------------------------------
Used a debugger to confirm correct http_transport_client behavior in cases where:
  - first submission is successful
  - submission fails until retries exhausted
  - submission successful after a few retries

Existing HTTP unit test suite covers retry.


rlclientlib/logger/http_transport_client.h
```diff
@@ -126,45 +154,60 @@ pplx::task<web::http::status_code> http_transport_client<TAuthorization>::http_r
...
+      // We have exhausted retry attempts, log and return the task describing the failure
+      api_status status;
+      auto msg = u::concat("(expected 201): Found ", response_code, ", failed after ", try_count, " retries.");
+      api_status::try_update(&status, error_code::http_bad_status_code, msg.c_str());
+      ERROR_CALLBACK(_error_callback, status);
+
+      return response_task;
+    }
+
+    if (status_codes::TooManyRequests == response_code)
```
lalo, line 191: do we care about 503?

rlclientlib/logger/http_transport_client.h
```diff
@@ -66,7 +66,13 @@ class http_transport_client : public i_sender
     int join();
 
   private:
-    pplx::task<web::http::status_code> send_request(size_t try_count);
+    // repeat request until success or _max_retries attempted
+    // returns the http::status_code of the final attempt.
+    pplx::task<web::http::status_code> send_request();
```
lalo, line 71: is this comment correct?

rlclientlib/logger/http_transport_client.h
```diff
@@ -66,7 +66,13 @@ class http_transport_client : public i_sender
...
+
+    // repeat request until success or _max_retries attempted
+    // returns the http_reponse of the final attempt.
+    pplx::task<http_response> send_request_with_retries(size_t try_count);
```
lalo, line 75: not very familiar with this code but line 103 of this file seems to have a queue of http_request_task, does that mean that we have separate count of retries for each task?

rlclientlib/logger/http_transport_client.h
```diff
@@ -126,45 +154,60 @@ pplx::task<web::http::status_code> http_transport_client<TAuthorization>::http_r
...
+
+    if (status_codes::TooManyRequests == response_code)
+    {
+      // if server says it's too busy, delay retry for a while to give it time to recover
+      static int const LINEAR_DELAY_MS_PER_FAILURE = 2000;
+      int delay_ms = (1 + try_count) * LINEAR_DELAY_MS_PER_FAILURE;
+
+      TRACE_ERROR(_trace, u::concat("HTTP request failed with ", response_code, ", retrying in ", delay_ms, "ms..."));      
+
+      concurrency::wait(delay_ms);
```
byronxu99, line 199: CI is giving an error for `concurrency::wait`. Missing a header?

rlclientlib/logger/http_transport_client.h
TrevorChristman: I filtered for 429 response trying to make the minimal change that would be effective for my use case.

There's dozens of things that could go wrong for any HTTP request, so if DNS is down, or network is down, or servers are down, probably doing an incremental backoff in retries would be sensible.

However, I'm more concerned against introducing large delays in this codepath, since there isn't a good cancellation or status route.  I'd prefer to avoid using large delays for all failure cases until cancel/status is handled better.

rlclientlib/logger/http_transport_client.h
TrevorChristman: Weird, it built locally for me in visual studio.  Will diagnose what's up.

olgavrou: looks like agents async library isn't cross platform but only available for windows

pplx is included via the cpprestsdk is there maybe something from there that could be re-used?

byronxu99: would standard library functions work here?
```cpp
#include <thread>
#include <chrono>
std::this_thread::sleep_for(std::chrono::milliseconds(delay_ms));
```

rlclientlib/logger/http_transport_client.h
lalo: yeah, I called out 503 since it is also related to overloaded server but it also depends on the specific server impl.

rlclientlib/logger/http_transport_client.h
TrevorChristman: @olgavrou, @byronxu99, thank you for this advice.

I dug around a bit and it seems that PPL is not intended to be cross platform, but cpprest does intend to be cross-platform.  cpprest ported to a cross platform pplx the minimal subset which would let them be successful.  In windows pplx just forwards to PPL, so PPL wait() was available during my testing.  In linux I can't see that pplx contains any wait functionality.

@byronxu99 using std sleep will effectively delay my program but it pauses the thread.  It doesn't assign the thread to go contribute to alternate PPLX tasks.  I guess we're not using PPLX deeply inside reinforcement_learning except for the surfaces that touch cpprest, so maybe it's not a big loss.

Let me go research for a while and see if there's any other options besides sleep_for.

rlclientlib/logger/http_transport_client.h
TrevorChristman: @lalo I reworked the retry logic accept a configured maximum retry timespan, and now all failures retry until either the retry attempts or retry timespan is exhausted.

For my 429 case I'll ask for 99999 retries and give a large time budget instead.  For everyone else they can either keep the existing behavior or opt in to the new timeboxed retries.

rlclientlib/logger/http_transport_client.h
TrevorChristman: Yes.  send_request() returns the status code field of the response object, send_request_with_retries() returns the full response object.

rlclientlib/logger/http_transport_client.h
TrevorChristman: http_transport_client manages up to N concurrent upload tasks.  Each individual concurrent upload is a launched pplx::task and is tracked in the collection of task objects you saw.

The current retry count is stored inside http_transport_client::http_request_task, which is the individual object created to track an individual concurrent upload.

rlclientlib/logger/http_transport_client.h
TrevorChristman: It appears to me that cpprest and PPLX are currently end-of-life, and don't contain a taskpool aware sleep or yield in their crossplatform library.

I couldn't find any better option than using std::this_thread::sleep to delay the HTTP retry.  The one-second delay I inserted is on the same order as a web request, and since PPLX isn't used much throughout reinforcement_learning we probably won't see much performance difference.  

Instead of sleeping for increasingly long periods of time I reworked the retry logic accept a configured maximum retry timespan, and now all failures retry until either the retry attempts or retry timespan is exhausted.

rlclientlib/logger/http_transport_client.h
```diff
@@ -66,15 +71,23 @@ class http_transport_client : public i_sender
...
+    pplx::task<http_response> send_request_with_retries(size_t try_count);
 
     i_http_client* _client;
     http_headers _headers;
     buffer _post_data;
 
     pplx::task<web::http::status_code> _task;
 
-    size_t _max_retries = 1;
+    std::chrono::time_point<std::chrono::system_clock> _start_time = std::chrono::system_clock::now();
```
byronxu99, line 88: shouldn't initialization to `system_clock::now()` be done in the constructor instead of right here?

rlclientlib/logger/http_transport_client.h
```diff
@@ -126,45 +162,55 @@ pplx::task<web::http::status_code> http_transport_client<TAuthorization>::http_r
...
     }
 
-    // If the response is not the expected code then it has failed. Retry if possible otherwise report background
-    // error.
-    if (code != status_codes::Created && code != status_codes::NoContent)
+    bool success = (response_code >= status_codes::OK) && (response_code < status_codes::MultipleChoices);
+    if (success) { return response_task; }
+
+    // If the response is not success class then it has failed. Retry if possible otherwise report background error.
+    auto deadline = _start_time + _max_retry_duration;
```
byronxu99, line 184: would this overflow when `_max_retry_duration = std::chrono::milliseconds::max()`?

rlclientlib/logger/http_transport_client.h
TrevorChristman: You are correct.  I removed the default assign in the latest diff.

byronxu99: Did you forget to commit/push something? I'm still seeing the call to `now()` both right here and again in the constructor. This line should be something like
```
std::chrono::time_point<std::chrono::system_clock> _start_time = 0
```

byronxu99 (approved):

lalo (approved):

PR #545, opened 2023-01-23
add schema for model update

jackgerrits:


homezcx (approved):

rlclientlib/schema/v2/ModelUpdateEvent.fbs
```diff
@@ -0,0 +1,10 @@
+﻿namespace reinforcement_learning.messages.flatbuff.v2;
+
+table ModelUpdate {
+    // VW ModelDelta
+    delta: [ubyte];
```
rajan-chari, line 5: Is this not just a collection of (hash, float) values?

jackgerrits: No a VW ModelDelta is pretty much a full VW model. This is to adhere to the model merging operations defined in VW.

rajan-chari: To be specific, my concern is around versioning.  If this is an opaque payload that changes over time, we would then have to worry about cases where the sender and reciever might be out of sync. Where as being specific about schema means we can simply send a different message that's either handled or not.

rajan-chari: Ah ok.  makes sense.  Thanks

rajan-chari (approved):
Looks good thanks.

rlclientlib/schema/v2/ModelUpdateEvent.fbs
jackgerrits: This work is still experimental, so I have we can update to consume the schematized VW model so we won't need to worry.

For now though it follows the current "status quo" of VW models, so its version is embedded within it and as long as the format doesn't change should be loadable.

PR #547, opened 2023-01-31
Add C# binding to Episodic RL interface

homezcx:
This PR adds the C# binding for the Episodic RL interface.  The usage example is implemented in the test as well:

```
        [TestMethod]
        public void Test_LiveModel_RequestEpisodicE2E()
        {
            LiveModel liveModel = this.ConfigureLiveModel(PseudoLocConfigJsonVwModel);
            EpisodeState state = new EpisodeState("episode-0");

            state.Update("event1", "event0", ContextJson);

            var response = liveModel.RequestEpisodicDecision("event1", null, ContextJson, ActionFlags.Default, state);
            Assert.AreEqual(0, response.ChosenAction);
        }
```

### DIFF
--- a/bindings/cs/rl.net.cli.test/UnicodeTest.cs
+++ b/bindings/cs/rl.net.cli.test/UnicodeTest.cs
@@ -30,7 +30,7 @@ namespace Rl.Net.Cli.Test
         const string PseudoLocJsonKey4 = "ℓôřú₥ ïƥƨú₥ δôℓôř";
         const string PseudoLocJsonValue4 = "áβç { δèƒ: ϱλï }";
 
-        const string PseudoLocConfigJson =
+        const string PseudoLocConfigJsonPdfModel =
 @"{
     ""ApplicationID"": ""ßïϱTèƨƭÂƥƥℓïçáƭïôñNá₥è-ℓôř"",
     ""IsExplorationEnabled"": true,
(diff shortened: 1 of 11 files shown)


rlclientlib/multistep.cc
```diff
@@ -5,7 +5,7 @@
 namespace reinforcement_learning
 {
 void episode_history::update(
-    const char* event_id, const char* previous_event_id, string_view context, const ranking_response& resp)
```
ataymano, line 8: Have you removed it since it is not used?

homezcx: yes. Found it unused.

ataymano: The reason why it was there is the following:
Although right now history is just single trivial class it supposed to be interface + multiple implementations. And interface (in order to be generic enough) should depend on response (since it is also part of history).
I am fine with removing it for now, but just wanted to emphasize that it will probably get back once we make progress in that direction.

homezcx: If you want to let me add it back in this PR or later, that is fine for me. I was thinking it should be the episode history being sent to decision stacks, not the ranking_response being sent to the history to be altered.

 For this topic discussion, I sent one email earlier "[Episodic Learning] Episode history design". If you could comment on that, it will be helpful.

bindings/cs/rl.net.cli.test/UnicodeTest.cs
```diff
@@ -296,6 +311,10 @@ private LiveModel ConfigureLiveModel()
             TempFileDisposable observationDisposable = new TempFileDisposable();
             this.TestCleanup.Add(observationDisposable);
 
+            TempFileDisposable episodenDisposable = new TempFileDisposable();
```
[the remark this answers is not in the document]
homezcx: Thanks. Fixed.

bindings/cs/rl.net.cli.test/UnicodeTest.cs
```diff
@@ -502,11 +532,50 @@ private void Run_LiveModelRequestContinuousAction_Test(LiveModel liveModel, stri
         [TestMethod]
         public void Test_LiveModel_RequestContinuousAction()
         {
-            LiveModel liveModel = this.ConfigureLiveModel();
+            LiveModel liveModel = this.ConfigureLiveModel(PseudoLocConfigJsonPdfModel);
 
             Run_LiveModelRequestContinuousAction_Test(liveModel, PseudoLocContextJsonWithPdf);
         }
 
+        private void Run_LiveModelRequestEpisodicDecisionWithFlags_Test(LiveModel liveModel, string eventId, string previousEventId, string contextJson, ActionFlags flags, EpisodeState episodes)
```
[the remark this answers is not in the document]
homezcx: Thanks. Fixed.

rlclientlib/multistep.cc
ataymano: Sorry for delay, can we actually keep ranking_response in update signature in order not to forget this dependency?

ataymano (approved):

PR #554, opened 2023-02-07
remove explore only from rank call, and send first rank call to vw model

olgavrou:
added unit test for rank call and for rank call with LAS

rank for mock function made to return a basic pdf since explore only is not accessed anymore

### DIFF
--- a/rlclientlib/live_model_impl.cc
+++ b/rlclientlib/live_model_impl.cc
@@ -83,12 +83,18 @@ int live_model_impl::choose_rank(
 
   // check arguments
   RETURN_IF_FAIL(check_null_or_empty(event_id, context, _trace_logger.get(), status));
-  if (!_model_ready)
-  {
-    RETURN_IF_FAIL(explore_only(event_id, context, response, status));
-    response.set_model_id("N/A");
-  }
-  else { RETURN_IF_FAIL(explore_exploit(event_id, context, response, status)); }
+
+  // The seed used is composed of uniform_hash(app_id) + uniform_hash(event_id)
+  const uint64_t seed = VW::uniform_hash(event_id, strlen(event_id), 0) + _seed_shift;
+
+  std::vector<int> action_ids;
+  std::vector<float> action_pdf;
+  std::string model_version;
+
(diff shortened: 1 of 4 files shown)


unit_test/mock_util.cc
```diff
@@ -69,10 +69,17 @@ std::unique_ptr<fakeit::Mock<m::i_model>> get_mock_model(m::model_type_t model_t
 {
   auto mock = std::unique_ptr<Mock<m::i_model>>(new fakeit::Mock<m::i_model>());
 
-  const auto choose_rank_fn = [](const char*, uint64_t, r::string_view, std::vector<int>&, std::vector<float>&,
-                                  std::string& model_version, r::api_status*)
+  const auto choose_rank_fn = [](const char*, uint64_t, r::string_view, std::vector<int>& actions,
+                                  std::vector<float>& scores, std::string& model_version, r::api_status*)
   {
     model_version = "model_id";
+    actions.resize(2);
```
olgavrou, line 76: some tests mock out the model rank call to this function
before the removal of explore_only the model would not get called and explore_only would populate the pdf

with that removed this function will get called when model rank is called, so mocking a pdf (no tests actually check what it is set to, only that is sums to 1)

ataymano (approved):

rlclientlib/live_model_impl.cc
```diff
@@ -83,12 +83,19 @@ int live_model_impl::choose_rank(
...
-  {
-    RETURN_IF_FAIL(explore_only(event_id, context, response, status));
-    response.set_model_id("N/A");
-  }
-  else { RETURN_IF_FAIL(explore_exploit(event_id, context, response, status)); }
+
+  // The seed used is composed of uniform_hash(app_id) + uniform_hash(event_id)
+  const uint64_t seed = VW::uniform_hash(event_id, strlen(event_id), 0) + _seed_shift;
+
+  std::vector<int> action_ids;
```
rajan-chari, line 90: It's unfortunate that we don't know the size here so we can potentially avoid multiple allocations as this vector grows.

rajan-chari (approved):
Looks good and probably long overdue.  Thanks!

PR #555, opened 2023-02-07
ci: benchmark live_model init

bassmang:


rajan-chari:
Wondering if the models should be called cb_adf_half ... etc.

benchmarks/benchmark_init.cc
```diff
@@ -0,0 +1,94 @@
...
+  int res[sizeof...(extra_args)] = {extra_args...};
+  auto load_model = res[0];
+  auto model_id = res[1];
+
+  std::ostringstream model_path;
+  model_path << "benchmarks/models/";
+  switch (model_id)
+  {
+    case 0:
+      model_path << "cb_small.m";
```
rajan-chari, line 38: Wondering if adding the model command line in a comment here would be useful as reference.

benchmarks/benchmark_init.cc
```diff
@@ -0,0 +1,94 @@
...
+    config.set(r::name::MODEL_FILE_NAME, model_path.str().c_str());
+  }
+  else { config.set(r::name::MODEL_SRC, r::value::NO_MODEL_DATA); }
+
+  for (auto _ : state)
+  {
+    r::api_status status;
+    r::live_model model(config);
+    model.init(&status);
+    if (status.get_error_code() != err::success)
```
rajan-chari, line 77: Should we have this test in all the other perf tests?

benchmarks/benchmark_init.cc
```diff
@@ -0,0 +1,94 @@
...
+    }
+    benchmark::ClobberMemory();
+  }
+}
+
+// characteristics of the benchmark examples that will be generated are:
+// x load model (on/off)
+// x model path
+
+BENCHMARK_CAPTURE(bench_init, no_model, false, 0)->Unit(benchmark::kMillisecond);
```
rajan-chari, line 91: I am just curious.  Is there a way to specify minimum number of iterations or time a test should run?

rajan-chari (approved):
Looks good.  Left some comments.  Thanks.

bassmang:
> Wondering if the models should be called cb_adf_half.m ... etc. instead of cb_half.m

fair point, update to `cb_explore_adf...`

benchmarks/benchmark_init.cc
bassmang: added comment, only uses --cb_explore_adf (and -b 18 by default)

benchmarks/benchmark_init.cc
bassmang: this is checked in the other benchmarks

benchmarks/benchmark_init.cc
bassmang: seems like there used to be that functionality, but they removed it because of frequent misconfiguration and only allow min time now

PR #558, opened 2023-02-07
Add EpisodicActionTakenEvent c# binding interface

homezcx:
The full example for calling episodic learning is as below:

```
        [TestMethod]
        public void Test_LiveModel_RequestEpisodicE2E()
        {
            string episodeId = "episode-0";
            string eventId = "event1";
            string previousEventId = "event0";
            LiveModel liveModel = this.ConfigureLiveModel(PseudoLocConfigJsonVwModel);
            EpisodeState state = new EpisodeState(episodeId);
            RankingResponse resp = new RankingResponse();

            state.Update(eventId, previousEventId, ContextJson, resp);

            var response = liveModel.RequestEpisodicDecision(eventId, previousEventId, ContextJson, ActionFlags.Deferred, state);
            Assert.AreEqual(0, response.ChosenAction);

            liveModel.QueueOutcomeEvent(episodeId, 0.5f);
            liveModel.QueueEpisodicActionTakenEvent(episodeId, eventId);
        }
```


olgavrou (approved):

bindings/cs/rl.net/LiveModel.cs
```diff
@@ -229,6 +229,36 @@ public static int LiveModelReportActionTaken(IntPtr liveModel, IntPtr eventId, I
                 return LiveModelReportActionTakenNative(liveModel, eventId, apiStatus);
             }
 
+            [DllImport("rlnetnative", EntryPoint = "LiveModelReportEpisodicActionTaken")]
+            private static extern int LiveModelReportEpisodicActionTakenNative(IntPtr liveModel, IntPtr episodeId, IntPtr eventId, IntPtr apiStatus);
```
ataymano, line 233: Not sure why do we need to expose notion of episodic in feedback methods.
We have generic notion of primary and secondary id which are also used in ccb and slates. Right now there is no per-slot activation for them but it could be in theory and there is no difference in implementation on client side.
So maybe either reuse ccb methods completely or just introduce custom wrappers on very last c# level to adjust signature (eventId/slotId -> episodeId/eventId)?

homezcx: This is to abstract the feedback from the real implementation. You are right that "it could be in theory and there is no difference in implementation on the client side.". So the underlying implementation at C++ layer could be the same.

This is a question of at what layer we want to expose the internal implementation. Since all C# codes are easy to be decompiled, I think matching the entry of the c++ method with the c# wrapper is better.

On the other hand, reusing CCB methods on the Episodic + CB implementation is really confusing to me.

homezcx: Fixed.

bindings/cs/rl.net.native/rl.net.live_model.h
```diff
@@ -87,6 +87,8 @@ extern "C"
 
   API int LiveModelReportActionTaken(
       livemodel_context_t* livemodel, const char* event_id, reinforcement_learning::api_status* status = nullptr);
+  API int LiveModelReportEpisodicActionTaken(livemodel_context_t* livemodel, const char* episode_id,
+      const char* event_id, reinforcement_learning::api_status* status = nullptr);
 
   API int LiveModelReportOutcomeF(livemodel_context_t* livemodel, const char* event_id, float outcome,
```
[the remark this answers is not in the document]
homezcx: Fixed.

ataymano (approved):

PR #559, opened 2023-02-08
[fix] leak fix

ataymano:


rlclientlib/live_model_impl.cc
```diff
@@ -496,64 +492,66 @@ int live_model_impl::init_loggers(api_status* status)
...
   _configuration.set(config_constants::CONFIG_SECTION, config_constants::INTERACTION);
   RETURN_IF_FAIL(_sender_factory->create(
-      &ranking_data_sender, ranking_sender_impl, _configuration, &_error_cb, _trace_logger.get(), status));
+      ranking_data_sender, ranking_sender_impl, _configuration, &_error_cb, _trace_logger.get(), status));
   RETURN_IF_FAIL(ranking_data_sender->init(_configuration, status));
 
   // Create a message sender that will prepend the message with a preamble and send the raw data using the
   // factory created raw data sender
-  l::i_message_sender* ranking_msg_sender = new l::preamble_message_sender(ranking_data_sender);
+  std::unique_ptr<l::i_message_sender> ranking_msg_sender(
```
olgavrou, line 505: `preamble_message_sender` moves the raw pointer given in its constructor into a unique pointer, so maybe instead of `release()`ing here `preamble_message_sender` constructor could also change to accept a unique pointer

olgavrou: same for time provider below (although more interfaces need to change for that)

unit_test/live_model_test.cc
```diff
@@ -1945,8 +1945,6 @@ BOOST_AUTO_TEST_CASE(live_model_using_endpoint_failure_no_uri)
   BOOST_CHECK_EQUAL(_rl->init(&status), r::error_code::http_model_uri_not_provided);
 }
```
olgavrou, line 1947: USE_AZURE_FACTORIES is defaulted to ON so it won't make a difference in the CI but maybe it should be added back in case someone needs to compile without them?

rlclientlib/live_model_impl.cc
ataymano: yeah I believe I left comment about it in one place and didn't want to grow PR too much. let's do it in next one

unit_test/live_model_test.cc
ataymano: yeah good catch - let me fx it

ataymano: actually this flag was in random place - most of the tests in this file are having this dependency as well.
Created an issue: https://github.com/VowpalWabbit/reinforcement_learning/issues/560

PR #562, opened 2023-02-09
Avoid using raw pointers

byronxu99:
Use STL types instead


rlclientlib/factory_resolver.cc
```diff
@@ -106,17 +106,14 @@ int file_model_loader_create(std::unique_ptr<m::i_data_transport>& retval, const
   TRACE_INFO(trace_logger, "File model loader created.");
   const char* file_name = config.get(name::MODEL_FILE_NAME, "current");
   const bool file_must_exist = config.get_bool(name::MODEL_FILE_MUST_EXIST, false);
-  auto* file_loader = new model_management::file_model_loader(file_name, file_must_exist, trace_logger);
+  std::unique_ptr<model_management::file_model_loader> file_loader(
```
jackgerrits, line 109: Could use VW::make_unique if you want to remove the new, but fine as is

jackgerrits (approved):

PR #569, opened 2023-03-09
feat: use lru_dedup_dict for rank call

bassmang:


include/live_model.h
```diff
@@ -107,6 +107,16 @@ class live_model
...
 
+  /**
+   * @brief Load dedup cache.
+   * Load the dedup cache from the specified file.  This cache is used to
+   * prevent duplicate actions from being sent to the online trainer.
+   * @param hash Hash of the dedup cache
+   * @param action_str Action string
+   * @return int Return error code.  This will also be returned in the api_status object
+   */
+  int add_lru_dedup_cache(uint64_t hash, std::string action_str, api_status* status);
```
olgavrou, line 118:
```suggestion
  int load_action(uint64_t action_id, std::string action_str, api_status* status);
```

rlclientlib/lru_dedup_cache.h
```diff
@@ -35,6 +35,7 @@ struct lru_dedup_cache
       void* context = nullptr);
   bool exists(uint64_t dedup_id);
   void clear(release_example_f release_example = lru_dedup_cache::noop_release_example_f, void* context = nullptr);
+  std::unordered_map<uint64_t, VW::example*>* get_dict() { return &dedup_examples; }
```
olgavrou, line 38: it is a struct you can just keep a reference to the `lru_dedup_cache` object and just access the dictionary when needed

rlclientlib/vw_model/safe_vw.cc
```diff
@@ -19,13 +19,14 @@ namespace reinforcement_learning
 {
 static const std::string SEED_TAG = "seed=";
 
-safe_vw::safe_vw(std::shared_ptr<safe_vw> master) : _master(std::move(master))
+safe_vw::safe_vw(std::shared_ptr<safe_vw> master, lru_dedup_cache* dedup_cache)
```
olgavrou, line 22: maybe call it `action_cache` dedup doesn't really mean anything in safe_vw

rlclientlib/vw_model/safe_vw.cc
```diff
@@ -120,6 +122,24 @@ void safe_vw::parse_context_with_pdf(string_view context, std::vector<int>& acti
   for (auto&& ex : examples) { _example_pool.emplace_back(ex); }
 }
 
+void safe_vw::add_lru_dedup_cache(uint64_t hash, std::string action_str)
+{
+  if (_dedup_cache == nullptr) { _dedup_cache = new lru_dedup_cache(); }
```
olgavrou, line 127: is this something that will be passed into `safe_vw` or created by `safe_vw`? it should not be both, ownership should be clear

rlclientlib/vw_model/vw_model.cc
```diff
@@ -67,6 +67,14 @@ int vw_model::update(const model_data& data, bool& model_ready, api_status* stat
   return error_code::success;
 }
 
+int vw_model::add_lru_dedup_cache(uint64_t hash, std::string action_str, api_status* status)
```
olgavrou, line 70: We should probably move the "parse action and populate the cache" functionality into `vw_model` and just have a pointer of it in each vw instance

that way whenever you use vw for a rank call we could just do

```
auto vw = _vw_pool.get_or_create()
vw.set_action_cache(&action_cache)
vw.rank(...)
```

olgavrou: the way it is done here the specific action is only passed into one of the vw instances in the object pool

include/live_model.h
bassmang: updated

rlclientlib/lru_dedup_cache.h
bassmang: updated

rlclientlib/vw_model/safe_vw.cc
bassmang: removed this

rlclientlib/vw_model/vw_model.cc
bassmang: removed lru_cache from safe_vw and constructor etc. Added as an arg to load_action and rank in safe_vw

PR #572, opened 2023-03-22
refactor: improve live_model interface

bassmang:
Build out new interface independent of old interface. Update testing and benchmarking to use new API. Handle bindings and deprecation in next step.


rlclientlib/episodic_loop.cc
```diff
@@ -0,0 +1,47 @@
+#include "episodic_loop.h"
```
rajan-chari, line 1: Episodic api exposes std::vector and std::string through it's public interface.  Let's remove episodic api from the new interface till we can resolve this issue.

include/cb_loop.h
```diff
@@ -0,0 +1,110 @@
...
+   * and then sampling from it.
+   * @param event_id  The unique identifier for this interaction.  The same event_id should be used when
+   *                  reporting the outcome for this action.
+   * @param context_json Contains action, action features and context features in json format
+   * @param resp Ranking response contains the chosen action, probability distribution used for sampling actions and
+   * ranked actions
+   * @param status  Optional field with detailed string description if there is an error
+   * @return int Return error code.  This will also be returned in the api_status object
+   */
+  int choose_rank(const char* event_id, string_view context_json, ranking_response& resp, api_status* status = nullptr);
```
rajan-chari, line 40: I agree with the general reasoning that string types should be passed in with a size to prevent certain failure modes.  For example, certain languages, don't have null terminated strings.  

However, string_view class in our public interface introduces a type with very large surface area.  I'd like to see this replaced with a simple "buffer" like class with a small surface area.  for example buffer or string_buffer struct with a const char* and a size.

rajan-chari: I have 2 primary issues with this:
1) We are introducing a type in our own public interface that we have little control over.
2) A large unnecessary surface area makes usage automated binding tools like SWIG difficult or impossible.

olgavrou: So the requirement for this API change is to provide a C style API aiming mostly at bindings, and not a C++ API, correct?

include/base_loop.h
```diff
@@ -0,0 +1,205 @@
+/**
```
olgavrou, line 1: Author and date update

include/ccb_loop.h
```diff
@@ -0,0 +1,164 @@
...
+#include <memory>
+
+namespace reinforcement_learning
+{
+class ccb_loop : public base_loop
+{
+public:
+  using base_loop::base_loop;
+
+  /**
```
olgavrou, line 30: also remove deprecated from all of the comments

include/model_mgmt.h
```diff
@@ -61,7 +61,8 @@ enum class model_type_t
   CB,
   CCB,
   SLATES,
-  CA
+  CA,
+  EPISODIC
```
olgavrou, line 65: there isn't an episodic model type, it gets translated to cb

include/slates_loop.h
```diff
@@ -0,0 +1,106 @@
...
+#include <functional>
+#include <memory>
+
+namespace reinforcement_learning
+{
+class slates_loop : public base_loop
+{
+public:
+  using base_loop::base_loop;
+
```
olgavrou, line 28: same comment here-> remove RL_DEPRECATED and DEPRECATED mention from comments

unit_test/live_model_test.cc
olgavrou, line 1: we should add at least some basic tests for the existing live_model API, not all of them since both APIs point to the same underlying implementation, but we should at least exercise its usage

rlclientlib/episodic_loop.cc
bassmang: removed episodic

include/ca_loop.h
```diff
@@ -0,0 +1,126 @@
...
+   * @param event_id  The unique identifier for this interaction.  The same event_id should be used when
+   *                  reporting the outcome for this action.
+   * @param context_json Contains context features in json format
+   * @param flags Action flags (see action_flags.h)
+   * @param response Continuous action response contains the chosen action and the probability density value of the
+   * chosen action location from the continuous range.
+   * @param status  Optional field with detailed string description if there is an error
+   * @return int Return error code.  This will also be returned in the api_status object
+   */
+  RL_DEPRECATED("New unified example builder interface is coming")
```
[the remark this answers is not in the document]
bassmang: yes I think we should

include/ccb_loop.h
bassmang: yea removed as well

include/ca_loop.h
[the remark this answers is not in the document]
bassmang: removed

include/model_mgmt.h
bassmang: this can be reverted, we're holding off on episodic for now

include/slates_loop.h
bassmang: removed

unit_test/live_model_test.cc
bassmang: yea good point, I'll keep a test for each loop type using the old live model in a separate file

bassmang: add live_model_test_legacy

include/cb_loop.h
rajan-chari: Some clarification:

With (1) I am primarily concerned with exposing classes with implementation and contracts that we don't have control over and unanticipated consequences of doing that.

Issues in this category are things like someone linking with a precompiled version of our libraries.  For example, someone building in Debug mode and linking with our libraries in Release mode or vice versa.  Classes allocated in client code but passed into our code can have undefined behavior.  Classes defined in the interface may throw under conditions that we are unaware of etc.  So C++ is fine as long as it's our classes.  A buffer classes that holds char * and size is perfectly acceptable and even preferable.  This keeps the spirit in with string_view was introduced intact.  i.e. we should carry the buffer and its length together as a unit.

With (2) The concern is binding with automated binding tools.  I am hoping that we get a number of languages for free with this approach... for example.. javascript bindings would be nice.

For example, neither of these 2 concerns preclude the use of a C++ iterator type interface.  We do implement this somewhere.

include/model_mgmt.h
bassmang: yea I agree, there should be a separate concept of `loop_type` and `model_type`, since the two can be used interchangeably for the other loop/model types

include/base_loop.h
```diff
@@ -0,0 +1,213 @@
...
+class i_model;           //
+}  // namespace model_management
+   //
+namespace utility
+{                     //
+class configuration;  //
+}  // namespace utility
+//////////////////////////////////
+
+struct loop_str
```
jackgerrits, line 45: I don't understand why our C++ interface must become a C interface? 

C++ has worked for many years. If we require a pure C interface that we want to provide for ABI stability and binding convenience then it is trivial to layer it upon this. But by dropping our core interface to C we lose so much expressiveness and productivity.

include/cb_loop.h
jackgerrits: @rajan-chari Please see my comment: https://github.com/VowpalWabbit/reinforcement_learning/pull/572/files#r1153409981

I think there are two issues being combined here and I don't see that value of combining them. Yes, having a ABI stable C interface is useful. But it is also useful to have a C++ interface. Why must they be the same thing? It is trivial to implement the ABI stable C interface on top of the core library being written in C++.

Users then get the choice of C if they need ABI stability, or C++ if that works best for them.

rajan-chari: The objection to my concerns and proposal is not clear.

Is the claim that the right C++ approach is to use string_view and other std types in our C++ interface?

include/base_loop.h
rajan-chari: structs are C++...

Keeping the interface surface area minimal using appropriate types for its intended purpose makes  it a good C++ interface.

Here the intent is to pass in buffer and size as one logical entity.  I think this choice serves that purpose.

include/base_loop.h
```diff
@@ -0,0 +1,213 @@
+/**
```
lalo, line 1: can all the new files be in a separate directory? some sort of organization of the previous api vs this new set

lalo: if we have a directory you can add a README file and add some comments of the reasoning behind this new api.

include/ca_loop.h
```diff
@@ -0,0 +1,111 @@
...
+#include "factory_resolver.h"
+#include "future_compat.h"
+#include "multi_slot_response.h"
+#include "multi_slot_response_detailed.h"
+#include "sender.h"
+
+#include <functional>
+#include <memory>
+
+namespace reinforcement_learning
```
lalo, line 22: should the new api be on the same namespace?

bassmang: I dont see a reason to move it, this namespace is more general than just the live model

include/base_loop.h
```diff
@@ -0,0 +1,214 @@
...
+{                     //
+class configuration;  //
+}  // namespace utility
+//////////////////////////////////
+
+struct str_view
+{
+  str_view(const char* str, int size) : str(str), size(size) {}
+  str_view(const char* str) : str(str), size(str == nullptr ? 0 : strlen(str)) {}
+  const char* str;
```
rajan-chari, line 49: Should the str and size be private?

include/base_loop.h
```diff
@@ -0,0 +1,214 @@
...
+  int size;
```
jackgerrits, line 50: size_t

include/base_loop.h
rajan-chari: On second thought.. making them private might not add much value.

include/base_loop.h
bassmang: fair point, moved to a subdir and added a readme

include/base_loop.h
bassmang: updated

rajan-chari (approved):
Looks good.  Thanks.

PR #573, opened 2023-03-23
Add support for --binary_to_json with multistep.

peterychang:
Converts multistep binary logs to cb dsjson. This functionality should only be used as part of the CFE workflow and the outputs should not be saved


olgavrou:
can we get a test here?

olgavrou (approved):

PR #574, opened 2023-03-24
refactor: move bindings cmake to top level

bassmang:


bassmang:
> What's the reason for this change?

To make the flags more obvious to users, and consistency with using them to determine subdirectories

jackgerrits (approved):

PR #580, opened 2023-04-05
feat: add id and counter

bassmang:


rlclientlib/federation/local_client.cc
```diff
@@ -84,13 +93,14 @@ int local_client::create(std::unique_ptr<i_federated_client>& output, const util
...
-  std::string initial_command_line(config.get(
-      name::MODEL_VW_INITIAL_COMMAND_LINE, "--cb_explore_adf --json --quiet --epsilon 0.0 --first_only --id N/A"));
+  std::string initial_command_line(config.get(name::MODEL_VW_INITIAL_COMMAND_LINE,
+      "--cb_explore_adf --json --quiet --epsilon 0.0 --first_only --id test_app_id"));
 
   // TODO try catch
   auto args = VW::make_unique<VW::config::options_cli>(VW::split_command_line(initial_command_line));
   auto logger = utility::make_vw_trace_logger(trace_logger);
   auto workspace = VW::initialize_experimental(std::move(args), nullptr, nullptr, nullptr, &logger);
+  workspace->id += "/0";  // initialize iteration id to 0
```
jackgerrits, line 103: Does the id already include app id?

rlclientlib/federation/local_client.cc
```diff
@@ -70,6 +72,13 @@ int local_client::report_result(const uint8_t* payload, size_t size, api_status*
       auto view = VW::io::create_buffer_view(reinterpret_cast<const char*>(payload), size);
       auto delta = VW::model_delta::deserialize(*view);
       auto new_model = *_current_model + *delta;
+
+      // Increment iteration id for new workspace
+      int iteration_id = stoi(_current_model->id.substr(_current_model->id.find("/") + 1, std::string::npos));
```
jackgerrits, line 77: Be sure to catch if this throws.

jackgerrits, line 77:
```suggestion
      int iteration_id = std::stoi(_current_model->id.substr(_current_model->id.find("/") + 1, std::string::npos));
```

rlclientlib/federation/local_client.cc
bassmang: its being passed in with the `--id` command in vw. Thinking it might make more sense to include it as an arg in create?

jackgerrits: Okay cool. I don't follow the arg in create? But i just wanted to double check it was set somewhere else already

jackgerrits (approved):

PR #584, opened 2023-05-04
[feat] multistep_loop APIs

peterychang:
Added 2 new APIs for `report_outcome` that take in the event ID and the episode object. Discussed with Wangda

### DIFF
--- a/examples/basic_usage_cpp/basic_usage_cpp.cc
+++ b/examples/basic_usage_cpp/basic_usage_cpp.cc
@@ -283,7 +283,7 @@ int basic_usage_multistep()
 
   r::api_status status;
 
-  r::live_model rl(config);
+  r::multistep_loop rl(config);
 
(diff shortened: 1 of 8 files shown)


peterychang:
I think the only remaining test that uses the old interface is rl_sim (used in the end-to-end tests)

bassmang:
> I think the only remaining test that uses the old interface is rl_sim (used in the end-to-end tests)

Would it be possible to add a test to `live_model_test.cc` like the other loop tests? And add something small to `live_model_test_legacy.cc` with the old interface. Otherwise lgtm

rajan-chari (approved):
Look good to me.  Thanks Peter

peterychang:
@bassmang Actually, I can't really add multistep into the normal testing framework because its only built into the clientlib when using the external_parser makefile. I'll merge this for now and look into expanding the external parser tests a bit more

PR #603, opened 2024-02-01
feat: static model loader

bassmang:


bindings/cs/rl.net.native/binding_static_model.h
```diff
@@ -0,0 +1,25 @@
...
+{
+
+namespace constants
+{
+const char* const BINDING_DATA_TRANSPORT = "BINDING_DATA_TRANSPORT";
+}
+class binding_static_model : public reinforcement_learning::model_management::i_data_transport
+{
+public:
+  binding_static_model(const char* weights, const size_t len);
```
rajan-chari, line 17: nit: weights -> vw_model

bindings/cs/rl.net.native/binding_static_model.h
```diff
@@ -0,0 +1,25 @@
+#pragma once
+
+#include "model_mgmt.h"
+
```
rajan-chari, line 4: a namespace alias might be useful for really long namespaces. For example...
```suggestion
namespace model_transport = reinforcement_learning::model_management;
```

bindings/cs/rl.net.native/binding_static_model.cc
```diff
@@ -0,0 +1,26 @@
...
+#include "err_constants.h"
+#include "model_mgmt.h"
+
+using namespace rl_net_native;
+using namespace reinforcement_learning;
+
+binding_static_model::binding_static_model(const char* weights, const size_t len) : weights(weights), len(len) {}
+
+int binding_static_model::get_data(
+    reinforcement_learning::model_management::model_data& data, reinforcement_learning::api_status* status)
```
rajan-chari, line 12: suggestion: judicious use of namespace alias would make things more readable.

bindings/cs/rl.net.native/binding_static_model.cc
```diff
@@ -0,0 +1,25 @@
...
+
+int binding_static_model::get_data(model_transport::model_data& data, reinforcement_learning::api_status* status)
+{
+  if (this->vw_model == nullptr || this->len == 0)
+  {
+    return reinforcement_learning::error_code::static_model_load_error;
+  }
+
+  char* buffer = data.alloc(this->len);
+  if (buffer == nullptr) { return reinforcement_learning::error_code::static_model_load_error; }
```
rajan-chari, line 19: Wondering if model_data class can use a method to encapsulate these operations.  i.e. alloc + copy + set size

bassmang: put a set_data function in data_data to handle this

bindings/cs/rl.net/FactoryContext.cs
```diff
@@ -20,6 +25,23 @@ public FactoryContext() : base(new New<FactoryContext>(CreateFactoryContext), ne
         {
         }
 
+        public FactoryContext(IEnumerable<byte> vwModelEnumerable) : base(
+            new New<FactoryContext>(() => {
+                var vwModelArray = vwModelEnumerable.ToArray();
+                GCHandle handle = GCHandle.Alloc(vwModelArray, GCHandleType.Pinned);
+                try {
```
rajan-chari, line 32: Is there a way to check if this makes a copy?  Unclear to me...

bassmang: There is a copy happening here but I believe it is necessary since they pass in a List which doesn't have a way to obtain a pointer to it's internal array

lokitoth: There is a way to avoid the copy on .NET 5 or above with List<> types, but not generally, no.

PR #604, opened 2024-02-01
Dev/azure oauth

peterychang:
Add capability to use OAUTH authorization for eventhub and blob storage access


cmake/Modules/AzureVcpkg.cmake
jackgerrits, line 1: rlclientlib already consumes dependencies from vcpkg. Why is this file needed?

examples/rl_sim_cpp/azure_credentials.h
```diff
@@ -0,0 +1,29 @@
...
+#include <azure/identity/azure_cli_credential.hpp>
+
+#include <chrono>
+#include <string>
+
+class AzureCredentials
+{
+public:
+  AzureCredentials();
+  int get_credentials(std::string& token_out, std::chrono::system_clock::time_point& expiry_out,
```
jackgerrits, line 19: Out params should be last

vcpkg.json
```diff
@@ -21,7 +22,17 @@
     "openssl",
     "rapidjson",
     "spdlog",
-    "zlib"
+    "zlib",
+    "azure-storage-blobs-cpp",
+    "azure-identity-cpp"
```
jackgerrits, line 27: These aren't needed right?

cmake/Modules/AzureVcpkg.cmake
peterychang: Builds were failing to find the module whenever I tried building without it. I'll take a closer look once I have a nuget out for Chad to test against

examples/rl_sim_cpp/azure_credentials.h
peterychang: Thats our standard here? I'll get that fixed

vcpkg.json
peterychang: azure-identity is needed for the example code, but the blobs aren't. How should we handle that?

jackgerrits: I'm not sure. Vcpkg must have some way to override this on the command line or something. It nothing else we can use manifest features like we do in vw. But this shouldn't make it into the default set.

examples/rl_sim_cpp/azure_credentials.h
jackgerrits: I thought it was a pretty common standard, not sure if we have a convention or not in this codebase

vcpkg.json
peterychang: I'll look into ways to remove conditionally remove it in the manifest

cmake/Modules/AzureVcpkg.cmake
peterychang: Figured it out, misread one of the docs, you either use vcpkg OR use this. Removed superfluous stuff

jackgerrits (changes requested):

CMakeLists.txt
```diff
@@ -33,6 +33,10 @@ if(RL_BUILD_BENCHMARKS)
   list(APPEND VCPKG_MANIFEST_FEATURES "benchmarks")
 endif()
 
+option(RL_LINK_AZURE_LIBS "Whether to build components requiring the use of Azure libraries. Requires C++14 or greater" OFF)
```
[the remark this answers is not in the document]
peterychang: I was planning on keeping it at least kind of generic in case there was ever a need to extend it with other azure libs

examples/rl_sim_cpp/azure_credentials.h
```diff
@@ -0,0 +1,30 @@
...
+#  include "configuration.h"
+#  include "future_compat.h"
+
+#  include <azure/identity/azure_cli_credential.hpp>
+#  include <azure/identity/default_azure_credential.hpp>
+#  include <chrono>
+#  include <memory>
+#  include <string>
+
+class AzureCredentials
```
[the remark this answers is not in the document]
peterychang: theoretically it can support any azure credential. There's just a bug with the DefaultCredentials that I couldn't use it for the CLI stuff here

jackgerrits (approved):

PR #611, opened 2024-10-18
fix for azure build

v-jameslongo:
fix local time conversion for azure build
added azure build cmake presets


rajan-chari (approved):
Looks good to  me.

PR #613, opened 2024-11-04
Missing header update wrt compiler differences

v-jameslongo:
**Summary**

This pull request adds the missing <cstdint> includes in container_iterator.h, decision_response.h, rl_sim_cpp/person.h and rl_sim_cpp/robot_joint.h to address uint32_t type errors that arise due to differences in how compilers handle standard headers.

**Details**

The uint32_t type was used in multiple files without including <cstdint>, which caused compilation issues on certain environments where the type wasn't implicitly available.

This fix ensures explicit inclusion of the necessary header, making the code more portable and consistent across different compiler configurations.

**Impact**

Resolves build failures on environments that are stricter about standard header inclusion.

Improves code portability and compatibility across different build environments.


rajan-chari (approved):
LGTM

